PR #11, opened 2015-01-12
BuildSystem improvements for CMake users of ezEngine

Manuzor:
As I told you personally via email, I adjusted your CMake scripts a bit in order to be able to have more than one `exports.cmake` file in a more "standard" location and I prefixed the exported targets with "ez". I also added a find-module that I wrote so users can add ezEngine modules as targets in their CMake scripts.

1. The `exports.cmake` file will now be generated to the output directory specified by the user in the cmake-gui, i.e. the `CMAKE_BINARY_DIR` variable.

1. CMake targets are now exported with the `ez` prefix, i.e. in CMake, after `include`-ing the `exports.cmake` file, you can now use this kind of code: `target_link_libraries(ezFoundation)`

1. Added the file `Code/BuildSystem/CMake/FindezEngine.cmake` which can be included by other users in their CMake scripts to quickly add ezEngine modules as dependencies. Documentation for this can be find in the script itself.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,1 @@
-Code/exports.cmake
 Output/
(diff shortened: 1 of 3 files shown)


jankrassnigg:
Sorry, I had totally forgotten about this. I have merged it into our Subversion repository now.

Manuzor:
Awesome, thanks!

PR #25, opened 2019-03-28
Sky light component 

C-Core:
Sky light component renders a portion of the scene (typically only the skybox) into a cubemap and calculates diffuse irradiance from it that is then applied as IBL to the scene.


jankrassnigg:
How do I approve?

PR #27, opened 2019-04-06
ezImage 64bit data support

marcappelsmeier:
Adds ezBlob as a container to store large data amounts.
Switched ezImage to use ezBlob & ezBlobPtr. Large textures now can be loaded in ezImage.
To fix all potential problems with the switch of GetDepthPitch and GetRowPitch to 64 bit I removed the pragma warning disable for size truncation and fixed all issues that appeared because of that. This might expose some problems in custom code which is not checked in.
Unit tests run.

### DIFF
--- a/Code/Engine/Core/Graphics/Implementation/Geometry.cpp
+++ b/Code/Engine/Core/Graphics/Implementation/Geometry.cpp
@@ -439,7 +439,7 @@ void ezGeometry::AddTesselatedRectXY(const ezVec2& size, const ezColor& color, e
   const ezVec2 halfSize = size * 0.5f;
   bool bFlipWinding = mTransform.GetRotationalPart().GetDeterminant() < 0;
 
-  const ezVec2 sizeFraction = size.CompDiv(ezVec2(uiTesselationX, uiTesselationY));
+  const ezVec2 sizeFraction = size.CompDiv(ezVec2(static_cast<float>(uiTesselationX), static_cast<float>(uiTesselationY)));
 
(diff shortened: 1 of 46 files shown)


marcappelsmeier:
> First of all, this is pretty awesome! I'm quite happy that I don't need to fix this :D
> 
> Regarding the warnings:
> I kinda agree with the change, I'm just not exactly sure anymore what the original reason was to disable them, it might have been done to smooth an integration (_whink whink_). But we'll see about that.
> 
> Regarding the blob:
> Should we split ezBlob and ezBlobPtr into two files?
> Do we really want to call it blob? How about "LargeArray"?
> 
> And it looks like ezBlobPtr is a 64-bit version of ezArrayPtr. Should it really stay that way? ezBlob is untyped, do you think ezBlobPtr keep the type template? I assumed it would also just be a (const) void* at all times.
> 
> Finally, should we add an ezImage test with a large texture (generated procedurally) ?

Regarding the file split, we can do it but since it is quite coupled and the blobptr header would need to be included in the blob header anyway I thought it's easier to do it this way.

I didn't want to add a new array class for two reasons: It could confuse people and the interface is quite distinct.

ezBlobPtr is indeed a copy of ezArrayPtr but with 64 bit compatibility. The reason is that the image libraries use the templated version in quite a few places to get the data as ezColor, ezUInt16 etc. This way the change was less than adding casts in many places.

I thought about adding a unit test but it would be one which needs to allocate at least 4GB RAM and should we go to Azure pipelines for testing then I'm pretty sure that the worker running the foundation tests won't have access to that much RAM (and I don't know what people have in their laptops / desktops).

PR #28, opened 2019-04-07
User/cm/delegate

Sanakan8472:
ezDelegate now supports lambdas of any size.
    -Any lambda that has captures will be stored in a heap-allocated storage.
    -Correctly handles captures of non-pod data now.
    -Added move semantics.
    -Any lambda that has captures will cause delegate compare to always return false.
    -Changed ezDelegate to use variadic templates.

### DIFF
--- a/Code/Engine/Foundation/Types/Implementation/DelegateHelper_inl.h
+++ b/Code/Engine/Foundation/Types/Implementation/DelegateHelper_inl.h
@@ -21,13 +21,25 @@ struct ezLambdaDelegateStorage : public ezLambdaDelegateStorageBase
   }
 
 private:
+  template < typename = typename std::enable_if< std::is_copy_constructible<Function>::value >::type>
   ezLambdaDelegateStorage(const Function& func)
     : m_func(func)
   {
   }
 
 public:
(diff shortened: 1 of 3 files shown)


Code/Engine/Foundation/Types/Implementation/DelegateHelper_inl.h
```diff
@@ -1,25 +1,63 @@
 
-template <typename R EZ_COMMA_IF(ARG_COUNT) EZ_LIST(typename ARG, ARG_COUNT)>
```
C-Core: We could get rid of this additional inl file now and move everything into Delegate_inl.h

Sanakan8472: Yes, I'll do that in a separate PR that will not have any functional changes.

Sanakan8472:
> 
> 
> Looks good to me, great work! One interesting use case that might not work with this implementation is a lambda that captures a unique ptr, or any other type that is only movable but not copyable.
> E.g.
> `ezUniquePtr something = ...; auto lamda = [blubb = std::move(something)]() { .. }`

Support added and unit test extended.

PR #29, opened 2019-04-16
Rewrote shader parser

C-Core:
- Rewrote shader parser using ezTokenizer instead of operating on strings directly.
- Added support for integer and float parsing to ezTokenizer.
- Added support for enum types in material parameters.
- Also use constructor reflection for attribute parameters to construct complex types like vec3, color etc.

### DIFF
--- a/Code/Engine/Foundation/CodeUtils/Implementation/Tokenizer.cpp
+++ b/Code/Engine/Foundation/CodeUtils/Implementation/Tokenizer.cpp
@@ -402,6 +402,20 @@ void ezTokenizer::HandleNonIdentifier()
   AddToken();
 }
 
+void ezTokenizer::GetAllLines(ezHybridArray<const ezToken*, 32>& Tokens) const
+{
+  Tokens.Clear();
+  Tokens.Reserve(m_Tokens.GetCount());
+
+  for (const ezToken& curToken : m_Tokens)
+  {
+    if (curToken.m_iType != ezTokenType::Newline)
+    {
+      Tokens.PushBack(&curToken);
+    }
+  }
+}
+
 ezResult ezTokenizer::GetNextLine(ezUInt32& uiFirstToken, ezHybridArray<ezToken*, 32>& Tokens)
 {
   Tokens.Clear();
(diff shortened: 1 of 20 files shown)


Code/Engine/RendererCore/ShaderCompiler/Implementation/ShaderParser.cpp
```diff
@@ -1,163 +1,386 @@
...
 
-        ++s; // skip (
-        szValueStart = s.GetData();
+        Accept(Tokens, uiCurToken, ",");
       }
-      else
+
+      // find matching constructor
+      auto& functions = pType->GetFunctions();
+      for (auto pFunc : functions)
```
C-Core: @Sanakan8472 Is this the right way to use constructor reflection? It seems to work ;)

Code/Engine/Foundation/CodeUtils/Implementation/Tokenizer.cpp
```diff
@@ -327,6 +340,84 @@ void ezTokenizer::HandleString(char terminator)
   AddToken();
 }
 
+void ezTokenizer::HandleNumber()
+{
+  if (m_uiCurChar == '0' && m_uiNextChar == 'x' || m_uiNextChar == 'X')
```
jankrassnigg: Please add brackets. As far as I remember && binds stronger than ||

Code/Engine/Foundation/CodeUtils/Implementation/Tokenizer.cpp
```diff
@@ -445,7 +550,7 @@ ezResult ezTokenizer::GetNextLine(ezUInt32& uiFirstToken, ezHybridArray<const ez
         {
           ezStringBuilder s1 = m_Tokens[uiFirstToken - 1].m_DataView;
           ezStringBuilder s2 = m_Tokens[uiFirstToken + 2].m_DataView;
-          ezLog::Warning("Line {0}: The \\ at the line end is in the middle of an identifier name ('{1}' and '{2}'). However, merging identifier names is currently not supported.", m_Tokens[uiFirstToken].m_uiLine, s1, s2);
+          ezLog::Warning("Line {0}: The \\ at the line end is in the middle of an identifier name ('{1}' and '{2}'). However, merging identifier names is m_uiCurCharly not supported.", m_Tokens[uiFirstToken].m_uiLine, s1, s2);
```
jankrassnigg: ORLY?

PR #32, opened 2019-04-30
Makes live updating resource data in the editor less verbose

jankrassnigg:
Messages for live updating resources are now independent of an engine side document context.

The Procedural Placement engine plugin could be removed now, unless there are other uses for it.

### DIFF
--- a/Code/Tools/EditorEngineProcess/EngineProcGameApp.cpp
+++ b/Code/Tools/EditorEngineProcess/EngineProcGameApp.cpp
@@ -557,9 +557,13 @@ void ezEngineProcessGameApplication::HandleResourceUpdateMsg(const ezResourceUpd
 
   if (hResource.IsValid())
   {
+    ezStringBuilder sResourceDesc;
+    sResourceDesc.Set(msg.m_sResourceType, "LiveUpdate");
+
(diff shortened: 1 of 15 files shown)


Code/Tools/EditorEngineProcess/EngineProcGameApp.cpp
```diff
@@ -534,3 +542,47 @@ void ezEngineProcessGameApplication::TransmitCVar(const ezCVar* pCVar)
...
+    return;
+  }
+
+  ezTypelessResourceHandle hResource = ezResourceManager::GetExistingResourceByType(pRtti, msg.m_sResourceID);
+
+  if (hResource.IsValid())
+  {
+    ezUniquePtr<ezResourceLoaderFromMemory> loader(EZ_DEFAULT_NEW(ezResourceLoaderFromMemory));
+    loader->m_ModificationTimestamp = ezTimestamp::CurrentTimestamp();
+    loader->m_sResourceDescription = "MaterialImmediateEditorUpdate";
```
C-Core: Build a nicer description from resource type + something like "ImmediateEditorUpdate"

jankrassnigg: Right, forgot about that line.

PR #38, opened 2019-05-09
fix simd math test

Sanakan8472:
Optimized builds generate 3.9999998 instead of 4.


jankrassnigg:
You don't need to create a branch and or for every change, you can just push to Dev for small changes. We usually create a branch for larger stuff or changes where we want someone else to review it.

PR #40, opened 2019-05-11
Linux build system and compile fixes.

Sanakan8472:


Code/Engine/Foundation/Algorithm/Implementation/HashingUtils_inl.h
```diff
@@ -120,7 +120,14 @@ struct ezHashHelper<T*>
 {
   EZ_ALWAYS_INLINE static ezUInt32 Hash(T* value)
   {
-    return ezHashHelper<size_t>::Hash(reinterpret_cast<size_t>(value) >> 4);
+    if constexpr (sizeof(T*) == 8)
```
Sanakan8472: size_t does not map to any ez typedef on linux (it's unsigned long it or something).

Code/Engine/RendererCore/Pipeline/Implementation/RenderData.cpp
```diff
@@ -81,7 +81,7 @@ void ezMsgExtractRenderData::AddRenderData(
   auto& cached = m_ExtractedRenderData.ExpandAndGetRef();
   cached.m_pRenderData = pRenderData;
   cached.m_uiCategory = category.m_uiValue;
-  cached.m_uiComponentIndex = 0xFFFF;
+  cached.m_uiComponentIndex = 0x7FFF;
```
Sanakan8472: Previously this did overflow as it only has 15 bits. I don't actually know the rules for shortened storage casts, did this even work correctly before?

Code/Engine/Foundation/Algorithm/Implementation/HashingUtils_inl.h
C-Core: Makes totally sense, size_t is not well defined enough. Though I don't know the guarantees of "if constexpr", is it optimized in debug builds as well? I would go for #if EZ_ENABLED(EZ_PLATFORM_64BIT) to be sure.

Code/Engine/Foundation/Communication/Implementation/IpcChannel.cpp
```diff
@@ -174,7 +174,7 @@ void ezIpcChannel::ReceiveMessageData(ezArrayPtr<const ezUInt8> data)
 
       ezProcessMessage* pMsg = (ezProcessMessage*)ezReflectionSerializer::ReadObjectFromBinary(reader, pRtti);
       ezUniquePtr<ezProcessMessage> msg(pMsg, ezFoundation::GetDefaultAllocator());
-      if (msg != NULL)
+      if (msg.Borrow() != nullptr)
```
C-Core: Borrow() shouldn't be necessary since ezUniquePtr has comparison operators for nullptr

Code/Engine/RendererCore/Pipeline/Implementation/RenderData.cpp
C-Core: Should be fine, the value is overwritten later anyways.

PR #47, opened 2019-05-25
User/ma/warning level4

marcappelsmeier:
Fixes which enable at least a subset of EZ to compile with warning level 4.


jankrassnigg:
Did you enable the switch in the cmake files?

PR #50, opened 2019-06-03
UWP build fixes

Sanakan8472:
-Various UWP compile fixes
-Added EZ_SUPPORTS_PROCESSES platform feature
-Made everything winrt in UWP as that's the only way to make cmake deploy dlls.
-Updated version of Lua to Version: 5.3.5 - 2018-06-26


Sanakan8472:
The UWP fixes commit is probably the only relevant one.

PR #59, opened 2019-06-20
C# support, RemoteTestHarness, UWP fixes

Sanakan8472:


CMakeLists.txt
```diff
@@ -14,11 +14,20 @@ ez_detect_project_name(DETECTED_NAME)
...
-	DESCRIPTION "An open-source game engine"
-	HOMEPAGE_URL "http://ezEngine.net"
-	LANGUAGES C CXX)
+if (CMAKE_HOST_SYSTEM_NAME STREQUAL "Windows")
+	# C# support only on windows desktop.
+	project (${EZ_SOLUTION_NAME}
+		VERSION 1.0
+		DESCRIPTION "An open-source game engine"
+		HOMEPAGE_URL "http://ezEngine.net"
+		LANGUAGES C CXX CSharp)
```
jankrassnigg: You could probably put the languages into a variable.

Code/BuildSystem/CMake/ezUtilsNuGet.cmake
```diff
@@ -0,0 +1,24 @@
+######################################
+### NuGet support
+######################################
+
+######################################
+### ez_requires_nuget()
+######################################
+
+function(ez_requires_nuget)
```
jankrassnigg: My "requires" functions meant to exclude the project, if X is not available.
Could you rename this?

Code/BuildSystem/CMake/ezUtilsTargetCS.cmake
```diff
@@ -0,0 +1,61 @@
+######################################
+### ez_create_target_cs(<LIBRARY | APPLICATION> <target-name>)
+######################################
+
+function(ez_create_target_cs TYPE TARGET_NAME)
+
+    set(ARG_OPTIONS NO_PCH NO_UNITY NO_QT NO_EZ_PREFIX)
```
jankrassnigg: Looks like you are not using most of these parameter options

CMakeLists.txt
Sanakan8472: Done!

Code/BuildSystem/CMake/ezUtilsTargetCS.cmake
Sanakan8472: Yep, did some cleanup.

Code/BuildSystem/CMake/ezUtilsNuGet.cmake
Sanakan8472: Good point, renamed it to ez_nuget_init(). The function should never fail unless MS server goes down or you don't have internet.

PR #64, opened 2019-07-01
User/jakras/collision interactions

jankrassnigg:


Code/EnginePlugins/PhysXPlugin/Components/Implementation/BreakableSheet.cpp
```diff
@@ -998,7 +998,7 @@ void ezBreakableSheetComponent::CreateUnbrokenPhysicsObject()
   {
     m_uiUnbrokenShapeId = pModule->CreateShapeId();
     PxFilterData filterData =
-        ezPhysX::CreateFilterData(m_uiCollisionLayerUnbroken, m_uiUnbrokenShapeId, true /* Contact reporting enabled */);
+        ezPhysX::CreateFilterData(m_uiCollisionLayerUnbroken, m_uiUnbrokenShapeId, true /* Contact reporting enabled */, false);
```
C-Core: Feels like it is time for an enum instead of more bool params

jankrassnigg:
Well, to give it a name, yes, but they are orthogonal, so it would be a bit
annoying to combine the two, ie "None, ReportContacts, SurfaceInteractions,
Both" ??

C-Core:
Then not an enum but some bit flags, these could also be reused to check for the bits in the contact callback,

PR #68, opened 2019-07-13
User/jakras/collision interactions

jankrassnigg:


Code/EnginePlugins/PhysXPlugin/Shapes/PxShapeComponent.h
```diff
@@ -3,6 +3,7 @@
 #include <GameEngine/Interfaces/PhysicsWorldModule.h>
 #include <PhysXPlugin/Components/PxComponent.h>
 #include <PhysXPlugin/Utilities/PxUserData.h>
+#include <PhysXPlugin/WorldModule/Implementation/PhysX.h>
```
C-Core: This header is implementation detail and must not be included in public header.

jankrassnigg: fixed

PR #75, opened 2019-08-11
Added macros to easily expose member functions to the scripting system

jankrassnigg:
For an example how to use the macros see TransformComponent.h/.cpp.

The feature aims to make it dead simple to make a function callable from (visual) script.
A couple of smaller additions had to be made in ezRTTI to auto-register message handlers and to ignore null-properties.
Also the visual script system got a little polish to make use of inout / out values properly.

Other than that all the "magic" is simply that the macros generate a dummy message type and message handler behind your back and hook everything up perfectly.

Review at your own risk though.


C-Core:
This looks very useful, however I'm not sure whether the implementation via messages is the right thing:
Initially messages were intended to be a generic way to communicate with components without knowing the actual type, something like the duck-typing system in java script. Also a message makes the most sense if multiple component types can handle this message.
Now if we generate one message type for each scriptable member function it is kind of against this initial idea.
Also it results in code bloat and the message handler tables in rtti can get very large and sparse if we have a lot of message types which results in more cache misses and memory waste.
 
I wonder if it would make more sense to use function reflection for this usecase?

PR #79, opened 2019-08-14
Calling reflected functions from Visual Script

jankrassnigg:
Calling reflected functions now works from visual scripts.

Functions are exposed as scriptable functions using the EZ_SCRIPTABLE_FUNCTION_PROPERTY macro. It allows to specify the parameter names and whether they are exposed as In, Out or Inout arguments.

In the Visual Script editor each function appears as a dedicated node type. "In" and "Inout" arguments show up as input data pins, "Out" and "Inout" arguments show up as output data pins. Function return values also show up as an output pin.
"In" and "Inout" arguments can also be set to constant values through the node properties, if the respective pin has no connection.

Only arguments of variant type "number" (ints + floats) and "bool" are exposed as pins. Other types, like strings are only modifiable as properties.


Code/Engine/Foundation/Reflection/Implementation/PropertyAttributes.h
```diff
@@ -712,14 +712,58 @@ class EZ_FOUNDATION_DLL ezAutoGenVisScriptMsgSender : public ezPropertyAttribute
...
 
-
 /// \brief Attribute for ezMessages to instruct the visual script framework to automatically generate a node for handling this type of
 /// message
 class EZ_FOUNDATION_DLL ezAutoGenVisScriptMsgHandler : public ezPropertyAttribute
 {
   EZ_ADD_DYNAMIC_REFLECTION(ezAutoGenVisScriptMsgHandler, ezPropertyAttribute);
 };
 
+/// \brief Attribute for ezMessages to instruct the visual script framework to automatically generate a node for handling this type of
```
C-Core: Copy paste comment

Code/Engine/Foundation/Reflection/Implementation/PropertyAttributes.h
```diff
@@ -712,14 +712,58 @@ class EZ_FOUNDATION_DLL ezAutoGenVisScriptMsgSender : public ezPropertyAttribute
...
+    m_sArg2 = szArg2;
+    m_sArg3 = szArg3;
+    m_sArg4 = szArg4;
+    m_sArg5 = szArg5;
+    m_sArg6 = szArg6;
+  }
+
+  const char* GetArgumentName(ezUInt32 index) const
+  {
+    switch (index)
```
C-Core: C has this great feature called arrays ;)

Code/Engine/GameEngine/VisualScript/VisualScriptResource.h
```diff
@@ -20,6 +20,7 @@ struct EZ_GAMEENGINE_DLL ezVisualScriptResourceDescriptor
     {
       m_isMsgSender = 0;
       m_isMsgHandler = 0;
+      m_isMsgFunctionCall = 0;
```
C-Core: isFunctionCall no Msg

Code/Engine/Foundation/Reflection/Implementation/PropertyAttributes.h
jankrassnigg: not sure if that works with replication, since these attributes need to be sent across process boundaries

jankrassnigg: Tried to use array properties, doesn't work with constant size arrays.

Code/Engine/RendererCore/Material/Implementation/MaterialResource.cpp
```diff
@@ -907,6 +912,8 @@ ezMaterialResource::CachedValues* ezMaterialResource::GetOrUpdateCachedValues()
 {
   if (!IsModified())
   {
+    // TODO: reading s_CachedValues outside the Mutex is NOT THREAD-SAFE !!!
```
jankrassnigg: @C-Core you should maybe have a look at this some time

C-Core: I know, I ran into this as well.

PR #81, opened 2019-08-21
Compile fixes for Qt 5.13.

marcappelsmeier:


Sanakan8472:
I've added some more fixes for ezInspector and TestFramework. CI is also updated with newest ThirdParty zips.

PR #89, opened 2019-09-27
Fix various issues with uninitialized memory

Ingrater:
Found by using valgrind on linux

### DIFF
--- a/Code/Engine/Foundation/Communication/Message.h
+++ b/Code/Engine/Foundation/Communication/Message.h
@@ -22,12 +22,21 @@ class EZ_FOUNDATION_DLL ezMessage : public ezReflectedClass
 {
   EZ_ADD_DYNAMIC_REFLECTION(ezMessage, ezReflectedClass);
 
-public:
-  ezMessage()
+protected:
+  explicit ezMessage(size_t messageSize)
   {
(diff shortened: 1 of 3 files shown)


Code/Engine/Foundation/Communication/Message.h
```diff
@@ -50,7 +50,8 @@ class EZ_FOUNDATION_DLL ezMessage : public ezReflectedClass
   /// \brief Calculates a hash of the message.
   EZ_ALWAYS_INLINE ezUInt32 GetHash() const
   {
-    return ezHashingUtils::xxHash32(this, m_uiSize);
```
jankrassnigg: Why?

Code/Engine/Foundation/Strings/Implementation/StringUtils.cpp
```diff
@@ -205,7 +205,7 @@ ezUInt32 ezStringUtils::ToUpperString(char* pString, const char* pStringEnd)
   char* pWriteStart = pString;
   const char* pReadStart = pString;
 
-  while (*pReadStart != '\0' && pReadStart < pStringEnd)
```
jankrassnigg: Oh yes, i already fixed some of those a while ago. Must have missed these.

Code/Engine/Foundation/Communication/Message.h
Ingrater: Change was originally requested by clemens as small performance optimization. After a brief discussion decided to revert.

PR #91, opened 2019-10-02
Better crash handling on posix systems. 

Ingrater:
Allow the OS to write a core dump by forwarding the terminating signal to the OS.

### DIFF
--- a/Code/Engine/Foundation/Basics/Platform/Win/PlatformFeatures_win.h
+++ b/Code/Engine/Foundation/Basics/Platform/Win/PlatformFeatures_win.h
@@ -78,5 +78,5 @@
 // Writing crashdumps is only supported on windows desktop
 #if EZ_ENABLED(EZ_PLATFORM_WINDOWS_DESKTOP)
 #  undef EZ_SUPPORTS_CRASH_DUMPS
-#  define EZ_SUPPORTS_CRASH_DUMPS EZ_OFF
+#  define EZ_SUPPORTS_CRASH_DUMPS EZ_ON
 #endif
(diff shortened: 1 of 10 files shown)


Code/Engine/Foundation/System/MiniDumpUtils.h
```diff
@@ -15,6 +15,9 @@ extern "C"
 /// \brief Functionality for writing process mini-dumps (callstacks, memory snapshots, etc)
 struct EZ_FOUNDATION_DLL ezMiniDumpUtils
 {
+  /// \brief Checks if the MiniDumpUtils are supported on the current platform
+  static bool IsSupported();
```
jankrassnigg: Since this is a compile time thing, should it maybe rather be a #define ? E.g. EZ_SUPPORTS_CRASHDUMPS ?

Could wrap the entire ezMiniDumpUtils class in that then.

Code/Engine/Foundation/Basics/Platform/Win/PlatformFeatures_win.h
```diff
@@ -75,3 +75,8 @@
 #  error "Unknown architecture."
 #endif
 
+// Writing crashdumps is only supported on windows desktop
+#if EZ_ENABLED(EZ_PLATFORM_WINDOWS_DESKTOP)
+#  undef EZ_SUPPORTS_CRASH_DUMPS
+#  define EZ_SUPPORTS_CRASH_DUMPS EZ_OFF
```
jankrassnigg: EZ_ON

Ingrater: ups ^^

PR #94, opened 2019-10-05
Add advanced docking system support

Sanakan8472:
Change Inspector to use Advanced Docking System
Changed editor to use advanced docking system   
    -Each project has its own layout
    -Opening a project that has no layout yet will use the layout of the last previously saved project
    -Only affects document windows and application panels for now


Code/Editor/EditorFramework/Panels/AssetBrowserPanel/AssetBrowserPanel.cpp
```diff
@@ -14,7 +14,7 @@ ezQtAssetBrowserPanel::ezQtAssetBrowserPanel()
 {
   setupUi(this);
 
-  m_pStatusBar = new QStatusBar(this);
+  m_pStatusBar = new QStatusBar(nullptr);
```
jankrassnigg: why?

Code/Tools/Libs/GuiFoundation/ContainerWindow/ContainerWindow.cpp
```diff
@@ -10,78 +10,56 @@
...
+  setMinimumSize(QSize(800, 600));
   m_bWindowLayoutRestored = false;
   m_pStatusBarLabel = nullptr;
   m_iWindowLayoutRestoreScheduled = 0;
 
-  s_AllContainerWindows.PushBack(this);
+  s_pContainerWindow = this;
 
-  setObjectName(GetUniqueName().GetData());
+  setObjectName("ezEditor");
```
jankrassnigg: Several hardcoded values here. Can we do better?

Code/Editor/EditorFramework/Panels/AssetBrowserPanel/AssetBrowserPanel.cpp
Sanakan8472: It's added manually to the dockWidgetContents later so adding it to 'this' will just unneccessarily reparent it again.

Code/Tools/Libs/GuiFoundation/ContainerWindow/ContainerWindow.cpp
Sanakan8472: Now that the ezQtContainerWindow is a singleton it is no longer needed to dynamically construct any names and all names become constants. If the constants are only used once I'd argue that moving them elsewhere only reduces readability.
The only thing that remains are the constants used for the ezSettings. As these are only used in the save and restore function I left them alone as it would generate more code to separate them out then is currently present in those functions so I'd argue it is of questionable value. However, I have deduplicated some of the shared code between those functions which cuts down on code size and constants.

PR #95, opened 2019-10-06
Stores a handle to the mesh resource in the texture context to keep the mesh data alive.

marcappelsmeier:


jankrassnigg:
Yep, that's the right fix. Would you terribly mind to do this for all other document types as well? :-P

jankrassnigg:
I did not expect this to work :D

After a quick check I think the same pattern also exists in ParticleContext.cpp and KrautTreeContext.cpp. If you don't want to fix them now, I can do that at some point, though.

PR #103, opened 2019-11-07
Add support for conservative rasterization in rasterizer state

marcappelsmeier:
Add support for conservative rasterization in rasterizer state and in the D3D11 implementation

### DIFF
--- a/Code/Engine/RendererCore/Shader/Implementation/ShaderStateDescriptor.cpp
+++ b/Code/Engine/RendererCore/Shader/Implementation/ShaderStateDescriptor.cpp
@@ -9,7 +9,7 @@ struct ezShaderStateVersion
     Version0 = 0,
     Version1,
     Version2,
-
+    Version3,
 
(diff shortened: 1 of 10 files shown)


Code/Engine/RendererCore/Shader/Implementation/ShaderStateDescriptor.cpp
```diff
@@ -163,6 +164,11 @@ void ezShaderStateResourceDescriptor::Load(ezStreamReader& stream)
     stream >> m_RasterizerDesc.m_fDepthBiasClamp;
     stream >> m_RasterizerDesc.m_fSlopeScaledDepthBias;
     stream >> m_RasterizerDesc.m_iDepthBias;
+
+    if (uiVersion >= ezShaderStateVersion::Version3)
+    {
+      stream >> m_RasterizerDesc.m_bScissorTest;
```
C-Core: Should be m_bConservativeRasterization

PR #105, opened 2019-11-14
Template document support

marcappelsmeier:
The document manager checks now if a template document exists in the "DocumentTemplates" folder in the current project directory. If a document matching the asset type for which a new document is about to be created exists it will clone this document instead of creating one from scratch.

### DIFF
--- a/Code/Tools/Libs/ToolsFoundation/Document/Implementation/DocumentManager.cpp
+++ b/Code/Tools/Libs/ToolsFoundation/Document/Implementation/DocumentManager.cpp
@@ -9,7 +9,7 @@
 #include <Foundation/Serialization/DdlSerializer.h>
 #include <Foundation/IO/FileSystem/DeferredFileWriter.h>
 #include <ToolsFoundation/Document/DocumentUtils.h>
-#include <Project/ToolsProject.h>
+#include <ToolsFoundation/Project/ToolsProject.h>
 
 // clang-format off
 EZ_BEGIN_DYNAMIC_REFLECTED_TYPE(ezDocumentManager, 1, ezRTTINoAllocator);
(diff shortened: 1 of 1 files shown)


Code/Tools/Libs/ToolsFoundation/Document/Implementation/DocumentManager.cpp
```diff
@@ -9,6 +9,7 @@
 #include <Foundation/Serialization/DdlSerializer.h>
 #include <Foundation/IO/FileSystem/DeferredFileWriter.h>
 #include <ToolsFoundation/Document/DocumentUtils.h>
+#include <Project/ToolsProject.h>
```
jankrassnigg: #include path should be <ToolsFoundation/Project/ToolsProject.h>

Code/Tools/Libs/ToolsFoundation/Document/Implementation/DocumentManager.cpp
```diff
@@ -219,6 +220,41 @@ ezStatus ezDocumentManager::CreateOrOpenDocument(bool bCreate, const char* szDoc
...
+      // it and use that as the new document instead of creating one from scratch.
+      if (bCreate)
+      {
+        ezString ProjectDirectory = ezToolsProject::GetSingleton()->GetProjectDirectory();
+        ezStringBuilder TemplateDocumentPath = ProjectDirectory;
+        ezStringBuilder DocumentFileName;
+        TemplateDocumentPath.AppendPath("DocumentTemplates", szDocumentTypeName);
+        TemplateDocumentPath.ChangeFileExtension("ezDocument");
+
+        if (ezOSFile::ExistsFile(TemplateDocumentPath.GetData()))
```
jankrassnigg: The ".GetData()" is redundant.

jankrassnigg: Same in all the calls below.

Code/Tools/Libs/ToolsFoundation/Document/Implementation/DocumentManager.cpp
```diff
@@ -219,6 +220,41 @@ ezStatus ezDocumentManager::CreateOrOpenDocument(bool bCreate, const char* szDoc
...
+          EZ_PROFILE_SCOPE(szDocumentTypeName);
+
+          ezUuid CloneUuid;
+          if (CloneDocument(TemplateDocumentPath.GetData(), sPath.GetData(), CloneUuid).Succeeded())
+          {
+            status = OpenDocument(szDocumentTypeName, sPath.GetData(), out_pDocument, flags, pOpenContext);
+
+            if(status.Failed())
+            {
+              ezLog::SeriousWarning("Couldn't open cloned template document, proceeding with normal creation behavior.");
```
jankrassnigg: Error ?

Code/Tools/Libs/ToolsFoundation/Document/Implementation/DocumentManager.cpp
```diff
@@ -219,6 +220,41 @@ ezStatus ezDocumentManager::CreateOrOpenDocument(bool bCreate, const char* szDoc
...
     {
+      // See if there is a default asset document registered for the type, if so clone
+      // it and use that as the new document instead of creating one from scratch.
+      if (bCreate)
+      {
+        ezString ProjectDirectory = ezToolsProject::GetSingleton()->GetProjectDirectory();
+        ezStringBuilder TemplateDocumentPath = ProjectDirectory;
+        ezStringBuilder DocumentFileName;
+        TemplateDocumentPath.AppendPath("DocumentTemplates", szDocumentTypeName);
+        TemplateDocumentPath.ChangeFileExtension("ezDocument");
```
jankrassnigg: Why ".ezDocument" ? Wouldn't it be easier to use the type specific extension, so that one can easily open the document and modify it? Or is it intended to hide the docs from the asset browser? In that case, you could also put them in a folder with the "_data" postfix, to hide them by default.

jankrassnigg:
Btw. you can ignore the build failure on Mac, it is a known problem, because the Mac updated to latest Qt automatically and the rest of the scripts currently cannot handle that.

Code/Tools/Libs/ToolsFoundation/Document/Implementation/DocumentManager.cpp
marcappelsmeier: I'd like to keep the warning as I don't want to break asset creation in case something goes wrong here.

Code/Tools/Libs/ToolsFoundation/Document/Implementation/DocumentManager.cpp
marcappelsmeier: Changed how it works, this is the simple implementation. The idea was to have support for different templates per asset type so that the user can choose in the New > ... context menu for example. Templates are now Default.Extension in the templates folder. This way they can also be edited easily from within the editor.

PR #107, opened 2019-11-24
TypeScript + Inspector improvements

jankrassnigg:


Code/Engine/GameEngine/Prefabs/Implementation/PrefabReferenceComponent.cpp
```diff
@@ -31,6 +31,12 @@ EZ_BEGIN_COMPONENT_TYPE(ezPrefabReferenceComponent, 4, ezComponentMode::Static)
     EZ_MAP_ACCESSOR_PROPERTY("Parameters", GetParameters, GetParameter, SetParameter, RemoveParameter)->AddAttributes(new ezExposedParametersAttribute("Prefab")),
   }
   EZ_END_PROPERTIES;
+  EZ_BEGIN_FUNCTIONS
+  {
+    EZ_SCRIPT_FUNCTION_PROPERTY(InstantiatePrefab),
```
C-Core: Do we really want to expose these as script functions? The prefab is instantiated when the component is activated and cleaned when the component is deactivated. So you can do all you need via simple activate and deactivate.

jankrassnigg: Right, made it private now as well.

PR #114, opened 2020-01-01
TypeScript: docs, tests, sample prefabs

jankrassnigg:
Added a "SendEventMessage" function to ezGameObject to be able to raise an event on another GO, such as "I applied damage to you", so that from that GO the message will bubble up to the script that handles it.


Code/Engine/Core/World/GameObject.h
```diff
@@ -348,16 +350,16 @@ class EZ_CORE_DLL ezGameObject
 
 
   /// \brief Sends a message to all components of this object.
-  bool SendMessage(ezMessage& msg);
+  EZ_ALWAYS_INLINE bool SendMessage(ezMessage& msg) { return SendMessageInternal(msg, false); }
```
C-Core: Move to inl file

Code/Engine/Core/World/Component.h
```diff
@@ -92,14 +92,11 @@ class EZ_CORE_DLL ezComponent : public ezReflectedClass
 
 
   /// \brief Sends a message to this component.
-  bool SendMessage(ezMessage& msg);
-  bool SendMessage(ezMessage& msg) const;
-
-  /// \brief Queues the message for the given phase and processes it later in that phase.
-  void PostMessage(const ezMessage& msg, ezObjectMsgQueueType::Enum queueType) const;
+  EZ_ALWAYS_INLINE bool SendMessage(ezMessage& msg) { return SendMessageInternal(msg, false); }
```
C-Core: I don't really understand why this extra bool is needed?

Code/Engine/Core/World/Implementation/World.cpp
```diff
@@ -111,6 +112,9 @@ void ezWorld::Clear()
   {
     DeleteObjectNow(it->GetHandle());
   }
+
+  // make sure all dead objects and components are cleared right now
+  Update();
```
C-Core: This is dangerous since it will call all the update logic which can have whatever side effects. If you really need this here you should only call DeleteDeadObjects(); and DeleteDeadComponents(); directly.

Code/Engine/Core/Messages/EventMessage.h
```diff
@@ -80,17 +77,19 @@ class ezEventMessageSender : public ezMessageSenderBase<EventMessageType>
...
-                                                       const ezGameObject* pSearchObject) const
+    const ezGameObject* pSearchObject) const
   {
     msg.m_hSenderObject = pSenderComponent->GetOwner() != nullptr ? pSenderComponent->GetOwner()->GetHandle() : ezGameObjectHandle();
     msg.m_hSenderComponent = pSenderComponent->GetHandle();
 
     if (*reinterpret_cast<ezUInt64*>(&m_hCachedReceiver) == 0xFFFFFFFFFFFFFFFF)
     {
-      m_hCachedReceiver = ezInternal::EventMessageSenderHelper::FindReceiver(msg, pSenderComponent, pSearchObject);
+      if (const ezComponent* pReceiver = pSenderComponent->GetWorld()->FindEventMsgHandler(msg, pSearchObject))
```
C-Core: You need to set the m_hCachedReceiver to an invalid component handle if no component is found, otherwise it will try to find an event handler component again and again.

Code/Engine/GameEngine/Messages/DamageMessage.h
```diff
@@ -1,19 +1,22 @@
...
-#include <Foundation/Communication/Message.h>
 
-struct EZ_GAMEENGINE_DLL ezMsgDamage : public ezMessage
+struct EZ_GAMEENGINE_DLL ezMsgDamage : public ezEventMessage
 {
-  EZ_DECLARE_MESSAGE_TYPE(ezMsgDamage, ezMessage);
+  EZ_DECLARE_MESSAGE_TYPE(ezMsgDamage, ezEventMessage);
 
   double m_fDamage = 0;
+  ezString m_sHitObjectName; ///< The actual game object that was hit (may be a child of the object to which the message is sent)
```
C-Core: Why the name and not a gameobject handle?

C-Core:
I like the addition of the SendEventMessage function. I guess we can live with the slightly more expensive search for the EventHandlerComponent in these cases.

Code/Engine/Core/World/Component.h
jankrassnigg: SendMessage always returned the bool. PostMessage does not.

jankrassnigg: It just shows them in the diff side-by-side, because one of the PostMessage variants was removed (was redundant).

Code/Engine/GameEngine/Messages/DamageMessage.h
jankrassnigg: So right now the script integration is not able to handle GameObject and Component handles, because for that the ezVariant requires an upgrade, which Christopher wanted to do at some point. Therefore instead of hacking anything in there to make handles work, I am just waiting for the variant change and may change this message then as well.

On the other hand, it also seems to be convenient to use a string here, because this way you can reuse the same name for objects that should have the same effect (multiple hit boxes for a complex object). However, that is probably not such a common use case, so wouldn't mind to change it again.

Code/Engine/Core/Messages/EventMessage.h
jankrassnigg: done

Code/Engine/Core/World/Implementation/World.cpp
jankrassnigg: done

Code/Engine/Core/World/GameObject.h
jankrassnigg: done

jankrassnigg:
@Sanakan8472 not sure what to do about the editor test failure. I can't think of a fix or work-around at the moment, other than to hack in, that the test skips the TypeScript assets, but that would be pretty awful.

Code/Engine/Core/World/Component.h
C-Core: I mean this bool bWasPostedMsg thing. Why is that needed?

jankrassnigg:
Ah, right. The message handler for typescript components has to synchronize
stuff, but not for posted messages. This bool allows me to optimize that
case.

C-Core <notifications@github.com> schrieb am Sa., 4. Jan. 2020, 12:09:

PR #117, opened 2020-01-15
Raycast placement component

marcappelsmeier:


Code/Engine/GameEngine/Gameplay/Implementation/RaycastComponent.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+#include <GameEngine/Gameplay/RaycastComponent.h>
+#include <GameEngine/Interfaces/PhysicsWorldModule.h>
+#include <Core/WorldSerializer/WorldReader.h>
+#include <Core/WorldSerializer/WorldWriter.h>
+#include <Core/Messages/TriggerMessage.h>
+#include <RendererCore/Debug/DebugRenderer.h>
+
+// clang-format off
+EZ_BEGIN_COMPONENT_TYPE(ezRaycastComponent, 1, ezComponentMode::Dynamic)
```
jankrassnigg: The component itself is not dynamic, is it? It only changes the transform of the target.

Code/Engine/GameEngine/Gameplay/Implementation/RaycastComponent.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+  float fHitDistance = m_fMaxDistance;
+  ezPhysicsHitResult hit;
+  if (m_pPhysicsWorldModule->CastRay(rayStartPosition, rayDir, m_fMaxDistance, m_uiCollisionLayerEndPoint, hit))
+  {
+    fHitDistance = hit.m_fDistance;
+
+    if (!pEndObject->IsActive() && m_bDisableTargetObjectOnNoHit)
+    {
+      pEndObject->SetActive(true);
```
jankrassnigg: Just ran into this as well: We may need to enhance the active state on objects to be recursive and editable in the editor (it glitches when you expose it atm).

Code/Engine/GameEngine/Gameplay/Implementation/RaycastComponent.cpp
```diff
@@ -0,0 +1,190 @@
...
+      {
+        pEndObject->SetActive(true);
+      }
+    }
+  }
+
+  if (!m_sTriggerMessage.IsEmpty() && m_uiCollisionLayerEndPoint != m_uiCollisionLayerTrigger)
+  {
+    ezPhysicsHitResult triggerHit;
+    if (m_pPhysicsWorldModule->CastRay(rayStartPosition, rayDir, m_fMaxDistance, m_uiCollisionLayerTrigger, triggerHit) && triggerHit.m_fDistance < fHitDistance)
```
jankrassnigg: Instead of m_fMaxDistance you could use the distance of the previous hit, you are only interested in things that are closer.

Code/Engine/GameEngine/Gameplay/Implementation/RaycastComponent.cpp
marcappelsmeier: Yeah, that is a thing I thought about. I have a change which makes this static but adds code which makes the target object dynamic if it isn't.

Code/Engine/GameEngine/Gameplay/Implementation/RaycastComponent.cpp
marcappelsmeier: If I have the situation (A ray start, B wall hit):
A --- 1 --- 2 --- B
And I move 1 out of the ray the distance increases but I still want to trigger that I hit 2 again.

Code/Engine/GameEngine/Gameplay/RaycastComponent.h
```diff
@@ -0,0 +1,82 @@
...
+
+public:
+
+  void SetTriggerMessage(const char* sz);     // [ property ]
+  const char* GetTriggerMessage() const; // [ property ]
+
+  void SetRaycastEndObject(const char* szReference); // [ property ]
+
+  ezGameObjectHandle m_hRaycastEndObject;
+  float m_fMaxDistance = 100.0f;
```
jankrassnigg: [ property ]

Code/Engine/GameEngine/Gameplay/Implementation/RaycastComponent.cpp
jankrassnigg: I think that can be left to the user, it will print a warning if you forget.

Code/Engine/GameEngine/Gameplay/Implementation/RaycastComponent.cpp
jankrassnigg: I assumed you make the raycast for static stuff first, so you will hit B first. Then you do the second raycast, now you only need the distance to B, because you can only hit 1 or 2, if they are closer than B.

marcappelsmeier: ah, right - that should've been fHitDistance, not m_fMaxDistance.

PR #123, opened 2020-01-23
Typescript + demo asset improvements

jankrassnigg:
* Raycast component is now updated in such a way that the target object is moved within the same frame
* Beamcomponent deactivates mesh if target object is inactive
* Added ezDebugRenderer:;DrawLines2D
* Last TS file change dates are cached now
* TS code is now updated on scene export
* Better JS error reporting (call stacks)
* Fixed a bug in the breakable sheet component, which meant that the "break point position" was in the wrong space.
* Added reading and writing cvars from scripts


Code/Engine/RendererCore/Debug/DebugRenderer.h
```diff
@@ -63,6 +63,8 @@ class EZ_RENDERERCORE_DLL ezDebugRenderer
 
   static void DrawLines(const ezDebugRendererContext& context, ezArrayPtr<Line> lines, const ezColor& color);
 
+  static void DrawLines2D(const ezDebugRendererContext& context, ezArrayPtr<Line> lines, const ezColor& color);
+
```
C-Core: To match the convention of the other methods it should be called Draw2DLines

PR #132, opened 2020-02-08
Logging cleanup and ETW support

Sanakan8472:
-Added ETW provider for UWP testing.
-Move various error handling code to use ezLog::Print instead of platform specific functions.
-Add ETW tracing support to the RemoteTestHarness.

### DIFF
--- a/Code/Engine/Foundation/Logging/Implementation/Win/ETWProvider_win.cpp
+++ b/Code/Engine/Foundation/Logging/Implementation/Win/ETWProvider_win.cpp
@@ -34,14 +34,13 @@ ezETWProvider::~ezETWProvider()
   TraceLoggingUnregister(g_ezETWLogProvider);
 }
 
-void ezETWProvider::LogMessge(ezLogMsgType::Enum eventType, ezUInt8 uiIndentation, const char* szText, const char* szTag)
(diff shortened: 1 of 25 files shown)


Code/Engine/Foundation/Logging/Implementation/Win/ETWProvider_win.cpp
```diff
@@ -0,0 +1,52 @@
...
+{
+  TraceLoggingUnregister(g_ezETWLogProvider);
+}
+
+void ezETWProvider::LogMessge(ezLogMsgType::Enum eventType, ezUInt8 uiIndentation, const char* szText, const char* szTag)
+{
+  TraceLoggingWrite(g_ezETWLogProvider,
+    "LogMessge",
+    TraceLoggingValue((int)eventType, "Type"),
+    TraceLoggingValue(uiIndentation, "Indentation"),
```
jakrams: Does it have any value to add the indentation and the tag to etw?

Sanakan8472: The tag, probably not. The indentation I left as a value to not have to make another copy of the string when using the ETW log writer. That way it is pretty light weight to use.

Sanakan8472: Removed 'Tag'.

PR #169, opened 2020-03-17
Update cpputf8 third party dependency

Ingrater:
to 944ef0561ddcd33eb4fd94934538458b2b2de252 from https://github.com/nemtrif/utfcpp


Code/UnitTests/FoundationTest/Strings/UnicodeUtilsTest.cpp
```diff
@@ -1,4 +1,4 @@
-﻿#include <FoundationTestPCH.h>
+#include <FoundationTestPCH.h>
```
jakrams: The fact that this line is highlighted as "changed" most likely means that you lost the UTF-8 BOM. This always happens when we modify these files and then the tests fail.

If the file still contains Unicode letters, you need to use Visual Studio's "Save As" dialog, then "Save with Encoding" and use Utf-8 WITH BOM. Probably need to do this twice, because VS is "a little buggy" and often does it incorrectly the first time.

PR #181, opened 2020-03-31
Added apply-to object reference to decals

jankrassnigg:
Also fixed how scaling is applied when the projection axis is not +X.
And fixed a crash in object picking.


Code/Engine/RendererCore/Decals/Implementation/DecalComponent.cpp
```diff
@@ -401,6 +401,54 @@ void ezDecalComponent::OnMsgExtractRenderData(ezMsgExtractRenderData& msg) const
...
+      {
+        m_uiApplyOnlyToId = pRenderComponent->GetUniqueIdForRendering();
+      }
+    }
+  }
+
+  if (uiPrevId != m_uiApplyOnlyToId && GetOwner()->IsStatic())
+  {
+    // TODO: clear render data cache ?
+    TriggerLocalBoundsUpdate();
```
C-Core: Why do we trigger a bounds update here?

jankrassnigg:
Ah right, I want to make sure for static objects the render data cache is
invalidated, could not find a better way.

C-Core <notifications@github.com> schrieb am Di., 31. März 2020, 21:39:

Code/Engine/RendererCore/Decals/Implementation/DecalComponent.cpp
jankrassnigg: fixed

Code/Engine/RendererCore/Meshes/Implementation/MeshComponentBase.cpp
```diff
@@ -192,13 +192,19 @@ void ezMeshComponentBase::OnMsgExtractRenderData(ezMsgExtractRenderData& msg) co
...
+
     // Determine render data category.
     ezRenderData::Category category = m_RenderDataCategory;
     if (category == ezInvalidRenderDataCategory)
     {
       if (hMaterial.IsValid())
       {
         ezResourceLock<ezMaterialResource> pMaterial(hMaterial, ezResourceAcquireMode::AllowLoadingFallback);
+
+        if (pMaterial.GetAcquireResult() == ezResourceAcquireResult::LoadingFallback)
```
jankrassnigg, line 205: @C-Core kind of an important fix. I already fixed this in the greybox component previously, but this way is more elegant IMO, so I applied the same fix there now.

PR #183, opened 2020-04-02
Decal + Particle Effect improvements

jankrassnigg:


Code/Engine/RendererCore/Lights/Implementation/DirectionalLightComponent.cpp
```diff
@@ -14,8 +14,8 @@ EZ_BEGIN_COMPONENT_TYPE(ezDirectionalLightComponent, 3, ezComponentMode::Static)
 {
   EZ_BEGIN_PROPERTIES
   {
-    EZ_ACCESSOR_PROPERTY("NumCascades", GetNumCascades, SetNumCascades)->AddAttributes(new ezClampValueAttribute(1, 4), new ezDefaultValueAttribute(3)),
-    EZ_ACCESSOR_PROPERTY("MinShadowRange", GetMinShadowRange, SetMinShadowRange)->AddAttributes(new ezClampValueAttribute(0.0f, ezVariant()), new ezDefaultValueAttribute(50.0f), new ezSuffixAttribute(" m")),
+    EZ_ACCESSOR_PROPERTY("NumCascades", GetNumCascades, SetNumCascades)->AddAttributes(new ezClampValueAttribute(1, 4), new ezDefaultValueAttribute(2)),
```
C-Core, line 17: Why do we change those default values?

Code/Engine/Texture/TexConv/Implementation/TextureAtlas.cpp
```diff
@@ -103,18 +116,21 @@ ezResult ezTexConvProcessor::LoadAtlasInputs(const ezTextureAtlasCreationDesc& a
...
+
+      // layer 0 must have the exact same size as the alpha texture
+      EZ_SUCCEED_OR_RETURN(ConvertAndScaleImage(srcItem.m_sLayerInput[0], item.m_InputImage[0], uiResX, uiResY));
+
+      // copy alpha channel into layer 0
+      ezImageUtils::CopyChannel(item.m_InputImage[0], 3, alphaImg, 0);
+
       // rescale all layers to be no larger than the alpha mask texture
-      for (ezUInt32 layer = 0; layer < atlasDesc.m_Layers.GetCount(); ++layer)
+      for (ezUInt32 layer = 1; layer < atlasDesc.m_Layers.GetCount(); ++layer)
```
C-Core, line 127: Is this rescaling really needed? Each layer has it's own rect in the atlas so they don't need to be the same size.

Code/Engine/RendererCore/Lights/Implementation/DirectionalLightComponent.cpp
jankrassnigg: The default costs A LOT of performance and in most games these would be "high end" settings. In my tests I could also not see a difference between 2 and 3 layers. This way the user doesn't accidentally waste performance when it is not needed for their use case.

Code/Engine/Texture/TexConv/Implementation/TextureAtlas.cpp
jankrassnigg: Well, this only happens when you combine a separate alpha channel with the others. If you already have your alpha channel in the base color texture, this code path is not taken.
The (current) rationally for me is, that I mostly implemented the separate alpha channel to combine it with arbitrary other textures (e.g. ground textures). Those are typically very high resolution and would fill up the atlas quickly. Open to suggestions, though. Could also add an option per decal asset.

PR #185, opened 2020-04-04
Fixed crashes in PhysX integration

C-Core:
* Fixed crash when accessing deleted actors in fetchresults when no simulation step was executed due to high framerate
* Fixed crash in raycasts or other queries when a shape was hit whose component was just deleted and the owner was already invalid. 
* Also made sure that user data is valid until the next simulation step is done when a component was deleted.

### DIFF
--- a/Code/EnginePlugins/PhysXPlugin/WorldModule/Implementation/PhysXWorldModule.cpp
+++ b/Code/EnginePlugins/PhysXPlugin/WorldModule/Implementation/PhysXWorldModule.cpp
@@ -208,7 +208,13 @@ namespace
       out_Result.m_hSurface = ezSurfaceResourceHandle(pSurface);
     }
   }
+
+  static thread_local ezDynamicArray<PxOverlapHit, ezStaticAllocatorWrapper> g_OverlapHits;
+  static thread_local ezDynamicArray<PxRaycastHit, ezStaticAllocatorWrapper> g_RaycastHits;
 } // namespace
+
+EZ_DEFINE_AS_POD_TYPE(PxOverlapHit);
+EZ_DEFINE_AS_POD_TYPE(PxRaycastHit);
 
 //////////////////////////////////////////////////////////////////////////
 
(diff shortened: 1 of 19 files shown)


jankrassnigg:
Does our frame allocator act like a stack allocator? Otherwise this could fill up or frame memory very quickly.

Code/EnginePlugins/PhysXPlugin/WorldModule/Implementation/PhysXWorldModule.cpp
```diff
@@ -655,11 +648,41 @@ ezUInt32 ezPhysXWorldModule::CreateShapeId()
...
-  EZ_ASSERT_DEV(uiShapeId != ezInvalidIndex, "Trying to delete an invalid shape id");
+  if (uiShapeId == ezInvalidIndex)
+    return;
 
   m_FreeShapeIds.PushBack(uiShapeId);
+
+  uiShapeId = ezInvalidIndex;
+}
+
+ezUInt32 ezPhysXWorldModule::AllocateUserData(ezPxUserData*& out_pUserData)
```
jankrassnigg, line 661: Are these functions thread safe?

Code/EnginePlugins/PhysXPlugin/WorldModule/Implementation/PhysXWorldModule.cpp
```diff
@@ -735,8 +758,12 @@ bool ezPhysXWorldModule::RaycastAll(ezPhysicsCastResultArray& out_Results, const
     filterData.flags |= PxQueryFlag::eDYNAMIC;
   }
 
+  ezArrayPtr<PxRaycastHit> raycastHits = EZ_NEW_ARRAY(ezFrameAllocator::GetCurrentAllocator(), PxRaycastHit, 256);
```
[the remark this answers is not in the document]
C-Core: It could, but then you would pay the price for default initialization of the whole array, even with SetCountUninitialized

Code/EnginePlugins/PhysXPlugin/WorldModule/Implementation/PhysXWorldModule.cpp
C-Core: No they aren't, but they don't need to be since the whole component init/deinit is synchronous anyways.

PR #197, opened 2020-04-21
Asset Curator Fixes

Sanakan8472:
Add set operations to ezHashSet
* Allow iterating ezHashSet while removing elements using iterators-based Remove function.

Add ezTaskGroupID parameter in ezOnTaskGroupFinishedCallback
* Made ezTaskGroupID compare and sortable.
* Expose ezOnTaskGroupFinishedCallback in StartSingleTask

Moved asset watcher to its own class
* Removed auto rescan filesystem logic. Everything is done in the background and only for specific folders.
  * Fixes all known updated issues when changing files externally via CC tools.
* Optimized various functions in the ezAssetCurator.
  * FindSubAsset no longer iterates over all assets unless bExhaustiveSearch is set.
  * m_ReferencedFiles is now a ezMap to allows finding all files in a folder efficiently
  * ezUuid keyed containers are now ezHashMap/ezHashSet

### DIFF
--- a/Code/Editor/EditorFramework/Assets/AssetCurator.h
+++ b/Code/Editor/EditorFramework/Assets/AssetCurator.h
@@ -16,6 +16,7 @@
 #include <Foundation/Threading/Mutex.h>
 #include <Foundation/Threading/TaskSystem.h>
 #include <Foundation/Time/Timestamp.h>
+#include <Foundation/IO/DirectoryWatcher.h>
 #include <GameEngine/Configuration/PlatformProfile.h>
 #include <ToolsFoundation/Document/DocumentManager.h>
 #include <tuple>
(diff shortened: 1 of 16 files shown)


Code/Editor/EditorFramework/Assets/AssetCurator.h
```diff
@@ -27,8 +28,9 @@ class ezDirectoryWatcher;
 class ezProcessTask;
 struct ezFileStats;
 class ezAssetProcessorLog;
+class ezAssetWatcher;
 
-#if 0 // Define to enable extensive curator profile scopes
+#if 1 // Define to enable extensive curator profile scopes
```
jankrassnigg, line 33: supposed to stay on ?

Sanakan8472: No, good catch!

PR #205, opened 2020-04-25
PhysX 4.1.1 upgrade

Sanakan8472:
* Fixed all compile issues according to migration guide.
* Removed ezPhysX::nvTools as no longer part of the PhysX github repo and unused.
* Binaries are automatically downloaded from third party releases: https://github.com/ezEngine/thirdparty/releases/tag/PhysX-4.1.1
* Docs updated with manual compile instructions: https://ezengine.github.io/docs/appendix/third-party-code.html
* UWP build improvements
  * Third party dlls are now marked as content and are PUBLIC target_sources, making sure that the files are packaged with the final application. The only downside is that to resolve the sources which are generator expressions we do need to call e.g. find_package(EzFmod REQUIRED) multiple times as otherwise the generator expressions are unknown in the target project that inherited the sources.
  * arm32 and arm64 build targets now have a different EZ_CMAKE_ARCHITECTURE_POSTFIX than UWP x86.
  * d3dcompiler_47.dll does not exist for arm in windows SDK. However, the dll is in system32 on the target OS so we skip the copy step for it.
  * Some smaller arm fixes in various cmake files.


Code/EnginePlugins/PhysXPlugin/WorldModule/Implementation/PhysXWorldModule.cpp
```diff
@@ -581,7 +581,9 @@ void ezPhysXWorldModule::Initialize()
 
     desc.flags |= PxSceneFlag::eENABLE_ACTIVE_ACTORS;
     desc.flags |= PxSceneFlag::eEXCLUDE_KINEMATICS_FROM_ACTIVE_ACTORS;
-    desc.flags |= PxSceneFlag::eENABLE_KINEMATIC_PAIRS;
+    desc.kineKineFilteringMode = PxPairFilteringMode::eKEEP;
+    //TODO: Do we need this? PhysX4 upgrade.
+    //desc.staticKineFilteringMode = PxPairFilteringMode::eKEEP;
```
C-Core, line 586: I can't think of a usecase for static <-> kinematic pairs, so we don't need it atm.

Code/Tools/Player/CMakeLists.txt
```diff
@@ -14,6 +14,28 @@ target_link_libraries(${PROJECT_NAME}
   GameEngine
 )
 
+if (EZ_CMAKE_PLATFORM_WINDOWS_UWP)
+  # Due to app sandboxing we need to explicitly name required plugins for UWP.
+  target_link_libraries(${PROJECT_NAME}
```
jankrassnigg, line 19: Could this be done through our finalize.cmake files? No high priority though. They are used to add a dependency to another project without the other project needing to know about it. Eg kraut can add a dependency to player, or any other custom plugin can do that, and player does not need to know about all active plugins.

Code/EnginePlugins/PhysXPlugin/WorldModule/Implementation/PhysXWorldModule.cpp
Sanakan8472: Ok removed.

Code/Tools/Player/CMakeLists.txt
Sanakan8472: No unfortunately it can't. The finalizer is called in a different scope, calling find_package there does nothing to help resolve the generator expression at generation time. We need to call find_package at the scope of Code\Tools\Player\CMakeLists.txt so that the project itself can resolve the generator expression. At least I don't know how else I can fix that.

PR #212, opened 2020-05-10
OpenXR plugin

Sanakan8472:
* XR support
  * OpenXR plugin added
  * Moved companion window logic into XRWindow base class
  * Basic input support
* Various camera related fixes to support XR
* Renamed existing VR classes to XR and moved all XR interface code into GameState/XR
* Fixed deployment of release build dependent binaries. Both debug and release dlls are now source files and switch deployment depending on build target
* Removed particle node from HololensRenderPipeline as it is not supported
* Added ReplaceExisitingNativeObject for texture resources


Code/Engine/GameEngine/GameApplication/Implementation/GameApplicationInit.cpp
```diff
@@ -260,13 +260,15 @@ void ezGameApplication::Init_LoadRequiredPlugins()
...
   ezPlugin::LoadPlugin("ezRenderDocPlugin");
-#  endif
+#  endif 
 
   // on sandboxed platforms, we can only load data through fileserve, so enforce use of this plugin
 #  if EZ_DISABLED(EZ_SUPPORTS_UNRESTRICTED_FILE_ACCESS)
   ezPlugin::LoadPlugin("ezFileservePlugin"); // don't care if it fails to load
 #  endif
 
+  ezPlugin::LoadPlugin("ezTypeScriptPlugin");
```
jankrassnigg: This should not be here

Code/Engine/GameEngine/GameState/Implementation/GameState.cpp
```diff
@@ -132,33 +79,65 @@ void ezGameState::CreateActors()
...
-  pWindowPlugin->m_pWindow = std::move(pMainWindow);
-  pWindowPlugin->m_pWindowOutputTarget = std::move(pOutput);
-
-  SetupMainView(pWindowPlugin->m_pWindowOutputTarget.Borrow(), pWindowPlugin->m_pWindow->GetClientAreaSize());
+  // Init XR
+  const ezXRConfig* pConfig = ezGameApplicationBase::GetGameApplicationBaseInstance()->GetPlatformProfile().GetTypeConfig<ezXRConfig>();
+  ezXRInterface* pXRInterface = nullptr;
+  if (pConfig && pConfig->m_bEnableXR)
+  {
+    if (ezXRInterface* pXR = ezSingletonRegistry::GetRequiredSingletonInstance<ezXRInterface>())
```
jankrassnigg: I think the required version will assert that the interface exists.

Code/Engine/GameEngine/GameApplication/Implementation/GameApplicationInit.cpp
Sanakan8472: I thought so too but I can't get it to load otherwise in UWP. It is marked in the project manifest but won't be loaded without it. What would be the general code to load this? Could it be that the dynamic loading is somehow broken in UWP?

jankrassnigg: The project manifest would do exactly the same thing. This doesn't add a link dependency or anything, so if this works, but the project manifest doesn't work, then you should debug through this. Ie. remove this call, put a breakpoint in LoadPlugin, and check what happens when you go through the manifest. Maybe for some reason the call doesn't happen then.

Either way, this plugin is supposed to be optional, you should not add this line.

Code/EditorPlugins/Scene/EnginePluginScene/RenderPipeline/EditorSelectedObjectsExtractor.cpp
```diff
@@ -145,17 +145,25 @@ void ezEditorSelectedObjectsExtractor::CreateRenderTargetView(const ezView& view
...
+      m_RenderTargetCamera.SetCameraMode(
+        pCamComp->GetCameraMode(), pCamComp->GetOrthoDimension(), pCamComp->GetNearPlane(), pCamComp->GetFarPlane());
+      break;
+    case ezCameraMode::PerspectiveFixedFovX:
+    case ezCameraMode::PerspectiveFixedFovY:
+      m_RenderTargetCamera.SetCameraMode(
+        pCamComp->GetCameraMode(), pCamComp->GetFieldOfView(), pCamComp->GetNearPlane(), pCamComp->GetFarPlane());
+      break;
+    case ezCameraMode::Stereo:
+      m_RenderTargetCamera.SetCameraMode(ezCameraMode::PerspectiveFixedFovY, 45,
```
jankrassnigg, line 161: Why is the FOV hardcoded ?

Code/Engine/Foundation/Application/Implementation/uwp/ApplicationEntryPoint_uwp.h
```diff
@@ -29,7 +29,14 @@ namespace ezApplicationDetails
 
     AppClass* pApp = new (appBuffer) AppClass(std::forward<Args>(arguments)...);
 
+#ifdef BUILDSYSTEM_ENABLE_OPENXR_SUPPORT
+    //TODO: Using OpenXR will crash if we create a core window, will have to use standard run method
```
jankrassnigg, line 33: This looks hacky, and even if a hack is needed, couldn't it be moved into less low-level code? I.e. couldn't it at least be done in ezUWPRun instead? Worst case, you forward to ezRun in there, but then at least the function that you would expect to be called on uwp is indeed called.

Code/Engine/GameEngine/Configuration/Implementation/XRConfig.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+#include <GameEngine/Configuration/XRConfig.h>
+#include <Foundation/IO/ChunkStream.h>
+
+// clang-format off
+EZ_BEGIN_DYNAMIC_REFLECTED_TYPE(ezXRConfig, 2, ezRTTIDefaultAllocator<ezXRConfig>)
+{
+  EZ_BEGIN_PROPERTIES
+  {
+    EZ_MEMBER_PROPERTY("EnableXR", m_bEnableXR)->AddAttributes(new ezDefaultValueAttribute(false)),
```
jankrassnigg: You don't need a default value attribute for 'false'

Code/Engine/GameEngine/GameApplication/Implementation/GameApplicationInit.cpp
jankrassnigg: The manifest doesn't do anything different, it also just calls this function. If that doesn't work, you should check whether it is actually called correctly. Though PhysX and Fmod should be loaded the same way. Maybe the DLL name in the manifest is incorrect?

Either way, this would make the typescript plugin not optional, which it should be.

Code/Engine/GameEngine/XR/Declarations.h
```diff
@@ -0,0 +1,69 @@
+#pragma once
+
+#include <GameEngine/GameEngineDLL.h>
+#include <Foundation/Reflection/Reflection.h>
+
+struct ezXRDeviceType
+{
+  typedef ezUInt8 StorageType;
```
jankrassnigg: Prefer "using" over "typedef" for new code.

Code/Engine/GameEngine/XR/Implementation/DeviceTrackingComponent.cpp
```diff
@@ -0,0 +1,153 @@
...
+        vPosition = state.m_vAimPosition;
+        qRotation = state.m_qAimRotation;
+      }
+      else
+      {
+        return;
+      }
+
+
+      if (state.m_bGripPoseIsValid)
```
jankrassnigg: So the 'aim' location is also only applied when the grid pose is valid ?

Code/Engine/GameEngine/XR/XRInputDevice.h
```diff
@@ -0,0 +1,51 @@
...
+
+  /// \brief Fills out a list of valid (connected) device IDs.
+  virtual void GetDeviceList(ezHybridArray<ezXRDeviceID, 64>& out_Devices) const = 0;
+  /// \brief Returns the deviceID for a specific type of device.
+  /// If the device is not connected, -1 is returned instead.
+  virtual ezXRDeviceID GetDeviceIDByType(ezXRDeviceType::Enum type) const = 0;
+  /// \brief Returns the current device state for a valid device ID.
+  virtual const ezXRDeviceState& GetDeviceState(ezXRDeviceID iDeviceID) const = 0;
+
+  ezEventSubscriptionID AddEventHandler(ezXRDeviceEvent::Handler handler) { return m_InputEvents.AddEventHandler(handler); }
```
jankrassnigg: The better way to expose m_InputEvents is by having a function that returns it as a const reference. Add/RemoveEventHandler are mutable on ezEvent for this to work and you don't need to expose all the different functions then.

Code/Engine/GameEngine/XR/XRInterface.h
```diff
@@ -0,0 +1,93 @@
+#pragma once
```
jankrassnigg, line 1: This file probably should be under GameEngine/Interfaces. Unless you have more XR files in this lib.

Code/EnginePlugins/OpenXRPlugin/OpenXRStartup.cpp
```diff
@@ -0,0 +1,40 @@
...
+  {
+  }
+
+  ON_HIGHLEVELSYSTEMS_STARTUP
+  {
+  }
+
+  ON_HIGHLEVELSYSTEMS_SHUTDOWN
+  {
+  }
```
jankrassnigg: Not needed?

Code/Tools/Player/CMakeLists.txt
```diff
@@ -34,6 +34,11 @@ if (EZ_CMAKE_PLATFORM_WINDOWS_UWP)
     target_link_libraries(${PROJECT_NAME} PUBLIC FmodPlugin)
   endif()
 
+  if (EZ_BUILD_OPENXR)
```
jankrassnigg, line 37: Isn't this already covered by the finalize.cmake file in the plugin?

Code/Engine/GameEngine/XR/Implementation/XRWindow.cpp
```diff
@@ -0,0 +1,171 @@
...
+  if (m_pCompanionWindow)
+  {
+    m_pCompanionWindow->GetNativeWindowHandle();
+  }
+  return ezWindowHandle();
+}
+
+bool ezWindowXR::IsFullscreenWindow(bool bOnlyProperFullscreenMode) const
+{
+  //#TODO: return companion state or XR?
```
jankrassnigg: I would expose the companion window as a separate window somehow, so that you can specifically ask for that and modify it. The HMD will always be fullscreen, so just returning true is the right choice.

Code/Engine/GameEngine/XR/XRWindow.h
```diff
@@ -0,0 +1,66 @@
...
+
+  virtual ezSizeU32 GetClientAreaSize() const override;
+
+  virtual ezWindowHandle GetNativeWindowHandle() const override;
+
+  virtual bool IsFullscreenWindow(bool bOnlyProperFullscreenMode) const override;
+
+  virtual void ProcessWindowMessages() override;
+
+  ezXRInterface* m_pVrInterface = nullptr;
```
jankrassnigg, line 31: private ?

Code/Engine/GameEngine/XR/XRWindow.h
```diff
@@ -0,0 +1,66 @@
...
+  ~ezWindowOutputTargetXR();
+
+  virtual void Present(bool bEnableVSync) override;
+  virtual ezResult CaptureImage(ezImage& out_Image) override;
+
+  ezXRInterface* m_pXrInterface = nullptr;
+  ezUniquePtr<ezWindowOutputTargetBase> m_pCompanionWindowOutputTarget;
+  ezGALTextureHandle m_hCompanionRenderTarget;
+  ezConstantBufferStorageHandle m_hCompanionConstantBuffer;
+  ezShaderResourceHandle m_hCompanionShader;
```
jankrassnigg, line 54: private

Code/Engine/GameEngine/XR/XRWindow.h
jankrassnigg: + expose the companion window

Code/Engine/RendererFoundation/Resources/Implementation/ProxyTexture.cpp
```diff
@@ -36,6 +36,12 @@ ezResult ezGALProxyTexture::DeInitPlatform(ezGALDevice* pDevice)
   return EZ_SUCCESS;
 }
 
+ezResult ezGALProxyTexture::ReplaceExisitingNativeObject(void* pExisitingNativeObject)
+{
+  //TODO
```
jankrassnigg: assert not implemented ?

Code/EnginePlugins/OpenXRPlugin/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
...
+get_filename_component(PROJECT_NAME ${CMAKE_CURRENT_SOURCE_DIR} NAME_WE)
+
+ez_create_target(LIBRARY ${PROJECT_NAME})
+
+target_link_libraries(${PROJECT_NAME}
+  PUBLIC
+  GameEngine
+)
+
+target_compile_definitions(${PROJECT_NAME} PUBLIC BUILDSYSTEM_ENABLE_OPENXR_SUPPORT)
```
jankrassnigg, line 15: I think you can now remove BUILDSYSTEM_ENABLE_MIXEDREALITY_SUPPORT and all related code, no ?

jankrassnigg: Ie the whole WindowsMixedReality library.

Code/EnginePlugins/OpenXRPlugin/OpenXRDeclarations.h
```diff
@@ -0,0 +1,56 @@
+#pragma once
+
+#define XR_LOG(exp)                                                                                              \
+  do                                                                                                                   \
+  {                                                                                                                    \
+    auto s = (exp);                                                                                                   \
+    if (s != XR_SUCCESS)                                                                                               \
+    {                                                                                                                  \
+      ezLog::Error("OpenXR failure: {0}:{1} {2}", __FILE__, __LINE__, s);                                                     \
```
jankrassnigg: EZ_SOURCE_FILE, EZ_SOURCE_LINE (+EZ_SOURCE_FUNCTION)

Code/EnginePlugins/OpenXRPlugin/OpenXRDeclarations.h
```diff
@@ -0,0 +1,56 @@
...
+      return;           \
+    }                                                                                                                  \
+  } while (false)
+
+#define EZ_CHECK_XR(exp)                                                                                               \
+  {                                                                                                                    \
+    if (exp != XR_SUCCESS)                                                                                             \
+    {                                                                                                                  \
+      ezLog::Error("OpenXR failure: {0}:{1}", __FILE__, __LINE__);                                                     \
```
jankrassnigg: should also be called _LOG

Code/EnginePlugins/OpenXRPlugin/OpenXRDeclarations.h
```diff
@@ -0,0 +1,56 @@
...
+  do                                                                                                                   \
+  {                                                                                                                    \
+    auto s = (exp);                                                                                                   \
+    if (s != XR_SUCCESS)                                                                                               \
+    {                                                                                                                  \
+      ezLog::Error("OpenXR failure: {0}:{1} {2}", __FILE__, __LINE__, s);                                                     \
+    }                                                                                                                  \
+  } while (false)
+
+#define XR_CHECK_LOG(exp)                                                                                              \
```
jankrassnigg: Should be named SUCCEED_OR_RETURN in some way. Maybe check whether all these different macros are needed.

Code/EnginePlugins/OpenXRPlugin/init.cmake
```diff
@@ -0,0 +1,44 @@
...
+function(ez_link_target_openxr TARGET_NAME)
+
+	ez_requires_openxr()
+
+	find_package(ezOpenXR REQUIRED)
+
+	if (EZOPENXR_FOUND)
+		target_link_libraries(${TARGET_NAME} PRIVATE ezOpenXR::Loader)
+
+		# TODO: How to check whether the target even has a dll, e.g. is a dynamic build?
```
jankrassnigg: IIRC there is a property on the target that tells you whether it is a SHARED library.

jankrassnigg: https://cmake.org/cmake/help/latest/manual/cmake-properties.7.html#properties-on-targets

Not sure which property, but I think I've used something like that before.

Code/Tools/Player/Main.cpp
```diff
@@ -26,14 +26,26 @@ ezResult ezPlayerApplication::BeforeCoreSystemsStartup()
...
+  // For now it's hardcoded and needs to be compiled in and ezFileserve started with the project special dir set, e.g.:
+  // -specialdirs project ".../ezEngine/Data/Samples/Testing Chambers
+#if EZ_DISABLED(EZ_SUPPORTS_UNRESTRICTED_FILE_ACCESS)
+  if (m_sSceneFile.IsEmpty())
+  {
+    m_sSceneFile = ">project/AssetCache/Common/Scenes/Surfaces.ezObjectGraph";
+    m_sAppProjectPath = ">project";
+  }
+  else
+#else
```
jankrassnigg, line 39: I have to think about how we could do this properly.

Code/Engine/GameEngine/GameState/Implementation/GameState.cpp
```diff
@@ -167,42 +146,49 @@ void ezGameState::ConfigureInputActions() {}
...
+      pView->SetViewport(ezRectFloat(0.0f, 0.0f, (float)viewportSize.width, (float)viewportSize.height));
+    }
+  }
 }
 
-void ezGameState::SetupMainView(
-  ezWindowOutputTargetBase* pOutputTarget, ezSizeU32 viewportSize, ezTypedResourceHandle<ezRenderPipelineResource> hRenderPipeline)
+ezView* ezGameState::CreateMainView(ezTypedResourceHandle<ezRenderPipelineResource> hRenderPipeline)
 {
   EZ_LOG_BLOCK("SetupMainView");
```
C-Core: >SetupMainView [](start = 16, length = 13)

Should be CreateMainView

Code/Engine/RendererFoundation/Device/Implementation/Device.cpp
```diff
@@ -642,6 +642,46 @@ ezGALTextureHandle ezGALDevice::CreateTexture(
   return ezGALTextureHandle();
 }
 
+ezResult ezGALDevice::ReplaceExisitingNativeObject(ezGALTextureHandle hTexture, void* pExisitingNativeObject)
+{
+  ezGALTexture* pTexture = nullptr;
+  if (m_Textures.TryGetValue(hTexture, pTexture))
+  {
+    for (auto it = pTexture->m_RenderTargetViews.GetIterator(); it.IsValid(); ++it)
```
C-Core, line 659: You need to do the same for m_ResourceViews and m_UnorderedAccessViews

Code/Engine/RendererFoundation/Device/Implementation/Device.cpp
```diff
@@ -642,6 +642,46 @@ ezGALTextureHandle ezGALDevice::CreateTexture(
...
+      {
+        EZ_VERIFY(pRenderTargetView->InitPlatform(this).Succeeded(),
+          "InitPlatform failed on a texture the previously succeded in the same call, is the new native object valid?");
+      }
+    }
+    return EZ_SUCCESS;
+  }
+  else
+  {
+    ezLog::Warning("DestroyTexture called on invalid handle (double free?)");
```
C-Core: >DestroyTexture [](start = 20, length = 14)

should be ReplaceExisitingNativeObject and remove the (double free?) at the end

Code/EnginePlugins/OpenXRPlugin/OpenXRInputDevice.cpp
```diff
@@ -0,0 +1,348 @@
...
+}
+
+XrResult ezOpenXRInputDevice::CreateAction(const char* actionName, XrActionType actionType, XrAction& out_action)
+{
+  XrActionCreateInfo actionCreateInfo{ XR_TYPE_ACTION_CREATE_INFO };
+  actionCreateInfo.actionType = actionType;
+  actionCreateInfo.countSubactionPaths = m_subActionPath.GetCount();
+  actionCreateInfo.subactionPaths = m_subActionPath.GetData();
+  strcpy_s(actionCreateInfo.actionName, actionName);
+  strcpy_s(actionCreateInfo.localizedActionName, actionName);
```
jankrassnigg: ezStringUtils

Code/EnginePlugins/OpenXRPlugin/OpenXRInputDevice.cpp
```diff
@@ -0,0 +1,348 @@
...
+  switch (actionType)
+  {
+    case XR_ACTION_TYPE_BOOLEAN_INPUT:
+      m_booleanActions.PushBack({ out_action, sLeft, sRight });
+      break;
+    case XR_ACTION_TYPE_POSE_INPUT:
+      m_poseActions.PushBack({ out_action, sLeft, sRight });
+      break;
+    default:
+      EZ_REPORT_FAILURE("NOt implemented");
```
jankrassnigg: Type. EZ_ASSERT_NOT_IMPL

Code/EnginePlugins/OpenXRPlugin/OpenXRInputDevice.cpp
```diff
@@ -0,0 +1,348 @@
...
+
+ezResult ezOpenXRInputDevice::SuggestInteractionProfileBindings(const char* szInteractionProfile, ezArrayPtr<Bind> bindings)
+{
+  XrInstance instance = m_pOpenXR->m_instance;
+
+  XrPath InteractionProfile = CreatePath(szInteractionProfile);
+
+  ezDynamicArray<XrActionSuggestedBinding> xrBindings;
+  xrBindings.Reserve(bindings.GetCount());
+  for (auto& binding : bindings)
```
jankrassnigg: const. Probably make the binding arrayptr an ezArrayPtr<const Bind>

Code/EnginePlugins/OpenXRPlugin/OpenXRInputDevice.cpp
```diff
@@ -0,0 +1,348 @@
...
+    }
+    else
+    {
+      m_bIsValid = false;
+    }
+  };
+
+  for (ezUInt32 uiSide : {0, 1})
+  {
+    ezXRDeviceState& state = m_DeviceState[uiSide + 1];
```
jankrassnigg: why +1 ?

Code/EnginePlugins/OpenXRPlugin/OpenXRSingleton.h
```diff
@@ -0,0 +1,207 @@
...
+
+class ezOpenXRInputDevice;
+
+struct ezGameApplicationExecutionEvent;
+
+EZ_DEFINE_AS_POD_TYPE(XrSwapchainImageD3D11KHR);
+EZ_DEFINE_AS_POD_TYPE(XrViewConfigurationView);
+EZ_DEFINE_AS_POD_TYPE(XrEnvironmentBlendMode);
+
+//#TODO: Make use of this to abstract render api.
```
jankrassnigg: This isn't used yet? @C-Core

Code/EnginePlugins/OpenXRPlugin/OpenXRSingleton.cpp
```diff
@@ -0,0 +1,1122 @@
...
+
+#include <Core/World/World.h>
+#include <GameEngine/XR/XRWindow.h>
+#include <GameEngine/XR/StageSpaceComponent.h>
+#include <GameEngine/ActorSystem/Actor.h>
+#include <RendererCore/Shader/ShaderResource.h>
+#include <algorithm>
+#include <vector>
+
+#pragma optimize("", off)
```
jankrassnigg: ?

Code/EnginePlugins/OpenXRPlugin/OpenXRSingleton.cpp
```diff
@@ -0,0 +1,1122 @@
...
+    desc.m_bDoubleBuffered = true;
+    desc.m_pWindow = nullptr;
+    desc.m_SampleCount = ConvertMSAACount(0);
+    ezGALOpenXRSwapChainDX11* pSwapChain = EZ_DEFAULT_NEW(ezGALOpenXRSwapChainDX11, desc);
+    if (!pSwapChain->InitPlatform(pDevice).Succeeded())
+    {
+      EZ_DEFAULT_DELETE(pSwapChain);
+      return nullptr;
+    }
+  }*/
```
jankrassnigg: remove

Code/EnginePlugins/OpenXRPlugin/OpenXRSingleton.cpp
```diff
@@ -0,0 +1,1122 @@
...
+    {
+      const ezMat4 mStageTransform = add.GetAsMat4();
+      const ezMat4 poseLeft = mStageTransform * ConvertPoseToMatrix(m_views[0].pose);
+      const ezMat4 poseRight = mStageTransform * ConvertPoseToMatrix(m_views[1].pose);
+
+      // EZ Forward is +X, need to add this to align the forward projection
+      const ezMat4 viewMatrix =
+        ezGraphicsUtils::CreateLookAtViewMatrix(ezVec3::ZeroVector(), ezVec3(1, 0, 0), ezVec3(0, 0, 1));
+      const ezMat4 mViewTransformLeft = viewMatrix * poseLeft.GetInverse();
+      const ezMat4 mViewTransformRight = viewMatrix * poseRight.GetInverse();
```
jankrassnigg, line 823: This may depend on the coordinate system configuration, which is set in the ezWorld. @C-Core should this be handled here?

Code/EnginePlugins/OpenXRPlugin/OpenXRSingleton.cpp
```diff
@@ -0,0 +1,1122 @@
...
+  }
+
+  {
+    // put the camera orientation into the sound listener and enable the listener override mode
+    if (ezSoundInterface* pSoundInterface = ezSingletonRegistry::GetSingletonInstance<ezSoundInterface>())
+    {
+      pSoundInterface->SetListener(-1, m_pCameraToSynchronize->GetCenterPosition(),
+        m_pCameraToSynchronize->GetCenterDirForwards(), m_pCameraToSynchronize->GetCenterDirUp(), ezVec3::ZeroVector());
+    }
+  }
```
jankrassnigg: Override mode is for the editor. It would be better to have a head-pose component that can be attached to the player prefab, because that's where you typically have the listener component.

Code/EnginePlugins/OpenXRPlugin/OpenXRSingleton.cpp
```diff
@@ -0,0 +1,1122 @@
...
+    default:
+      EZ_REPORT_FAILURE("Unknown format {}", format);
+      break;
+  }
+  return ezGALResourceFormat::Invalid;
+}
+
+ezGALMSAASampleCount::Enum ezOpenXR::ConvertMSAACount(ezUInt32 count)
+{
+  switch (count)
```
jankrassnigg: The enums have the right values, you can literally do an int cast.

Code/EnginePlugins/OpenXRPlugin/OpenXRSingleton.cpp
```diff
@@ -0,0 +1,1122 @@
...
+  ezMat4 m;
+  ezMat3 rot = ConvertOrientation(pose.orientation).GetAsMat3();
+  ezVec3 pos = ConvertPosition(pose.position);
+  m.SetTransformationMatrix(rot, pos);
+  return m;
+}
+
+ezGALResourceFormat::Enum ezOpenXR::ConvertTextureFormat(int64_t format)
+{
+  switch (format)
```
jankrassnigg, line 1045: We should remove ezGALResourceFormat at some point and just use ezImageFormat everywhere instead.
(@marcappelsmeier)

You can use ezImageFormatMappings to map between DXGI and ezImageFormat.

You can then use ezTextureUtils to map from ezImageFormat to ezGALResourceFormat.

Code/EnginePlugins/OpenXRPlugin/OpenXRSingleton.cpp
C-Core: Yes, if you want to make this future proof, get the coordinate system provider from the world to obtain the forward axis etc

---
In reply to: [423524360](https://github.com/ezEngine/ezEngine/pull/212#discussion_r423524360) [](ancestors = 423524360)

Code/EnginePlugins/OpenXRPlugin/OpenXRSingleton.h
C-Core: I guess this was the first try to get the native d3d texture into our render pipeline. I haven't looked at the OpenXR implementation in detail by this class is most likely not needed anymore.

---
In reply to: [423519047](https://github.com/ezEngine/ezEngine/pull/212#discussion_r423519047) [](ancestors = 423519047)

Code/BuildSystem/CMake/FindEzOpenXR.cmake
```diff
@@ -0,0 +1,88 @@
...
+ez_pull_architecture_vars()
+
+if ((EZ_OPENXR_DIR STREQUAL "EZ_OPENXR_DIR-NOTFOUND") OR (EZ_OPENXR_DIR STREQUAL ""))
+	ez_nuget_init()
+	execute_process(COMMAND ${NUGET} restore ${CMAKE_SOURCE_DIR}/Code/EnginePlugins/OpenXRPlugin/packages.config -PackagesDirectory ${CMAKE_BINARY_DIR}/packages
+	WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
+	set (EZ_OPENXR_DIR "${CMAKE_BINARY_DIR}/packages/OpenXR.Loader.1.0.6.2" CACHE PATH "Directory of OpenXR runtime installation" FORCE)
+endif()
+
+if (CMAKE_SYSTEM_NAME STREQUAL "WindowsStore")
```
Ingrater: please use EZ_CMAKE_PLATFORM_WINDOWS_UWP instead

Code/BuildSystem/CMake/FindEzOpenXR.cmake
```diff
@@ -0,0 +1,88 @@
...
+	find_path(EZ_OPENXR_DIR include/openxr/openxr.h)
+
+	if (EZ_CMAKE_ARCHITECTURE_ARM)
+		if(CMAKE_SIZEOF_VOID_P EQUAL 8)
+			set(OPENXR_BIN_PREFIX "arm64_uwp")
+		else()
+			set(OPENXR_BIN_PREFIX "arm_uwp")
+		endif()
+	else()
+		if(CMAKE_SIZEOF_VOID_P EQUAL 8)
```
Ingrater: please use EZ_CMAKE_ARCHITECTURE_64BIT instead

Code/BuildSystem/CMake/FindEzOpenXR.cmake
```diff
@@ -0,0 +1,88 @@
...
+	set (EZ_OPENXR_DIR "${CMAKE_BINARY_DIR}/packages/OpenXR.Loader.1.0.6.2" CACHE PATH "Directory of OpenXR runtime installation" FORCE)
+endif()
+
+if (CMAKE_SYSTEM_NAME STREQUAL "WindowsStore")
+	# UWP builds
+	set (OPENXR_DYNAMIC ON)
+	find_path(EZ_OPENXR_DIR include/openxr/openxr.h)
+
+	if (EZ_CMAKE_ARCHITECTURE_ARM)
+		if(CMAKE_SIZEOF_VOID_P EQUAL 8)
```
Ingrater: please use EZ_CMAKE_ARCHITECTURE_64BIT instead

Code/BuildSystem/CMake/FindEzOpenXR.cmake
```diff
@@ -0,0 +1,88 @@
...
+		endif()
+	else()
+		if(CMAKE_SIZEOF_VOID_P EQUAL 8)
+			set(OPENXR_BIN_PREFIX "x64_uwp")
+		else()
+			set(OPENXR_BIN_PREFIX "Win32_uwp")
+		endif()
+	endif()
+
+elseif (CMAKE_SYSTEM_NAME STREQUAL "Windows")
```
Ingrater: please use EZ_CMAKE_PLATFORM_WINDOWS

Code/BuildSystem/CMake/FindEzOpenXR.cmake
```diff
@@ -0,0 +1,88 @@
...
+	# Desktop Windows builds
+
+	set (OPENXR_DYNAMIC ON)
+	find_path(EZ_OPENXR_DIR include/openxr/openxr.h)
+
+	if(CMAKE_SIZEOF_VOID_P EQUAL 8)
```
Ingrater: please use EZ_CMAKE_ARCHITECTURE_64BIT

Code/BuildSystem/CMake/FindEzOpenXR.cmake
```diff
@@ -0,0 +1,88 @@
...
+if (TARGET ezOpenXR::Loader)
+	return()
+endif()
+
+set (EZ_OPENXR_DIR "EZ_OPENXR_DIR-NOTFOUND" CACHE PATH "Directory of OpenXR runtime installation")
+set (EZ_OPENXR_PREVIEW_DIR "" CACHE PATH "Directory of OpenXR preview include root")
+mark_as_advanced(FORCE EZ_OPENXR_DIR)
+mark_as_advanced(FORCE EZ_OPENXR_PREVIEW_DIR)
+
+ez_pull_architecture_vars()
```
Ingrater: does not longer exist, use ez_pull_compiler_and_architecture_vars instead.

HeadClot:
Got a question for @Sanakan8472 Will this plugin support OpenVR? Just curious.

Code/Engine/GameEngine/GameState/Implementation/GameState.cpp
Sanakan8472: Nice find, I didn't knew there were two different functions.

Code/EditorPlugins/Scene/EnginePluginScene/RenderPipeline/EditorSelectedObjectsExtractor.cpp
Sanakan8472: This is for the preview in the editor. The problem is that the ezCamera has projection matrices but the ezCameraComponent does not so even if the mode is set on a ezCameraComponent to Stereo it has no data to back it up. Not quite sure how this should be fixed so for the time being I hardcoded the preview camera data as the ezCameraComponent cannot provide full stereo data.

Code/Engine/GameEngine/Configuration/Implementation/XRConfig.cpp
Sanakan8472: Right!

Code/Engine/GameEngine/XR/Declarations.h
Sanakan8472: Oh I didn't know I could use those interchangeable for use in the ezEnum template.

Code/Engine/GameEngine/XR/Implementation/DeviceTrackingComponent.cpp
Sanakan8472: Good catch, this code is wrong. We can only get here if we found a valid pose so there shouldn't be any check here at all.

Code/Engine/GameEngine/XR/XRInterface.h
Sanakan8472: I had it there before but it exploded over time and it will only get worse so I thought I'd just put all of them under XR. Also most of these interface will have shared base implementations anyways so they aren't pure interfaces to begin with. At the moment we have ezXRInterface, ezXRInputDevice, ezWindowXR, ezWindowOutputTargetXR, ezActorPluginWindowXR all with varying amount of base code already. Additionally I will need to add an anchor interface to support anchor components and spatial mapping support.

Code/EnginePlugins/OpenXRPlugin/OpenXRStartup.cpp
Sanakan8472: Yep and I can't remember why I added it in the first place.

Code/Engine/GameEngine/XR/Implementation/XRWindow.cpp
Sanakan8472: So timing in XR is key and I need full control on when the window is rendered. If both the XR and the companion window are handled by separate ezActors I lose control on the order of execution for timing and dependency purposes (I can only render the companion view as long as I have rendered into the XR texture and am still holding onto that texture.
So I opted for wrapping the companion window into the XR window but that obviously causes some ambiguity. Returning the XR state should be the correct choice in most cases though.

Code/Engine/RendererFoundation/Resources/Implementation/ProxyTexture.cpp
Sanakan8472: @C-Core I have actually no idea what ezGALProxyTexture does and what I should put there.

Code/EnginePlugins/OpenXRPlugin/CMakeLists.txt
Sanakan8472: Yes I can but will do that in a separate commit to not bloat this one even further.

Sanakan8472:
@HeadClot this PR will only add OpenXR as an implementation to the XR interface but the same interface will support OpenVR in the future. There is still old code in the repo for OpenVR support that I'll get back to a working state again soon as iteration times are much better in OpenVR compared to OpenXR when using a Windows Mixed Reality HMD.

Code/Engine/Foundation/Application/Implementation/uwp/ApplicationEntryPoint_uwp.h
Sanakan8472: Oh it is hacky for sure. Unfortunately there is currently no other way to get OpenXR working. I'll try moving it into ezUWPRun.

Code/Engine/GameEngine/XR/XRInputDevice.h
Sanakan8472: Oh that's nicer.

Code/Tools/Player/CMakeLists.txt
Sanakan8472: The problem is that to be able to call target_link_libraries against these plugins I need to be able to resolve their generator expressions which requires calling find_package again which does not work in the finalize function as it is not cause the project to be able to resolve the generator expression probably because it's called in a different scope. I haven't found a way around this so far.

Code/EnginePlugins/OpenXRPlugin/OpenXRDeclarations.h
Sanakan8472: Will do.

Code/EnginePlugins/OpenXRPlugin/OpenXRDeclarations.h
Sanakan8472: Cleaning up log functions a bit.

Code/EnginePlugins/OpenXRPlugin/OpenXRDeclarations.h
Sanakan8472: Cleaning up log functions a bit.

Code/EnginePlugins/OpenXRPlugin/init.cmake
Sanakan8472: Actually used this in another place already to do the deployment of the dlls. This works:
get_target_property(_dll_location ezOpenXR::Loader IMPORTED_LOCATION)

Code/Tools/Player/Main.cpp
Sanakan8472: Couldn't we move this problem to the fileserve side? Make the fileserve server decide which scene to load? That way we could switch projects and scenes without the need to change anything on the player.

Code/Engine/RendererFoundation/Resources/Implementation/ProxyTexture.cpp
C-Core: ezGALProxyTexture allows you to treat an arrayslice or cubemap face as a regular texture2D. This is useful if you want to e.g. render one face of a reflection probe but the pipeline does only support rendering into regular texture2D. I can't think of a use case why you want to replace the native object on such a texture so assert not implemented is fine for now.

Code/Engine/RendererFoundation/Device/Implementation/Device.cpp
Sanakan8472: Will do!

Code/Engine/RendererFoundation/Device/Implementation/Device.cpp
Sanakan8472: Copy paste error :)

Code/Engine/RendererFoundation/Resources/Implementation/ProxyTexture.cpp
Sanakan8472: Thanks for the info! Will add not implemented assert.

Code/EnginePlugins/OpenXRPlugin/OpenXRInputDevice.cpp
Sanakan8472: 0 is head, 1 is left hand, 2 is right hand. Replaced the code with a more readable version.

Code/EnginePlugins/OpenXRPlugin/OpenXRSingleton.cpp
Sanakan8472: woopsi!

Code/EnginePlugins/OpenXRPlugin/OpenXRSingleton.cpp
Sanakan8472: Ok I'll put that on my list of things to improve. Too big of a task to do in this PR but might be a good idea to do that before reviving the OpenVR plugin,

Code/Engine/GameEngine/GameApplication/Implementation/GameApplicationInit.cpp
Sanakan8472: Found the problem: The typescript plugin was not selected to be loaded for the project but the editor loads it regardless so it worked in the editor.

Code/EnginePlugins/OpenXRPlugin/OpenXRSingleton.h
Sanakan8472: Removing it.

Code/Tools/Player/Main.cpp
jankrassnigg: I was thinking of doing this via a cvar in the player, which stores the last used scene file and can be used as a fallback for the next time.

What we would then need to change, is that the player doesn't assert and shutdown when nothing is given, but just renders an empty scene.

Then you could launch the player, have an empty scene, connect with ezInspector, modify the cvar, restart ezPlayer, and have it automatically load the "last scene" for you.

At some point I also want to add that you can type console commands into ezInspector, then we could literally have a "LoadScene" command and execute that in ezPlayer through ezInspector.

Actually, even now you could already listen to changes to the "last scene" cvar and already react when the inspector modifies it.

jankrassnigg: This would not require a change to FileServe and would be a more general solution.

Code/EnginePlugins/OpenXRPlugin/OpenXRSingleton.cpp
Sanakan8472: Ok works now without it thanks to the listener component.

Code/EnginePlugins/OpenXRPlugin/OpenXRSingleton.cpp
Sanakan8472: Ok will replace it with static assert that this enum remains that way.

Code/Engine/Foundation/Application/Implementation/uwp/ApplicationEntryPoint_uwp.h
Sanakan8472: Ok I can't move this into ezUWPRun as that would hardcode it in foundation for all projects which has serious side effects. With the current solution it only affects projects that link against OpenXR.

Code/Engine/GameEngine/XR/XRWindow.h
Sanakan8472: Done!

Code/Tools/Player/Main.cpp
Sanakan8472: Yep that could work. Some goal I have is to have a better editor integration. I'll see first if I can get the EditorEngineProcess working on Hololens again which probably would be the best in terms of iterations time. Alternatively I'd like to be able to havea one-click solution to against the player on a device which would require setting the scene externally either by fileserve or inspector API.

Code/EnginePlugins/OpenXRPlugin/OpenXRSingleton.cpp
Sanakan8472: I'll use that now for the color formats but I still need the switch case for the depth format as they are not in the image format enum.

Code/BuildSystem/CMake/FindEzOpenXR.cmake
Sanakan8472: Thanks for the tip, rebasing now.

Code/Engine/RendererFoundation/Device/Implementation/Device.cpp
```diff
@@ -642,6 +642,64 @@ ezGALTextureHandle ezGALDevice::CreateTexture(
...
+      }
+    }
+
+    EZ_VERIFY(pTexture->DeInitPlatform(this).Succeeded(), "DeInitPlatform should never fail.");
+    EZ_VERIFY(pTexture->ReplaceExisitingNativeObject(pExisitingNativeObject).Succeeded(),
+      "Failed to replace native texture, make sure the input is valid.");
+    EZ_VERIFY(pTexture->InitPlatform(this, {}).Succeeded(),
+      "InitPlatform failed on a texture the previously succeded in the same call, is the new native object valid?");
+    
+    for (auto it = pTexture->m_RenderTargetViews.GetIterator(); it.IsValid(); ++it)
```
C-Core, line 694: You should of course also re-init resourceViews and UnorderedAccessViews here ;)

Code/EditorPlugins/Scene/EnginePluginScene/RenderPipeline/EditorSelectedObjectsExtractor.cpp
jankrassnigg: Hm, I guess I understand the matrix stuff, but here specifically you hardcode the FOV, although the camera component does have a FOV value that you could put in. Maybe I don't really get it after all. Could discuss this in detail offline, I guess. For the time being you can also create an issue to not forget about this (unless it's so obvious for you, that you think that's not necessary).

Code/Engine/Foundation/Application/Implementation/uwp/ApplicationEntryPoint_uwp.h
jankrassnigg: What I dislike about this is that it means that you need to set a compile switch to get XR working, which means you can't just enable this functionality by enabling the XR plugin in the editor. Again, we should probably discuss the details offline.

Code/Tools/Player/CMakeLists.txt
jankrassnigg: Yeah, CMake is another technology that fails at solving the problems it is meant to solve. Let's just ignore this for the time being.

Code/Tools/Player/Main.cpp
jankrassnigg: Would be happy to brainstorm about this some more. There are probably multiple solutions with different pros and cons. Not super high priority though.

Code/EnginePlugins/OpenXRPlugin/OpenXRSingleton.cpp
jankrassnigg: Frankly, I wouldn't mind ignoring this for the time being. Was just pointing it out, but in practice it's not a problem.

Code/EnginePlugins/OpenXRPlugin/OpenXRSingleton.cpp
jankrassnigg: Okay, maybe it would be best to extend that then? I guess it's not in there, because a DDS file can't use the format. Again, not high priority, but we seem to have a proliferation of custom image format handling code.

Code/Engine/RendererFoundation/Device/Implementation/Device.cpp
Sanakan8472: And I was about to click the submit button! Thanks for looking it over. Will fix it today.

PR #227, opened 2020-05-14
Run clang-tidy on Foundation

Ingrater:
Run clang-tidy modernize-use-using, modernize-use-equals-default, modernize-deprecated-headers, modernize-use-nullptr on Foundation


C-Core (changes requested):
I don't like the replacement to use auto, if people like to specify the full type they should be able to do it, so auto is completely optional.
Also I don't see any value in replacing the typedefs with using since it is exactly the same.
=default and the header change are fine by me.

jankrassnigg:
> I don't like the replacement to use auto, if people like to specify the full type they should be able to do it, so auto is completely optional.

I agree with that.

> I don't see any value in replacing the typedefs with using since it is exactly the same.

I disagree, "using" is better readable and it is necessary when you want to have a templated typedef, so for consistency alone it's better to use one style and not two.

> =default and the header change are fine by me.

Me too.

Ingrater:
> 
> 
> I don't like the replacement to use auto, if people like to specify the full type they should be able to do it, so auto is completely optional.
> Also I don't see any value in replacing the typedefs with using since it is exactly the same.
> =default and the header change are fine by me.

The auto changes are debateable. They are only used in sutations where the type is clear from the right hand side expression anyway. It doesn't auto replace everything with auto.
I do like the typedef to using replacement because its better readable. Everytime I see a typedef I have to think about which way around it was. The using syntax is more intuitive in my opinion.

jankrassnigg (changes requested):
Don't do the auto replacement, please.

Code/Engine/Texture/Image/Formats/TgaFileFormat.cpp
```diff
@@ -156,7 +156,7 @@ ezResult ezTgaFileFormat::WriteImage(ezStreamWriter& stream, const ezImageView&
 
     ezInt32 iRLE = 0;
 
-    ezColorLinearUB pc;
+    ezColorLinearUB pc = {};
```
jankrassnigg, line 159: This seems unnecessary.

jankrassnigg (approved):

Code/Engine/Texture/Image/Formats/TgaFileFormat.cpp
Ingrater: It is not. Because the constructor of ezColorLinearUB  is now = default the compiler can now detect usage of uninitialized ezColorLinearUB instances. "= {}" initializes this instace of ezColorLinearUB to zero and gets rid of the compiler warning.

Code/Engine/Foundation/Math/Color8UNorm.h
```diff
@@ -19,7 +19,8 @@ class EZ_FOUNDATION_DLL ezColorBaseUB
   ezUInt8 a;
 
   /// \brief Default-constructed color is uninitialized (for speed)
-  ezColorBaseUB(){};
+  ezColorBaseUB() = default;
+  ;
```
C-Core: extra semicolon, because why not :P

Code/Engine/Foundation/Math/Color8UNorm.h
```diff
@@ -72,7 +73,8 @@ class EZ_FOUNDATION_DLL ezColorGammaUB : public ezColorBaseUB
   EZ_DECLARE_POD_TYPE();
 
   /// \brief Default-constructed color is uninitialized (for speed)
-  ezColorGammaUB(){};
+  ezColorGammaUB() = default;
+  ;
```
C-Core: here as well

C-Core (approved):

PR #235, opened 2020-05-20
Adds textured triangle support to the debug renderer

jankrassnigg:
Also fixes a memory leak and adds sample code to the sample game plugin.


Code/Engine/RendererCore/Debug/Implementation/DebugRenderer.cpp
```diff
@@ -102,7 +103,7 @@ namespace
     }
 
     ezUInt64 m_uiLastRenderedFrame;
-    PerContextData* m_pData[2];
+    ezUniquePtr<PerContextData> m_pData[2];
```
jankrassnigg, line 106: This was a memory leak, and when you used one of the 'textured' calls (e.g. the textured Draw2DRectangle variant), D3D would complain about live GPU objects at shutdown.

Code/Engine/RendererCore/Debug/Implementation/DebugRenderer.cpp
```diff
@@ -386,7 +388,7 @@ void ezDebugRenderer::Draw2DLines(const ezDebugRendererContext& context, ezArray
 }
 
 //static
-void ezDebugRenderer::DrawCross(const ezDebugRendererContext& context, const ezVec3& globalPosition, float fLineLength, const ezColor& color)
+void ezDebugRenderer::DrawCross(const ezDebugRendererContext& context, const ezVec3& globalPosition, float fLineLength, const ezColor& color, const ezTransform& transform /*= ezTransform::IdentityTransform()*/)
```
jankrassnigg, line 391: added an overall transform option

jankrassnigg:
Related new documentation: http://ezengine.net/debugging/debug-rendering.html

Code/Engine/RendererCore/Debug/Implementation/DebugRenderer.cpp
C-Core: Indeed, good catch!

Code/Engine/RendererCore/Debug/Implementation/DebugRenderer.cpp
C-Core: makes sense

C-Core (approved):

PR #247, opened 2020-05-24
https://github.com/ezEngine/ezEngine/projects/5#card-38317759

### LINKED ISSUE #228: Open in vscode should always work
VS code can be installed in two ways: either in the user directory, or for all users in the Program Files directory.

To launch vsc we currently only support the user directory method.
Editor: *Tools > Open in Visual Studio Code*

If installed differently, it should work too. It is unclear how to do that, though, AFAIK there is no registry key for where vsc is installed. But there may be icon or file associations stored in the registry.
Some research is required to figure this out.

zerowil:
#228 opened by jankrassnigg
The Editor now obtains the Visual Studio code path from the registry.

### DIFF
--- a/Code/Tools/Libs/GuiFoundation/UIServices/Implementation/UIServices.cpp
+++ b/Code/Tools/Libs/GuiFoundation/UIServices/Implementation/UIServices.cpp
@@ -220,7 +220,15 @@ void ezQtUiServices::OpenInExplorer(const char* szPath, bool bIsFile)
 
 ezStatus ezQtUiServices::OpenInVsCode(const QStringList& arguments)
 {
-  QString sVsCodeExe = QStandardPaths::locate(QStandardPaths::GenericDataLocation, "Programs/Microsoft VS Code/Code.exe", QStandardPaths::LocateOption::LocateFile);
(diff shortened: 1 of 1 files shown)


jankrassnigg:
@zerowil thanks for the fix! We tested it on two machines, one where VS code could not be opened currently, and one where it already worked. On the first machine the fix actually fixed the problem, on the second machine it broke opening vsc.

So a proper fix should do both code paths, the old and the new. Could you change your fix such that it tries the registry key, and if that doesn't result in a valid path to vsc, it tries to do what it did so far? That should cover all cases then.

Cheers!

zerowil:
>@zerowil thanks for the fix! We tested it on two machines, one where VS code could not be opened currently, and one where it already worked. On the first machine the fix actually fixed the problem, on the second machine it broke opening vsc.
>
>So a proper fix should do both code paths, the old and the new. Could you change your fix such that it tries the registry key, and if that doesn't result in a valid path to vsc, it tries to do what it did so far? That should cover all cases then.
>
>Cheers!


Certainly @jankrassnigg !

zerowil:
The solution did not meet all cases, so the solution that was already working at a given time was also maintained.

jankrassnigg:
Looks good, thanks a lot!

PR #265, opened 2020-06-07
Adding more OpenXR features

### LINKED ISSUE #240: Add coordinate system support to OpenXR
At the moment the default EZ coordinate system is assumed. This functionality should be extracted out of the plugin to be able to re-use and test it.

### LINKED ISSUE #241: Add advanced input to XR interface / OpenXR plugin
Only poses and click is supported atm.

- Needs a mapping for each supported controller to a general set of actions as well as a query API to test if certain features are supported.
- Articulated hand tracking

### LINKED ISSUE #244: Add anchor support to XR interface
ezSpatialAnchorComponent should be rerouted to use the XR interface instead of being WMR dependent.

Sanakan8472:
- Added ezCoordinateSystemConversion class. Allows for converting between two coordinate systems (RH, LH, uniformly scaled).
- Changed ezSpatialAnchorComponent to use the new ezXRSpatialAnchorsInterface (persistence not supported so removed for now).
- Added ezXRHandTrackingInterface. Returns an array of bones for each hand part with the first index being the tip and the last index always being the wrist. Number of bones in between is implementation defined.
  - Added ezVisualizeHandComponent to visualize the bones returned by the above interface.
- Added support for advanced input methods to ezXRInputDevice. All features of motion controllers are supported except for vibration now.
- Add stereo rendering support to debug rendering.
  - Moved MaterialStereoGeometryShader.h to StereoGeometryShader.h to make it reusable without forcing USE_WORLDPOS.
  - Added line drawing support.
- Natvis fix for ezStaticArray
Resolves #244, resolves #241, resolves #240


Sanakan8472:
As all the OpenXR coordinate are in local space I actually don't need to do anything to support ezCoordinateSystemProvider. Just Aligning the stage space to the coordinate system will do the trick. Thus the new coordinate system class isn't actually needed right now. I could use it to convert EZ<->OpenXR but it would be a bit slower as it is generalized compared to the simple swizzle code right now.

Data/Base/Shaders/Materials/StereoGeometryShader.h
```diff
@@ -0,0 +1,114 @@
+#pragma once
```
C-Core, line 1: formatting is pretty wild for this file

C-Core (approved):

Data/Base/Shaders/Materials/StereoGeometryShader.h
Sanakan8472: clang format to the rescue!

PR #267, opened 2020-06-10
Format everything using clang-format

Ingrater:
* Third party code is not touched by clang-format. This is done by adding per folder .clang-format files which cause clang-format to ignore the files.
* First a run has been done with ColumnLimit: 1000 to get long lines back (as suggested by @jankrassnigg). Then a second run with ColumnLimit: 0 has been done.

Looking for feedback if the run with ColumnLimit: 1000 is worth it.
Also pay attention to the resorting of header files.
I am aware that the code currently does not build (due to header file re-sorting)


jankrassnigg (approved):
Looks good!

C-Core (approved):
I like it. Are we going to enforce clang format in the future? Otherwise this will degrade pretty quickly.

Sanakan8472:
I suppose we can extend CI to check for formatting?
I'm not a big fan of very long lines and to be fair I have no idea what effect column count 1000 has vs 0. My question would be is the column count reverseable? Can I set it locally to 180 or something and before checkin flatten everything again to unreadable long lines?

jankrassnigg:
@Sanakan8472 I prefer to use VS for that: Edit > Advanced > Word Wrap

That means you get lines wrapped according to your preferred screen size. That scales much better to big and small screens.

jankrassnigg:
Column count 1000 just means to force everything (back) on a line that is one statement. This reversed where we previously automatically inserted line breaks. Column count 0 means that from now on, Clang-format will not change line breaks, neither adding nor removing line breaks. That means every manual line break that you type going forward, will not be removed by clang-format. This is actually what we allowed for a while now, it's the least annoying setting.

jankrassnigg:
Having a different clang-format config locally is theoretically possible, but really a bad idea.

Sanakan8472:
Oh I didn't know this setting existed, thanks @jankrassnigg . I don't have any complaints then. Regarding the includes we can prevent the sorting messing the ordering by adding a newline right? So it's just a matter of adding some manual separations to fix the compile issues, right?

Ingrater:
> 
> 
> I like it. Are we going to enforce clang format in the future? Otherwise this will degrade pretty quickly.

Yes, the idea is to enforce this via CI. The PR already contains the CI script.

Ingrater:
> 
> 
> Oh I didn't know this setting existed, thanks @jankrassnigg . I don't have any complaints then. Regarding the includes we can prevent the sorting messing the ordering by adding a newline right? So it's just a matter of adding some manual separations to fix the compile issues, right?

Yes, adding a newline will prevent ordering. clang-format only sorts the includes in consecutive blocks.

PR #270, opened 2020-06-25
Add c?r?(begin|end) functions for ezArrayMap…

JohannStudanski:
… to make it work in range based for loops (etc.), incl. test.

### DIFF
--- a/Code/Engine/Foundation/Containers/Implementation/ArrayMap_inl.h
+++ b/Code/Engine/Foundation/Containers/Implementation/ArrayMap_inl.h
@@ -179,6 +179,7 @@ VALUE& ezArrayMapBase<KEY, VALUE>::GetValue(ezUInt32 index)
 template <typename KEY, typename VALUE>
 EZ_ALWAYS_INLINE ezDynamicArray<typename ezArrayMapBase<KEY, VALUE>::Pair>& ezArrayMapBase<KEY, VALUE>::GetData()
 {
+  m_bSorted = false;
   return m_Data;
 }
 
--- a/Code/Engine/Foundation/Containers/Implementation/ArrayMap_inl.h
+++ b/Code/Engine/Foundation/Containers/Implementation/ArrayMap_inl.h
@@ -177,6 +177,18 @@ VALUE& ezArrayMapBase<KEY, VALUE>::GetValue(ezUInt32 index)
 }
 
 template <typename KEY, typename VALUE>
+EZ_ALWAYS_INLINE ezDynamicArray<typename ezArrayMapBase<KEY, VALUE>::Pair>& ezArrayMapBase<KEY, VALUE>::GetData()
+{
+  return m_Data;
+}
+
+template <typename KEY, typename VALUE>
+EZ_ALWAYS_INLINE const ezDynamicArray<typename ezArrayMapBase<KEY, VALUE>::Pair>& ezArrayMapBase<KEY, VALUE>::GetData() const
+{
+  return m_Data;
+}
+
+template <typename KEY, typename VALUE>
 template <typename CompatibleKeyType>
 VALUE& ezArrayMapBase<KEY, VALUE>::FindOrAdd(const CompatibleKeyType& key, bool* bExisted)
 {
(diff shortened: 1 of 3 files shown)


Code/Engine/Foundation/Containers/Implementation/ArrayMap_inl.h
```diff
@@ -176,6 +176,18 @@ VALUE& ezArrayMapBase<KEY, VALUE>::GetValue(ezUInt32 index)
   return m_Data[index].value;
 }
 
+template <typename KEY, typename VALUE>
+EZ_ALWAYS_INLINE ezDynamicArray<typename ezArrayMapBase<KEY, VALUE>::Pair>& ezArrayMapBase<KEY, VALUE>::GetData()
```
jankrassnigg, line 180: This would allow people to modify the keys and thus destroy the sorting. I would reset m_bSorted to false here, to guarantee a sorting step before the next lookup. This is not necessary for the const version.

Code/UnitTests/FoundationTest/Containers/ArrayMapTest.cpp
```diff
@@ -7,6 +7,90 @@
...
+
+    //{ // No finding by key (or value) using std::find without a custom comparator functor (same as in MapTest)
+    //  auto itfound = std::find(begin(m), end(m), 500);
+    //  EZ_TEST_INT(itfound->key, 499);
+    //}
+
+    //{ // No finding by key (or value) using std::find without a custom comparator functor (same as in MapTest)
+    //  auto itfound = std::find(rbegin(m), rend(m), 500);
+    //  EZ_TEST_INT(itfound->key, 499);
+    //}
```
jankrassnigg: Is this supposed to get activated?

Code/Engine/Foundation/Containers/Implementation/ArrayMap_inl.h
jankrassnigg: You could add a unit test for modifying the keys, and checking that the still lookup works afterwards, if you like, though I don't see that as super important.

jankrassnigg (approved):
Great addition ! I was aware of this shortcoming but had not time to do this myself, so this is really welcome :-)

Code/UnitTests/FoundationTest/Containers/ArrayMapTest.cpp
JohannStudanski: I don't know! I saw the same code commented out in the MapTest.cpp (Lines 16 & 20), so I added them here as well. I thought they were left commented out there so that e.g. if there's a push to add the comparators it's a little more obvious where they should be tested or something. In any case, I just wanted to mirror the code there.

If you'd rather I just leave it out I can of course do that, too.

Code/Engine/Foundation/Containers/Implementation/ArrayMap_inl.h
JohannStudanski: Good point! I was torn on exposing the method publicly or not doing it and making the begin and end methods friends, Marc favored the public method, so I did that. Should've then added at least the m_bSorted update for the non-const method and a test tho, totally right, I'll do that now.

Code/UnitTests/FoundationTest/Containers/ArrayMapTest.cpp
jankrassnigg: Yeah, remove it. Or try if it would compile now. The comment in the map test is probably very, very old. Maybe it would even compile now? It might work since we added begin/end for ezMap.

JohannStudanski: No it doesn't (it can't, not even in the std:::maps, and I tried): std::find runs an equality comparison between the value you pass and each element in the iterable, but since the iterable elements are pairs in a map, you can only either a) pass a kv-Pair (it'll find that of course), or b) pass in a key or value AND a functor that does the comparison correctly.

It would of course be possible to add the functor—and/or a Find method that searches for values instead of keys... ¯\_(ツ)_/¯

jankrassnigg: Then just get rid of it.

PR #279, opened 2020-07-13
Added compile time version of xxHash32.

Ingrater:
Includes small VS2019 16.6.2 build fix.


C-Core (approved):
Looks good, surprisingly uncomplicated, compilers seem to accept a lot more complicated code as constexpr nowadays.

jankrassnigg:
WOW! That was fast! 👍 

That means we can replace the MurmurHash in ezHashedString with xxHash now. Any reason to keep MurmurHash around, at all? Are there any situations where it might still be preferable?

jankrassnigg (approved):

PR #285, opened 2020-07-21
Microsoft Contributions

bethau:


Code/Engine/Foundation/Math/Implementation/Quat_inl.h
```diff
@@ -132,6 +141,11 @@ bool ezQuatTemplate<Type>::IsNaN() const
 template <typename Type>
 bool ezQuatTemplate<Type>::IsEqualRotation(const ezQuatTemplate<Type>& qOther, float fEpsilon) const
 {
+  if (v.IsEqual(qOther.v, 0.00001f) && ezMath::IsEqual(w, qOther.w, 0.00001f))
```
C-Core, line 144: Use fEpsilon instead of hardcoded value here?

C-Core (approved):

Code/Engine/Foundation/Math/Implementation/Quat_inl.h
Ingrater: fEpsilon is used to compare the angle between the two quaternions. The first if however just checks if the quaternions are the same. Thus a different epsilon value.

PR #288, opened 2020-08-05
Update to Lua 5.4.0 from lua.org

leapach:


Code/ThirdParty/Lua/luaconf.h
```diff
@@ -11,12 +11,19 @@
 #include <limits.h>
 #include <stddef.h>
 
-#ifdef BUILDSYSTEM_COMPILE_ENGINE_AS_DLL
```
jankrassnigg: This needs to be there.

jankrassnigg (changes requested):

Issue #290, opened 2020-08-08
Ambiguous call error with latest MSVC


jayrulez:
https://github.com/ezEngine/ezEngine/blob/c68cf77173178554f2dfffb10ef2c696b8f869ad/Code/Engine/Foundation/Strings/Implementation/FormatStringImpl.h#L98

from 
https://github.com/ezEngine/ezEngine/blob/c68cf77173178554f2dfffb10ef2c696b8f869ad/Code/Engine/Texture/Image/Formats/WicFileFormat.cpp#L231

PR #291, opened 2020-08-08
Fixes issue 25

jayrulez:


jankrassnigg:
Cool thanks! 

If you want GitHub to automatically link to the issue and close it for you, you should reference it with a hash and it's correct number. In this case #290.

PR #299, opened 2020-08-14
Font rendering with freetype

jayrulez:
This is a preliminary pull request and RFC for font rendering of truetype fonts.

The implementation has several components.

1. Runtime font rendering infrastructure.
This component is located under RendererCore/Fonts. I originally thought of adding it under GameEngine/Fonts but I now believe RendererCore/Fonts is more appropriate. 
RendererCore contains representations for glyphs, a bitmap (Face properties, glyph<->character mapping, image atlas(es)), FontResource (engine font resource representation) and TextData which is a utility for composing text data. 

Most of this is adapted from the font rendering implementation in the bsf engine.

2. A tool that imports true type fonts into a format usable by the runtime
This tool takes arguments for an input file and an output file.
The input file should be a ttf file (and maybe otf too).
The tool extracts the relevant information to create a runtime font representation and save it in a file that can be loaded as an ezFontResource.

This tool works so far from testing. It is bare bones now but I will improve it later.

3. A font rendering sample app
This is very WIP and is just a copy of the ShaderExplorer sample right now.

My goal now is to move on to actually rendering fonts before further cleanups and tool improvement.

For this, I'm thinking of introducing an ezTextSprite class and associated ezTextSpriteRenderer.
I'm wondering if I can piggypack on the existing sprite rendering functionality (still need to investigate this) or if I will need to do a ground up implementation.

@jankrassnigg Let me know if this looks okay so far.

Also, feel free to make recommendations for the TextSprite and TextSpriteRenderer implementations.

Freetype is added as a third party dependency but is only needed by the FontImporter tool.


jankrassnigg:
As a reference to others: This PR is meant for reviewing the current state, not to complete it already. I have therefore converted it to a draft PR.

Code/Engine/GameEngine/GameApplication/Implementation/GameApplicationInit.cpp
```diff
@@ -223,6 +224,11 @@ void ezGameApplication::Init_SetupDefaultResources()
   {
     ezResourceManager::AllowResourceTypeAcquireDuringUpdateContent<ezDecalAtlasResource, ezTexture2DResource>();
   }
+
+  // Font
```
jankrassnigg: Have a look at the startup system:
http://ezengine.net/runtime/configuration/startup.html

That allows you to set such stuff up in your code, so you don't need to modify the application code.

Code/Engine/RendererCore/Font/FontBitmap.h
```diff
@@ -0,0 +1,60 @@
...
+  ezUInt32 m_SpaceWidth;
+
+  /** All characters in the font referenced by character ID. */
+  ezMap<ezUInt32, ezFontGlyph> m_Characters;
+};
+
+/**	Contains textures and data about every character for a bitmap font of a specific size. */
+struct EZ_RENDERERCORE_DLL ezFontBitmap : public ezFontBitmapBase
+{
+  ezFontBitmap() = default;
```
jankrassnigg: if the constructor is empty and declared inline anyway, you don't need it at all

Code/Engine/RendererCore/Font/FontGlyph.h
```diff
@@ -0,0 +1,57 @@
...
+  ezResult Deserialize(ezStreamReader& stream);
+};
+
+/**	Describes a single character in a font of a specific size. */
+struct EZ_RENDERERCORE_DLL ezFontGlyph
+{
+  EZ_DECLARE_POD_TYPE();
+
+  ezFontGlyph()
+    : m_CharacterId(0)
```
jankrassnigg: Prefer to just initialize the variables with = X, instead of using a constructor.

Code/Engine/RendererCore/Font/FontResource.h
```diff
@@ -0,0 +1,86 @@
...
+};
+
+struct EZ_RENDERERCORE_DLL ezRawFont
+{
+  EZ_DISALLOW_COPY_AND_ASSIGN(ezRawFont);
+
+  ezRawFont() = default;
+
+  ezResult Write(ezStreamWriter& stream) const;
+  ezResult Read(ezStreamReader& stream);
```
jankrassnigg: These days we call these functions "Serialize" and "Deserialize". With those names, they work better with Streams (some internal magic). Though we haven't changed the names of many existing Write/Read functions, which is why you still see them everywhere.

Just a note, though, not important to change.

Code/Engine/RendererCore/Font/FontResource.h
```diff
@@ -0,0 +1,86 @@
...
+
+  ezString m_Name;
+  ezString m_FamilyName;
+  ezMap<ezUInt32, ezRawFontBitmap> m_BitmapPerSize;
```
jankrassnigg, line 27: Is this just for looking up the font for each size? ezMap has a high memory overhead (4KB), so if you only have few elements it might be quite a waste. You could have a look at ezArrayMap, which is just an array that is always sorted.

C-Core:
I guess this meant to be the foundation for font rendering in UI? It should be in it's own plugin since it is orthogonal to other UI plugins like RmlUi. Furthermore if you want font rendering in 2D UI it doesn't work to pre-bake a texture atlas but you have to generate it on demand at runtime, since the font sizes will be very likely resolution depended.
I only took a brief look at the code so I can't say anything about it yet, the only thing I noticed so far is that the documentation style doesn't much the rest of the ez code.

Code/Engine/RendererCore/Font/FontResource.h
jayrulez: Yes, seems ezArrayMap will work here.

jayrulez:
> I guess this meant to be the foundation for font rendering in UI?  

Yes

> It should be in it's own plugin since it is orthogonal to other UI plugins like RmlUi.

I opted to make the font rendering available from the renderer core because that's what I've seen most of the engines I've checked do.
It really is a generic feature and can be used by other parts of the engine without enabling XYZUIPlugin e.g.: It could be used by the debug renderer, some hyothetical TextSpriteComponent etc...

> Furthermore if you want font rendering in 2D UI it doesn't work to pre-bake a texture atlas but you have to generate it on demand at runtime, since the font sizes will be very likely resolution depended.

This is first step and I haven't looked into resolution independence yet but most of the implementations I have seen for font rendering do pre-bake textures. I do intend to add a runtime rasterized version for fonts later though (This means I'll need to use freetype as a runtime dependency though. Not a problem but something I want to leave until later).

I can move this all to a plugin for now and extract just the font rendering bits for the renderer later if it is found to be useful there.

jankrassnigg:
> I opted to make the font rendering available from the renderer core because that's what I've seen most of the engines I've checked do.

> I can move this all to a plugin for now and extract just the font rendering bits for the renderer later if it is found to be useful there.

I think it would be useful to have in the core engine. On the other hand, we prefer to have any higher level functionality in a plugin, so that it is easy to mix and match the features that you really need. We do have a very basic text renderer (with a hardcoded font), but that is really only for debug rendering.

I'd first get it fully working, at all. Then we can have a look into generating font atlases on the fly (how long does the font rasterization take?). And then we can decide whether it should be a separate plugin, or whether some parts should be in renderer core. It shouldn't be hard to move the code around.

Code/Engine/RendererCore/Font/Implementation/FontBitmap.cpp
```diff
@@ -0,0 +1,73 @@
...
+  {
+    return it.Value();
+  }
+
+  return m_MissingGlyph;
+}
+
+ezResult ezRawFontBitmap::Serialize(ezStreamWriter& stream) const
+{
+  stream << m_Size;
```
jankrassnigg, line 20: We typically write a byte for a version number at the beginning, so that we can later change the file format and implement backwards compatibility.

Code/Engine/RendererCore/Font/Implementation/FontBitmap.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+  stream << texturesCount;
+
+  ezTgaFileFormat fileFormat;
+
+  for (ezUInt32 i = 0; i < texturesCount; i++)
+  {
+    if (fileFormat.WriteImage(stream, m_Textures[i], ezLog::GetThreadLocalLogSystem(), "tga").Failed())
+    {
+      ezLog::Error("Failed to write DDS image chunk to ezFont file.");
```
jankrassnigg, line 36: You are writing this as a TGA at the moment. Though a DDS would be better, that way you can later change the image format or add compression, without the need to change the file format.

Code/Engine/RendererCore/Font/Implementation/FontResource.cpp
```diff
@@ -0,0 +1,235 @@
...
+  ModifyMemoryUsage().m_uiMemoryCPU = sizeof(ezFontResource);
+}
+
+ezResourceLoadDesc ezFontResource::UnloadData(Unload WhatToUnload)
+{
+  ezResourceLoadDesc res;
+  res.m_uiQualityLevelsDiscardable = 0;
+  res.m_uiQualityLevelsLoadable = 0;
+  res.m_State = ezResourceState::Unloaded;
+
```
jankrassnigg, line 93: You can clear your arrays of data and texture handles here, then those textures can be freed.

Code/Engine/RendererCore/Font/Implementation/FontResource.cpp
```diff
@@ -0,0 +1,235 @@
...
+        m_GPUMemory += uiMemory;
+
+        textureHandle = ezResourceManager::CreateResource<ezTexture2DResource>(textureResourceId, std::move(textureDescriptor));
+      }
+
+      fontBitmap.m_TexturePages[texturePageIndex] = textureHandle;
+    }
+
+    m_FontDataPerSize.Insert(rawFontBitmap.m_Size, std::move(fontBitmap));
```
jankrassnigg: Although std::move is not too bad, there is only so much it can do. For example inplace data (hybrid arrays, strings) still need to be copied and some memory allocations still need to be repeated (the devil is in the details). Best guaranteed performance is achieved by simply working on the final object to begin with.

Assuming your m_FontDataPerSize is an ezArrayMap, you would do this:

```
m_FontDataPerSize.Insert(fontSize, std::move(ezFontBitmap())); // insert an empty object
ezUInt32 idx = m_FontDataPerSize.Find(fontSize);
ezFontBitmap& fontObj = m_FontDataPerSize.GetValue(idx);

...read data directly into fontObj...
```

And if you dislike the unnecessary "Find" call (which also triggers a sort), you can add to ezArrayMapBase::Insert to just return the new index (which is just m_Data.GetCount() - 1).

Code/Engine/RendererCore/Font/Implementation/FontResource.cpp
```diff
@@ -0,0 +1,235 @@
...
+  out_NewMemoryUsage.m_uiMemoryGPU = m_GPUMemory;
+}
+
+EZ_RESOURCE_IMPLEMENT_CREATEABLE(ezFontResource, ezFontResourceDescriptor)
+{
+  ezResourceLoadDesc res;
+  res.m_uiQualityLevelsDiscardable = 0;
+  res.m_uiQualityLevelsLoadable = 0;
+  res.m_State = ezResourceState::Loaded;
+
```
jankrassnigg, line 206: To do? You can also remove the creatable stuff, if this font cannot be created.

Code/Engine/RendererCore/Font/Implementation/FontResource.cpp
```diff
@@ -0,0 +1,235 @@
...
+  return res;
+}
+
+const ezFontBitmap& ezFontResource::GetBitmap(ezUInt32 size) const
+{
+  //if (!m_FontDataPerSize.Contains(size))
+  //  EZ_REPORT_FAILURE("Invalid font size requested.");
+
+  auto& bitmap = m_FontDataPerSize.GetValue(size);
```
jankrassnigg: Should this have any kind of error handling or fallback behavior ?

Code/Engine/RendererCore/Font/Implementation/FontResource.cpp
```diff
@@ -0,0 +1,235 @@
...
+
+  return bitmap;
+}
+
+ezInt32 ezFontResource::GetClosestSize(ezUInt32 size) const
+{
+  ezInt32 minDiff = ezMath::MaxValue<ezInt32>();
+  ezInt32 bestSize = size;
```
jankrassnigg: You could start by calling LowerBound and then just compare with the next item. Only those two would need to be checked.

Though since you only have a couple of items in the array, the current solution is performance wise okay too.

Code/Samples/FontSample/Main.cpp
```diff
@@ -0,0 +1,323 @@
+#include <Core/Graphics/Camera.h>
```
jankrassnigg: I know that most of our tools have a "Main.cpp", but this is actually inconvenient. We now prefer to call the file by the name of the tool (e.g. "FontSample.cpp"), so that it is easier to find the file with Visual Assist etc.

Code/Tools/FontImporter/TextureAtlasLayout.h
```diff
@@ -0,0 +1,140 @@
...
+
+class ezTextureAtlasLayout
+{
+  struct ezTextureAtlasNode :public ezRefCounted
+  {
+    ezUInt32 x;
+    ezUInt32 y;
+    ezUInt32 Width;
+    ezUInt32 Height;
+    ezSharedPtr<ezTextureAtlasNode> Left;
```
jankrassnigg: This can probably be a uniqueptr.

jankrassnigg: Same for the root node pointer, and then the class doesn't need to derive from refcounted.

Code/Tools/FontImporter/TextureAtlasLayout.cpp
```diff
@@ -0,0 +1,195 @@
...
+    m_Width = ezMath::Max(m_Width, element.Output.x + element.Input.Width);
+    m_Height = ezMath::Max(m_Height, element.Output.y + element.Input.Height);
+  }
+
+  return true;
+}
+
+void ezTextureAtlasLayout::Clear()
+{
+  if (m_RootNode)
```
jankrassnigg: the if is not needed

Code/Tools/FontImporter/TextureAtlasLayout.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+    if (dw > dh)
+    {
+      left = EZ_DEFAULT_NEW(ezTextureAtlasNode, node->x, node->y, element.Input.Width, node->Height);
+      right = EZ_DEFAULT_NEW(ezTextureAtlasNode, node->x + element.Input.Width, node->y, node->Width - element.Input.Width, node->Height);
+    }
+    else
+    {
+      left = EZ_DEFAULT_NEW(ezTextureAtlasNode, node->x, node->y, node->Width, element.Input.Height);
+      right = EZ_DEFAULT_NEW(ezTextureAtlasNode, node->x, node->y + element.Input.Height, node->Width, node->Height - element.Input.Height);
```
jankrassnigg: Could assign directly to node->Left etc

Code/Tools/FontImporter/TextureAtlasLayout.cpp
```diff
@@ -0,0 +1,195 @@
...
+  ezDynamicArray<ezTextureAtlasPage> pages;
+
+  pages.Reserve(layouts.GetCount());
+
+  for (auto& layout : layouts)
+  {
+    pages.PushBack({layout.GetWidth(), layout.GetHeight()});
+  }
+
+  return pages;
```
jankrassnigg: Again, there is only so much that move semantics can do to prevent unnecessary work. That copying containers around is 'efficient' is unfortunately a myth that the C++ standards committee is actively pushing, for reasons that are not quite clear to me (I think they want to believe this themselves). The reality is, that when you move a container, there are often things that cannot be optimized away, but without digging really deep into your data structures (and ours), it is hard to tell whether it will be fast or slow.

Consequently, we just never do this. Instead we would pass in the output container as a ref parameter, and work directly on it. Some people consider that pattern ugly, but talking about beauty and using C++ at the same time is weird anyway ;-)

Code/Tools/FontImporter/FontImporterApp.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+ezFontImporterApp::ezFontImporterApp()
+  : ezApplication("FontImporter")
+{
+}
+
+ezResult ezFontImporterApp::BeforeCoreSystemsStartup()
+{
+  ezStartup::AddApplicationTag("tool");
+  ezStartup::AddApplicationTag("fontimporter");
```
jankrassnigg: not needed (application tags are a bit of a hack and unless you know that you need it, you don't need to add your own)

Code/Tools/FontImporter/FontImporterApp.cpp
```diff
@@ -0,0 +1,145 @@
...
+  if (ParseCommandLine().Failed())
+    return ezApplication::ApplicationExecution::Quit;
+
+  ezFontImportOptions importOptions;
+  ezRawFont font;
+
+  if (m_FontImporter.Import(m_sInputFile, importOptions, font).Failed())
+  {
+    ezLog::Error("Failed to import font file: {0}.");
+  }
```
jankrassnigg, line 51: On error you should quit here.

Code/Tools/FontImporter/FontImporter.h
```diff
@@ -0,0 +1,19 @@
+#pragma once
+
+#include <Foundation/Basics.h>
+#include <Foundation/Strings/String.h>
+#include <FontImportOptions.h>
+#include <ft2build.h>
+#include FT_FREETYPE_H
```
jankrassnigg, line 7: I'm still wondering how this #include works. What exactly is FT_FREETYPE_H and where does it come from?

Code/Tools/FontImporter/FontImporter.h
```diff
@@ -0,0 +1,19 @@
...
+class ezFontImporter
+{
+public:
+  void Startup();
+  ezResult Import(const ezString& inputFile, const ezFontImportOptions& importOptions, ezRawFont& outFont);
+  void Shutdown();
+
+private:
+  FT_Library m_Library;
+  const static int sMaximumTextureSize = 2048;
```
jankrassnigg, line 18: I think we can increase this to 8K these days.

Code/Tools/FontImporter/FontImporter.cpp
```diff
@@ -0,0 +1,382 @@
+#include <FontImporterPCH.h>
+#include <Foundation/IO/FileSystem/FileReader.h>
```
jankrassnigg, line 3: A tip regarding #includes and the PCH file: Clang-format will rearrange include directives to sort them alphabetically. That's sometimes a problem, and always a problem with the PCH, because that one has to be the first. Add an empty line after the PCH include, to prevent clang-format from sorting it with the others.

Code/Tools/FontImporter/FontImporter.cpp
```diff
@@ -0,0 +1,382 @@
...
+        ezFileWriter file;
+
+        file.Open(fontImageAbsPath);
+
+        auto ezResult = ezImageFileFormat::GetWriterFormat("png")->WriteImage(file, outImg, ezLog::GetThreadLocalLogSystem(), "png");
+
+        EZ_ASSERT_ALWAYS(ezResult == EZ_SUCCESS, "Failed to write file.");
+
+        file.Close();
+      }
```
jankrassnigg, line 358: Maybe add a command line switch for this feature, so that you can keep it in, but disable it by default later.

Code/Engine/RendererCore/Font/TextData.h
```diff
@@ -0,0 +1,360 @@
...
+public:
+  /**
+		 * Initializes a new text data using the specified string and font. Text will attempt to fit into the provided area.
+		 * If enabled it will wrap words to new line when they don't fit. Individual words will be broken into multiple
+		 * pieces if they don't fit on a single line when word break is enabled, otherwise they will be clipped. If the
+		 * specified area is zero size then the text will not be clipped or word wrapped in any way.
+		 *
+		 * After this object is constructed you may call various getter methods to get needed information.
+		 */
+  EZ_RENDERERCORE_DLL ezTextData(const ezString32& text, const ezFontResourceHandle& font, ezUInt32 fontSize, ezUInt32 width = 0, ezUInt32 height = 0, bool wordWrap = false, bool wordBreak = true);
```
jankrassnigg, line 237: You can just mark the entire class with EZ_RENDERERCORE_DLL and don't need to do it for every function.

Code/Engine/RendererCore/Font/TextData.h
```diff
@@ -0,0 +1,360 @@
...
+  ezDynamicArray<ezPageInfo> m_PageInfos;
+  ezUInt32 m_NumPageInfos;
+
+  ezFontResourceHandle m_Font;
+  const ezFontBitmap* m_FontBitmap;
+
+  // Static buffers used to reduce runtime memory allocation
+protected:
+  /** Stores per-thread memory buffers used to reduce memory allocation. */
+  // Note: I could replace this with the global frame allocator to avoid the extra logic
```
jankrassnigg, line 314: That's certainly a good idea. That would get rid of the static thread_local ezBufferData* MemBuffer right ? Because I don't like that variable.

Code/Engine/RendererCore/Font/Implementation/TextData.cpp
```diff
@@ -0,0 +1,786 @@
...
+void ezTextData::Initialize()
+{
+  if (MemBuffer == nullptr)
+    MemBuffer = EZ_DEFAULT_NEW(ezBufferData);
+}
+
+
+////////////////////////////////////////////////////////////////////////////////
+
+thread_local ezTextData::ezBufferData* ezTextData::MemBuffer = nullptr;
```
jankrassnigg, line 590: Try to get rid of this.

jankrassnigg (commented):
Whew, done with the first round of reviewing. That's a lot of code! Overall looking pretty good.

1. Do you know how well it handles Unicode? Ie. can I generate fonts for Cyrillic, Chinese, etc. with this?

1. Since you are porting this from the bsf engine (IIRC) we should definitely give credit to them somehow. Either by mentioning this with a code comment in some files, or with a Readme file or so. You can also put your own name in their, if you like.

Code/Tools/FontImporter/TextureAtlasLayout.h
jayrulez: I was having some issues when I used unique pointer there. I don't remember what it was right now but I can revert and then get back to you with the error.

Code/Tools/FontImporter/FontImporter.h
jayrulez: See https://www.freetype.org/freetype2/docs/tutorial/step1.html

Code/Engine/RendererCore/Font/Implementation/FontResource.cpp
jayrulez: The only call site so far check that the size exist first. I'm not comfortable with leaving it like that though. hence the commented validation check. I'm not sure I want to fail here though but may be the best option. I will uncomment the validation code for now but I'll think how best to proceed later.

Code/Engine/RendererCore/Font/TextData.h
jayrulez: This has to come after I remove the tls MemBuffer.

Code/Tools/FontImporter/TextureAtlasLayout.h
jayrulez: I needed to create a move constructor for ezTextureAtlasLayout to move m_RootNode when PushBack is called on the array.

Code/Engine/RendererCore/Font/Implementation/FontResource.cpp
jayrulez: I submitted a separate PR for that. I'll update this branch accordingly.

Code/Engine/RendererCore/Font/Implementation/FontResource.cpp
jayrulez: > Whew, done with the first round of reviewing. That's a lot of code! Overall looking pretty good.
> 
> 1. Do you know how well it handles Unicode? Ie. can I generate fonts for Cyrillic, Chinese, etc. with this?

Yes, Would just need to include character ranges for the characters we need.
Chinese for example is in CJK which is huge. 

Here's a sample output for a part of the range:

NotoSansSC-Black otfNotoSansSC-Black_30_0

The CJK common range includes about seven pages * 2K textures in total.
This contains Chinese, Japanese and others. The key is to only specify the ranges you actually need. We should probably import a different font file per language though...

> 2. Since you are porting this from the bsf engine (IIRC) we should definitely give credit to them somehow. Either by mentioning this with a code comment in some files, or with a Readme file or so. You can also put your own name in their, if you like.
Yep, I'll definitely add the necessary credits.

jayrulez: CK common range I used for the is ezCharacterRange(19968, 40959)

Ingrater:
@jayrulez could you please rebase to latest dev?

jayrulez:
> @jayrulez could you please rebase to latest dev?

Done

C-Core:
> > I opted to make the font rendering available from the renderer core because that's what I've seen most of the engines I've checked do.
> 
> > I can move this all to a plugin for now and extract just the font rendering bits for the renderer later if it is found to be useful there.
> 
> I think it would be useful to have in the core engine. On the other hand, we prefer to have any higher level functionality in a plugin, so that it is easy to mix and match the features that you really need. We do have a very basic text renderer (with a hardcoded font), but that is really only for debug rendering.
> 
> I'd first get it fully working, at all. Then we can have a look into generating font atlases on the fly (how long does the font rasterization take?). And then we can decide whether it should be a separate plugin, or whether some parts should be in renderer core. It shouldn't be hard to move the code around.

I don't really see how one would use the text rendering feature alone. We have the basic font for debug rendering already. Anything else will most likely need more UI related stuff like layouting etc. If you use e.g. the RmlUI plugin for that all the font rendering stuff in renderer core is just unused bloat and increases the binary size. Therefore I think we should move it to an ez UI plugin.

PR #310, opened 2020-08-22
Fixed type conversion warnings

C-Core:
and enabled the warning again in the build settings

### DIFF
--- a/Code/BuildSystem/CMake/ezUtilsCppFlags.cmake
+++ b/Code/BuildSystem/CMake/ezUtilsCppFlags.cmake
@@ -117,11 +117,10 @@ function(ez_set_build_flags_msvc TARGET_NAME)
 	# 4100 = unreferenced formal parameter
 	# 4189 = local variable is initialized but not referenced
 	# 4127 = conditional expression is constant
-	# 4244 = conversion from 'int 32' to 'int 16', possible loss of data
 	# 4245 = signed/unsigned mismatch
 	# 4389 = signed/unsigned mismatch
 	# 4310 = cast truncates constant value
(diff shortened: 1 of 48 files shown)


C-Core:
@jankrassnigg The typescript plugin converts a component handle to an uint32 although a component handle is 64bit, this doesn't seem to be correct?

E.g.
C:\Source\ezEngine\Code\EnginePlugins\TypeScriptPlugin\TsBinding\TsComponent.cpp(111): warning C4244: 'initializing': conversion from 'ezComponentId::StorageType' to 'ezUInt32', possible loss of data (compiling source file C:\Source\ezEngine\Workspace\Win64\Code\EnginePlugins\TypeScriptPlugin\TsBinding\unity_6905569600eebe6ba989059ef60dbe1195557b23.cpp)

jankrassnigg:
> @jankrassnigg The typescript plugin converts a component handle to an uint32 although a component handle is 64bit, this doesn't seem to be correct?

Yes, looks fishy. It works, because nobody calls that function :D There is a pointer overload for that function, which does something entirely different, and that is used everywhere. Probably some left-over from the early development phase.

Can you just remove that function in your change?

C-Core:
> Yes, looks fishy. It works, because nobody calls that function :D There is a pointer overload for that function, which does something entirely different, and that is used everywhere. Probably some left-over from the early development phase.
> 
> Can you just remove that function in your change?

I removed the function but the same problem exists in ezTypeScriptBinding::DeleteTsComponent. What should I do with this function? I can't remove it since it used in ezTypeScriptComponent::Deinitialize.

jankrassnigg:
I checked that, it was dead code as well by now. We have a 'garbage collection' step to remove stuff and this call had no effect anymore.

I removed those functions in latest dev.

PR #316, opened 2020-08-27
Added new small array container class

C-Core:
which is a hybrid array with minimal memory overhead, but can only store up to 65535 elements.

ezTagSet uses this small array as storage and now has a memory footprint of 16 bytes instead of 40 bytes.
ezGameObject uses a small array for components and is now down to 128 bytes :)

### DIFF
--- a/Code/Engine/Core/World/Implementation/GameObject.cpp
+++ b/Code/Engine/Core/World/Implementation/GameObject.cpp
@@ -109,9 +109,9 @@ void ezGameObject::Reflection_RemoveComponent(ezComponent* pComponent)
   }
 }
 
-const ezHybridArray<ezComponent*, ezGameObject::NUM_INPLACE_COMPONENTS>& ezGameObject::Reflection_GetComponents() const
+ezHybridArray<ezComponent*, ezGameObject::NUM_INPLACE_COMPONENTS> ezGameObject::Reflection_GetComponents() const
 {
-  return m_Components;
+  return ezHybridArray<ezComponent*, ezGameObject::NUM_INPLACE_COMPONENTS>(m_Components);
 }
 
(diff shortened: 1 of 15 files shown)


Code/Engine/Core/World/Implementation/GameObject.cpp
```diff
@@ -244,6 +244,12 @@ void ezGameObject::ConstChildIterator::Next()
   m_pObject = m_pWorld->GetObjectUnchecked(m_pObject->m_NextSiblingIndex);
 }
 
+ezGameObject::~ezGameObject()
+{
+  m_Components.Clear();
+  m_Components.Compact(GetWorld()->GetAllocator());
```
jankrassnigg, line 251: Maybe add a comment why this is necessary.

Code/Engine/Foundation/Containers/SmallArray.h
```diff
@@ -0,0 +1,256 @@
+#pragma once
+
+#include <Foundation/Algorithm/Sorting.h>
+#include <Foundation/Math/Math.h>
+#include <Foundation/Memory/AllocatorWrapper.h>
+#include <Foundation/Types/ArrayPtr.h>
+
+#ifndef ezSmallInvalidIndex
+#  define ezSmallInvalidIndex 0xFFFF
```
jankrassnigg: Could we use static constexpr for this now?

Code/Engine/Foundation/Containers/SmallArray.h
```diff
@@ -0,0 +1,256 @@
...
+#endif
+
+/// \brief Implementation a dynamically growing array with in place storage and small memory overhead.
```
jankrassnigg: Typo

jankrassnigg (approved):

PR #322, opened 2020-09-05
Rewrote the ModelImporter

jankrassnigg:
* New importer is in ModelImporter2 project
* Renamed the skeleton visualization component + added drag n drop
* Removed the scene import functionality

Note:
* The default orientation for imported meshes was reversed. That means existing meshes may flip now.
* Order of sub-meshes may change, you may need to reassign materials.


Code/Tools/Libs/ModelImporter2/ImporterAssimp/AssimpUtils.cpp
```diff
@@ -0,0 +1,63 @@
+#include <ModelImporterPCH.h>
+
+#include <assimp/matrix4x4.h>
+#include <assimp/quaternion.h>
+#include <assimp/types.h>
+#include <assimp/vector3.h>
+
+namespace ezModelImporter2
+{
+  ezColor ConvertAssimpType(const aiColor4D& value, bool invert = false)
```
jayrulez: VS complains here:
bool invert/* = false*/ instead

Code/Tools/Libs/ModelImporter2/ImporterAssimp/AssimpUtils.cpp
```diff
@@ -0,0 +1,63 @@
...
+  {
+    if (invert)
+      return ezColor(1.0f - value.r, 1.0f - value.g, 1.0f - value.b, 1.0f - value.a);
+    else
+      return ezColor(value.r, value.g, value.b, value.a);
+  }
+
+  ezColor ConvertAssimpType(const aiColor3D& value, bool invert = false)
```
jayrulez: same as above

Code/Tools/Libs/ModelImporter2/ImporterAssimp/AssimpUtils.cpp
jankrassnigg: Interesting, didn't complain for me.

PR #324, opened 2020-09-07
Fix for #314: BoundingSphere ExpandToInclude fails with distances below 1

### LINKED ISSUE #314: [BUG] Bounding(Box)Sphere set radius to -1, which doesn't work with ExpandToInclude
ezBoundingSphere and ezBoundingBoxSphere::SetInvalid() set the radius to -1.

Using this with ExpandToInclude(vertex) with a vertex that is closer than 1 unit, fails to grow the box, because the sphere uses the squared radius to determine how much to grow. Since sqr(-1) is 1, it thinks it doesn't need to grow.

This can result in crashes in places that expect a valid bounding sphere.

amir9480:


jankrassnigg:
Hey, thanks for the contribution :)

It took me a while to realize why you chose the epsilon value. I guess it's a good solution that doesn't cost any extra performance.

To finish this, can you do a couple of follow up changes?

1) I think you can use SmallEpsilon().
2) Can you have a look at BoundingSphereTest.cpp and BoundingBoxSphereTest.cpp and add a quick test that actually tests your fix? Ie. something that would have failed without it.
3) As mentioned in #314, can you check ezSimdBSphere and ezSimdBBoxSphere and do the same fix (+ unit test), if necessary?

That would properly finish the necessary work. Many thanks for the help!

amir9480:
Hi @jankrassnigg 

My pleasure to contribute
ezEngine is one of my inspiration to make my own game engine.

I converted `DefaultEpsilon` to `SmallEpsilon` and also added this solution to `ezSimdBSphere` & `ezSimdBBoxSphere` classes.
and added tests for that.
I hope I did it right.

jankrassnigg:
Awesome, thank you :)

PR #326, opened 2020-09-16
Mirrored mesh import fix

C-Core:
Reverted mesh importing behavior back to how it was before. Fixed pin context menu in render pipeline assets to not contain the "debug" option


jankrassnigg (approved):

jankrassnigg:
I wouldn't be surprised if the GameEngine test failed because you checked in new image comparison images. With the flip axis change, the images should not need an update.

C-Core:
My local test caches were broken and I only saw that the new images looked "better" since the text on the texture was not mirrored, but yeah they shouldn't have changed. Reverted them again.

PR #327, opened 2020-09-30
Particle Rendering fixes

### LINKED ISSUE #281: [BUG] Particle culling broken when attached to moving object
The bounds for particle effects is updated only at 10Hz. When a particle effect is attached to a moving object, those bounds are moved along with the parent, although they are in world space.

As a result, the bounds jitter back and forth, and when the owner rotates, they can end up outside the camera view frustum easily, which results in flickering, because every 10th of a second the bounds are corrected.

Additionally: Some effects sometimes pop up too late (bullet impacts). It seems as if they do a bounds update during the first frame, which results in a zero bbox, which means the effect is culled entirely, and only after a 10th second the bbox will be non-zero and the effect will appear suddenly. Not 100% certain whether that is what's happening, though.

Forcing a bounds update every frame fixes both issues, but is not intended.

C-Core:
* Ensure that particles are only rendered in the views where they are actually visible
* Simplified a few places and elimated some frame delays in particle effect transform and bounds update


Code/EnginePlugins/ParticlePlugin/Finalizer/ParticleFinalizer_Volume.cpp
```diff
@@ -47,53 +47,43 @@ void ezParticleFinalizer_Volume::QueryOptionalStreams()
...
-      fMaxSize[4] = ezMath::Max(fMaxSize[4], (float)pSize[idx + 4]);
-      fMaxSize[5] = ezMath::Max(fMaxSize[5], (float)pSize[idx + 5]);
-      fMaxSize[6] = ezMath::Max(fMaxSize[6], (float)pSize[idx + 6]);
-      fMaxSize[7] = ezMath::Max(fMaxSize[7], (float)pSize[idx + 7]);
-
-      idx += 8;
+      const float x = pSize[i + 0];
+      const float y = pSize[i + 0];
+      const float z = pSize[i + 0];
+      const float w = pSize[i + 0];
```
jankrassnigg: 4x +0 ?

C-Core: yeah for copy pasta :D

Code/EnginePlugins/ParticlePlugin/System/ParticleSystemInstance.cpp
```diff
@@ -338,11 +337,10 @@ void ezParticleSystemInstance::ReinitializeStreamProcessors(const ezParticleSyst
 
 ezParticleSystemInstance::ezParticleSystemInstance()
 {
-  m_BoundingVolume = ezBoundingSphere(ezVec3::ZeroVector(), 0.25f);
+  m_BoundingVolume.SetInvalid();
```
jankrassnigg, line 340: Was this a problem ?

jankrassnigg (approved):
Need to fix the incorrect indices. Otherwise all fine.

Code/EnginePlugins/ParticlePlugin/System/ParticleSystemInstance.cpp
C-Core: Yes, since the bounding volume is in global space (for non-local space particles) this will be transformed to somewhere in the first frame which could result in particles never spawed for effect that are paused when invisible. It might be not a problem anymore with the new transform and bounds update code in place, but I would leave it in for consistency. The bounds are also invalid if there are no particles spawned yet.

Code/EnginePlugins/ParticlePlugin/Components/ParticleComponent.cpp
```diff
@@ -342,7 +336,7 @@ void ezParticleComponent::OnMsgExtractRenderData(ezMsgExtractRenderData& msg) co
   if (msg.m_pView->GetCameraUsageHint() == ezCameraUsageHint::Shadow)
     return;
 
-  m_EffectController.SetIsInView();
```
jankrassnigg, line 345: I still think that the removal of this line might affect the first update.

jankrassnigg (changes requested):

jankrassnigg (approved):

PR #332, opened 2020-10-11
OpenXR upgrade

### LINKED ISSUE #237: Revisit OpenXR integration into ApplicationEntryPoint_uwp.h
Currently OpenXR takes care of creating its own window which prevents keyboard/mouse input to be set up as in a usual UWP app.

Sanakan8472:
* Updated OpenXR loader and include NuGet packages to the newest version.
  * Replace preview extensions with final versions.
* Basic implementation of XR_MSFT_HOLOGRAPHIC_WINDOW_ATTACHMENT_EXTENSION_NAME. Resolves #237
* WACK compliance for enent and zlib
* Compiler fix for newest windows SDK which renamed _TlgPragmaUtf8Begin etc.
* Rewrote ezUwpApplication in cppwinrt
* Fallback to flatscreen if XR actor can't be created.
* ezDeviceTrackingComponent extended with option to exclude rotation and scale
* Compile fix for ezPropertyAnimComponent which relies on CommonMessages.h but does not include it itself.

### DIFF
--- a/Code/Engine/GameEngine/GameState/Implementation/GameState.cpp
+++ b/Code/Engine/GameEngine/GameState/Implementation/GameState.cpp
@@ -115,9 +115,22 @@ void ezGameState::CreateActors()
       ezView* pView = nullptr;
       EZ_VERIFY(ezRenderWorld::TryGetView(m_hMainView, pView), "");
       ezUniquePtr<ezActor> pXRActor = pXRInterface->CreateActor(pView, ezGALMSAASampleCount::Default, std::move(pMainWindow), std::move(pOutput));
-      ezActorManager::GetSingleton()->AddActor(std::move(pXRActor));
+
+      if (pXRActor)
+      {
+        ezActorManager::GetSingleton()->AddActor(std::move(pXRActor));
+        return;
+      }
+      else
+      {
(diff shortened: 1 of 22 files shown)


Code/Engine/GameEngine/GameState/Implementation/GameState.cpp
```diff
@@ -125,9 +125,22 @@ void ezGameState::CreateActors()
       ezView* pView = nullptr;
       EZ_VERIFY(ezRenderWorld::TryGetView(m_hMainView, pView), "");
       ezUniquePtr<ezActor> pXRActor = pXRInterface->CreateActor(pView, ezGALMSAASampleCount::Default, std::move(pMainWindow), std::move(pOutput));
-      ezActorManager::GetSingleton()->AddActor(std::move(pXRActor));
```
jankrassnigg, line 118: I guess the commented out code 20 lines above could go now. Also this piece could be moved into a helper function.

jankrassnigg (approved):

Code/Engine/GameEngine/GameState/Implementation/GameState.cpp
Sanakan8472: Code was removed already after rebase.

PR #338, opened 2020-10-15
Allow using an 'ezSdkRoot.txt' redirection file to point to the SDK dir

jankrassnigg:
If ezEngine is hooked up as a submodule, ez apps can't find the proper folder where the SDK root is located.

With this change you can put a file 'ezSdkRoot.txt' into the top most folder of your repository. It has to contain the relative path from this directory, to the ez SDK directory.

So if your repository is in "C:\MyRepo" and ez is a submodule located in "C:\MyRepo\ezSubmodule", you would put the file into "C:\MyRepo" and it should only contain the string "ezSubmodule".

Then all tools (ezEditor etc) and all ez(Game)Applications will automatically find the correct root.

### DIFF
--- a/Code/Engine/Foundation/IO/FileSystem/Implementation/FileSystem.cpp
+++ b/Code/Engine/Foundation/IO/FileSystem/Implementation/FileSystem.cpp
@@ -676,8 +676,7 @@ ezResult ezFileSystem::ResolvePath(const char* szPath, ezStringBuilder* out_sAbs
   return EZ_SUCCESS;
 }
 
-
-ezResult ezFileSystem::FindFolderWithSubPath(const char* szStartDirectory, const char* szSubPath, ezStringBuilder& result)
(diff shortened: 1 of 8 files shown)


Code/Engine/Foundation/IO/FileSystem/Implementation/FileSystem.cpp
```diff
@@ -702,16 +702,36 @@ ezResult ezFileSystem::FindFolderWithSubPath(const char* szStartDirectory, const
   result = szStartDirectory;
   result.MakeCleanPath();
 
-  ezStringBuilder FullPath;
+  ezStringBuilder FullPath, sRedirection;
 
   while (!result.IsEmpty())
   {
     FullPath = sStartDirAbs;
+    FullPath.AppendPath("ezSdkRoot.txt");
```
jankrassnigg: I think the name of this file will be configurable later, as it is currently hacked into this function.

jayrulez:
Works perfectly. Thanks!

PR #345, opened 2020-11-14
Custom variant type support

Sanakan8472:
ezVariant
- ezVariant::Type enum and ezVariant::TypeDeduction has been extracted into ezVariantType and ezVariantTypeDeduction and moved into VariantType.h
- Added GetReflectedType function which returns the ezRTTI type of the object inside the ezVariant. For TypedPointer and TypedObject this will return the type of the referenced pointer/object.
- Const corectness: operator[] is now returning a const object.
- DispatchTo can now take any number of parameters and supports return values.
- Unsafe non-const GetData replaced with GetWriteAccess which ensures that a shared value has ref count 1 or is cloned to make it so to allow exclusive access.
- GetData and GetWriteAccess will for TypedPointer and TypedObject return the address of the pointed to object.

Pointers
- Added ezTypedPointer struct - a raw ptr + ezRTTI type.
- ezVariantType::ReflectedPointer and VoidPointer are replaced with TypedPointer, storing a ezTypedPointer struct.
- ezDynamicCast / ezVariant::Get<T*> can now cast an ezVariant the same way as any statically known type would via dynamic_cast<T>.
- Type can be blank at which all cast will fail except for void* / const void*.

Custom variant type
- ezVariantType::TypedObject added, allowing to store custom types by value inside an ezVariant. The storage inside the ezVariant is ezTypedObject.
- EZ_DECLARE_CUSTOM_VARIANT_TYPE declares a custom type, allowing to be directly assigned and stored inside an ezVariant.
- EZ_DEFINE_CUSTOM_VARIANT_TYPE defines a custom type. Extends the custom type to allow ot to be compared, hashed and serialized. Requires the type to implement ezHashHelper<T> and ezStreamWriter ezStreamReader operators.
  - Functionlity defined by EZ_DEFINE_CUSTOM_VARIANT_TYPE is accessible via ezVariantTypeRegistry, but should not be needed to be accessed outside of ezVariant / serialization code.
- A both declared and defined custom type has the exact same capabilities as a built in type.
- Custom types should only have member variables and no dynamic data. This is currently an arbitrary limitation (works as implemented).
- Custom types that are sizeof(T) <= 16 and POD will be inlined inside the ezVariant.
- ezVariant::CopyTypedObject and ezVariant::MoveTypedObject can be used to initialize an ezVariant to a TypedObject if the type is not known at compile time.
- ezVarianceTypeFloat, ezVarianceTypeTime, ezVarianceTypeAngle are now value types.
- Function reflection parameters for custom types will accept both values passed by value as well as those passed as pointers.

Migration to a new custom variant type:
- Create a patch for any class using the type as a property and use ezAbstractObjectNode::InlineProperty to patch it.
- Property widgets need to be changed to handle the property by property value instead of it being an object.
- Add reasonable default and clamp values where applicable. Any previous default that were set on the fields of the custom type's fields will be ignored as it assumed to be an opague data type and is not inspected deeper.

Misc fixes:
- ezReflectionUtils::IsValueType added to quickly detect types that are stored by value in an ezVariant.
- Property-based dispatcher added to reflection utils: void DispatchTo(Functor& functor, const ezAbstractProperty* pProp, ...)
- const corectness in ezAbstractSetProperty
- Test fixes when running on intel onboard GPU.


Code/Engine/Foundation/IO/Implementation/OpenDdlUtils.cpp
```diff
@@ -1372,6 +1411,33 @@ void ezOpenDdlUtils::StoreVariant(ezOpenDdlWriter& writer, const ezVariant& valu
...
+      ezTypedObject obj = value.Get<ezTypedObject>();
+      if (ezVariantTypeRegistry::GetSingleton()->FindVariantTypeInfo(obj.m_pType))
+      {
+        writer.BeginObject(obj.m_pType->GetTypeName(), szName, bGlobalName);
+        {
+          ezHybridArray<ezAbstractProperty*, 32> properties;
+          obj.m_pType->GetAllProperties(properties);
+          for (const ezAbstractProperty* pProp : properties)
+          {
+            // Custom types should be POD and only consist of member properties.
```
jankrassnigg, line 1425: should it be an error, if there are non-member types ?

Code/Engine/Foundation/Reflection/Implementation/VariantAdapter.h
```diff
@@ -90,11 +89,31 @@ struct ezIsStandardType<T, ezVariant>
 
 //////////////////////////////////////////////////////////////////////////
 
+/// \brief Used to determine if the given type is a
```
jankrassnigg: is a .... ?

Code/Engine/Foundation/Types/Implementation/VariantTypeRegistry.cpp
```diff
@@ -0,0 +1,74 @@
...
+const ezVariantTypeInfo* ezVariantTypeRegistry::FindVariantTypeInfo(const ezRTTI* pType) const
+{
+  const ezVariantTypeInfo* pTypeInfo = nullptr;
+  m_TypeInfos.TryGetValue(pType, pTypeInfo);
+  return pTypeInfo;
+}
+
+void ezVariantTypeRegistry::PluginEventHandler(const ezPluginEvent& EventData)
+{
+  UpdateTypes();
```
jankrassnigg: I think you could filter out a few redundant events.

Code/Engine/Foundation/Types/Variant.h
```diff
@@ -5,18 +5,35 @@
...
-class ezVariant;
-struct ezTime;
-class ezUuid;
-struct ezStringView;
+class ezRTTI;
+
+/// \brief Defines a reference to an immutable object owned by an ezVariant.
+///
+/// Used to store custom types inside an ezVariant. As lifetime is governed by the ezVariant, it is generally not safe to store an ezTypedObject.
+/// This class is needed to be able to difference between ezVariantType::TypedPointer and ezVariantType::TypedObject e.g. in ezVariant::DispatchTo.
```
jankrassnigg: "differentiate" instead of "difference"

Code/Tools/Libs/GuiFoundation/DockPanels/ApplicationPanel.moc.h
```diff
@@ -31,3 +32,4 @@ class EZ_GUIFOUNDATION_DLL ezQtApplicationPanel : public ads::CDockWidget
 
   ezQtContainerWindow* m_pContainerWindow;
 };
+EZ_DECLARE_REFLECTABLE_TYPE(EZ_GUIFOUNDATION_DLL, ezQtApplicationPanel);
```
jankrassnigg, line 35: Why is this reflected now ?

jankrassnigg (approved):
Not that I understand much of this PR :D

Code/Tools/Libs/GuiFoundation/DockPanels/ApplicationPanel.moc.h
Sanakan8472: Not strictly necessary but I fixed all cases where we stored a void pointer inside an ezVariant. By reflecting it you can now do dyncasts instead of just assuming it is the type you expected. Now I might have missed a few cases where we read the variants, so if you come across a .Get<void*>() or .ConvertTo<void*>() wrapped in a static_cast, please replace it by using the type directly.

Code/Engine/Foundation/Types/Implementation/VariantTypeRegistry.cpp
Sanakan8472: I hope this us correct now. I look at AfterLoadingBeforeInit  and AfterUnloading

Code/Engine/Foundation/IO/Implementation/OpenDdlUtils.cpp
Sanakan8472: Yep, added error handling.

PR #353, opened 2020-12-19
Allow importing clips with duration less than 1 second

### LINKED ISSUE #352: [BUG]Unable to import animation clip where duration is less than 1
**Bug description:**
The model importer fail to import animation clips where the clip duration is less than 1 second.

https://github.com/ezEngine/ezEngine/blob/dev/Code/Tools/Libs/ModelImporter2/ImporterAssimp/AssimpAnimationImport.cpp#L74

this line will result in 0 if the anim duration is less than 1.

I'm not sure what is the best way to handle this. Is there some minimum duration for an animation clip than can be used as a lower bound here?

**How to Reproduce:**
Try to import an animation clip where the duration is less than 1 second.

**Dev-branch commit hash:**
618164c8b233a6f4e7b5145f766f9448504a223e

jayrulez:
Fixes  #352


jankrassnigg:
Looks reasonable.

PR #358, opened 2020-12-27
RendererFoundation refactor

### LINKED ISSUE #349: [FEATURE]Modernize RenderFoundation API
As a follow up to the topic here: https://github.com/ezEngine/docs/wiki/Vulkan-Support

I do agree that the best way to Vulkan support for ez is to upgrade the RenderFoundation API in place to properly abstract it for both Vulkan and DX11 (at least).

I've experimented with porting two separate low level APIs to ezEngine from WickedEngine and Veldrid (A C++ port of the C# code using ez base libraries).

Both experiments worked to some extent and I've gained a little bit of experience with some low level graphics programming. I believe with that I may be able to help with modernizing the ez RenderFoundation.

Comparing ez to these other libs (with are considered modern) at a high level, these are the things I think we should be looking into for modifying RenderFoundation:

1. Command lists
This encapsulates a list of commands that can be submitted to the device in the order that they were recorded.
A command list would do essentially what Context does. However, instead of submitting each command immediately, the commands are recorded and submitted at once.

In DX11, this can be achieved by creating a deferred context per abstracted command list and create a native command list from the deferred context then submit then execute the command list using the immediate context.
Command lists would not be thread safe but you can create a command list for any thread, record commands there and execute them on the main/render thread using the immediate context.
I think implementing command lists like this in D3D11 is not necessary though but would bring it a lot closer to the potential vulkan port.

This could be implemented without too much changes to many changes to the consuming code. The Command list feature could exist alongside the current Context class until everything is upgraded to use it.

I could actually work on this as it is low hanging.

2. Monolithic pipeline state
In newer graphics APIs, the pipeline state is created upfront and set as a monolith. Pipeline states can be cached to avoid recreating them when it is not necessary.

In DX11, an abstracted pipeline state would encapsulate your shaders, blend, rasterizer states etc...
Again, implementing this for DX11 would be only to make it look more like Vulkan.

This is something I could also work on for the most part.

3. Render Passes
We could use this as a start https://developer.samsung.com/galaxy-gamedev/resources/articles/renderpasses.html

@C-Core @jakrams if you have any thoughts about the design you'd like please leave them here.

C-Core:
Added concept of passes and command encoder to renderer foundation

Also added a factory for device creation. This allows us to switch between renderer implementations via command line argument e.g. "-renderer DX11" or "-renderer Vulkan".


Code/Engine/GameEngine/CMakeLists.txt
```diff
@@ -9,17 +9,21 @@ ez_create_target(LIBRARY ${PROJECT_NAME})
...
 
 target_link_libraries(${PROJECT_NAME}
   PUBLIC
   RendererCore
   ozz
+)
+
+add_dependencies(${PROJECT_NAME}
+  RendererDX11
+  RendererVulkan
```
C-Core: @jankrassnigg Any idea how to add this runtime dependency without getting an error in cmake if RendererVulkan or RendererDX11 does not exist (when EZ_BUILD_EXPERIMENTAL_VULKAN is not enabled)

Code/Engine/GameEngine/GameApplication/Implementation/GameApplicationInit.cpp
```diff
@@ -209,6 +205,16 @@ void ezGameApplication::Init_SetupDefaultResources()
...
 
+const char* GetRendererNameFromCommandLine()
+{
+#ifdef BUILDSYSTEM_ENABLE_VULKAN_SUPPORT
+  constexpr const char* szDefaultRenderer = "Vulkan";
+#else
+  constexpr const char* szDefaultRenderer = "DX11";
+#endif
+
+  return ezCommandLineUtils::GetGlobalInstance()->GetStringOption("-renderer", 0, szDefaultRenderer);
```
jankrassnigg: Have a look at ezCommandLineOptionEnum.

jankrassnigg (approved):

Code/Engine/GameEngine/CMakeLists.txt
jankrassnigg: I think so. Will have a look later.

jankrassnigg: In CMake you can check whether a target exists:

  if (TARGET blub)
    add_dependencies(bla blub)
  endif()

So that's how you can add the dependency only if CMake knows about it. Unfortunately, CMake evaluates that if right away, and therefore may not know about a target, that will be declared later. That's why I introduced the 'finalize.cmake' files. Those are run after all the projects are declared. So if you put such a file into this directory, you can do that check and it should always work.

Just have a look at some other finalize.cmake file, they are all used for similar purposes.

C-Core: I've added the following to finalize.cmake for the GameEngine:

if (TARGET RendererDX11)
  add_dependencies(${PROJECT_NAME} RendererDX11)
endif()

then cmake complains:

CMake Error at Code/Engine/GameEngine/finalize.cmake:2 (add_dependencies):
  Cannot add target-level dependencies to non-existent target "ezVs2019".

  The add_dependencies works for top-level logical targets created by the
  add_executable, add_library, or add_custom_target commands.  If you want to
  add file-level dependencies see the DEPENDS option of the add_custom_target
  and add_custom_command commands.
Call Stack (most recent call first):
  Code/BuildSystem/CMake/ezUtils.cmake:477 (include)
  CMakeLists.txt:65 (ez_finalize_projects)

Code/Engine/Foundation/Utilities/CommandLineOptions.h
```diff
@@ -222,7 +222,7 @@ class EZ_FOUNDATION_DLL ezCommandLineOptionString : public ezCommandLineOptionDo
   ezCommandLineOptionString(const char* szSortingGroup, const char* szArgument, const char* szLongDesc, const char* szDefaultValue, bool bCaseSensitive = false);
 
   /// \brief Returns the value of this option. Either what was specified on the command line, or the default value.
-  ezString GetOptionValue(LogMode logMode, const ezCommandLineUtils* pUtils = ezCommandLineUtils::GetGlobalInstance()) const; // [tested]
+  const char* GetOptionValue(LogMode logMode, const ezCommandLineUtils* pUtils = ezCommandLineUtils::GetGlobalInstance()) const; // [tested]
```
C-Core, line 225: @jankrassnigg I changed the return value to const char* because you can easily end up with garbage strings if you return an ezString by value but the user assign it to a const char* then.

PR #368, opened 2021-01-29
View additions

C-Core:
Added lod camera to view
Also added additional permutation vars that can be set on the view. This is used to set a permutation var for the material preview in the editor that can be used by shaders to deactivate parts that need additional data which is not set in the material preview.


jankrassnigg:
Awesome!

PR #373, opened 2021-02-11
Blackboard component

C-Core:
* Added a new component that owns a blackboard. Initial values of the blackboard can be set via Editor UI, which is important for data binding with RmlUi because you can't add new blackboard entries after the binding has been established. Changing the values afterwards is of course possible.
* Changed that the RmlUi component creates its context lazy. This allows other components to setup data bindings and then activate the RmlUi component independent of component init order.
* Added an "autobind blackboards" option to the RmlUi component where it searches for blackboard components on its owner object and their parents and automatically binds their blackboards.
* Various other smaller fixes


Code/EnginePlugins/RmlUiPlugin/Components/Implementation/RmlUiCanvas2DComponent.cpp
```diff
@@ -333,3 +384,28 @@ void ezRmlUiCanvas2DComponent::UpdateCachedValues()
...
+
+  m_AutoBindings.Clear();
+
+  if (m_bAutobindBlackboards)
+  {
+    ezGameObject* pObject = GetOwner();
+    while (pObject != nullptr)
+    {
+      ezBlackboardComponent* pBlackboardComponent = nullptr;
+      if (pObject->TryGetComponentOfBaseType(pBlackboardComponent))
```
jankrassnigg: Could have multiple blackboards on one object

Code/Engine/GameEngine/Gameplay/BlackboardComponent.h
```diff
@@ -0,0 +1,72 @@
...
+EZ_DECLARE_REFLECTABLE_TYPE(EZ_GAMEENGINE_DLL, ezBlackboardEntry);
+
+//////////////////////////////////////////////////////////////////////////
+
+struct ezMsgUpdateLocalBounds;
+struct ezMsgExtractRenderData;
+
+using ezBlackboardComponentManager = ezComponentManager<class ezBlackboardComponent, ezBlockStorageType::Compact>;
+
+class EZ_GAMEENGINE_DLL ezBlackboardComponent : public ezComponent
```
jankrassnigg, line 30: Some documentation comments would be nice

jankrassnigg (approved):

PR #380, opened 2021-02-25
Event message handler improvements

C-Core:
* Event messages will now be sent to all components that handle the message type on one object.
* Also they will be sent to all global event handler which handle the message type if no handler is found in the hierarchy.
* Added pass through flag to pass unhandled event messages to parent objects

Other fixes:
* Ensure that mesh normals and tangents are normalized before encoding into vertex buffer, otherwise limited precision like e.g. 10bit will not work correctly.
* Messages now use hashed strings instead of storing the hash in a plain uint.


Code/Engine/Core/Messages/Implementation/Messages.cpp
```diff
@@ -32,7 +32,7 @@ EZ_BEGIN_DYNAMIC_REFLECTED_TYPE(ezMsgComponentInternalTrigger, 1, ezRTTIDefaultA
 {
   EZ_BEGIN_PROPERTIES
   {
-    EZ_MEMBER_PROPERTY("UsageStringHash", m_uiUsageStringHash)
+    EZ_ACCESSOR_PROPERTY("UsageStringHash", GetUsageStringHash, SetUsageStringHash)
```
jankrassnigg: Have you checked that the samples still work? Unfortunately in TypeScript there is no 64 bit int support (because all numbers are doubles), that's why I had kept this as a 32 bit hash for now. TypeScript also has a function for hashing strings, to do this comparison, and that only returns a 32 bit value.

Code/Engine/Foundation/Algorithm/StringHash.h
```diff
@@ -0,0 +1,49 @@
...
+#pragma once
+
+#include <Foundation/Basics.h>
+
+/// \brief Wrapper class to provide type safety for string hashes
+struct ezStringHash
+{
+  constexpr ezStringHash() = default;
+
+  EZ_ALWAYS_INLINE constexpr ezStringHash(ezUInt64 uiValue)
```
jankrassnigg: explicit

Code/Engine/Foundation/Algorithm/StringHash.h
```diff
@@ -0,0 +1,49 @@
...
+    : m_uiValue(uiValue)
+  {
+  }
+
+  EZ_ALWAYS_INLINE constexpr operator ezUInt64() const { return m_uiValue; }
+  EZ_ALWAYS_INLINE constexpr operator ezInt64() const { return m_uiValue; }
```
jankrassnigg: Why the int64 overload ?

Code/Engine/Foundation/Algorithm/StringHash.h
```diff
@@ -0,0 +1,49 @@
...
+  EZ_ALWAYS_INLINE constexpr bool operator<(const ezStringHash& other) const { return m_uiValue < other.m_uiValue; }
+  EZ_ALWAYS_INLINE constexpr bool operator>(const ezStringHash& other) const { return m_uiValue > other.m_uiValue; }
+
+  EZ_ALWAYS_INLINE constexpr bool operator==(ezUInt64 uiValue) const { return m_uiValue == uiValue; }
+  EZ_ALWAYS_INLINE constexpr bool operator!=(ezUInt64 uiValue) const { return m_uiValue != uiValue; }
+  EZ_ALWAYS_INLINE constexpr bool operator<(ezUInt64 uiValue) const { return m_uiValue < uiValue; }
+  EZ_ALWAYS_INLINE constexpr bool operator>(ezUInt64 uiValue) const { return m_uiValue > uiValue; }
+
+private:
+  ezStringHash(ezUInt32 uiValue);
```
jankrassnigg: = delete ?

Code/Engine/Foundation/Algorithm/StringHash.h
```diff
@@ -0,0 +1,49 @@
...
+  constexpr ezStringHash() = default;
+
+  EZ_ALWAYS_INLINE constexpr ezStringHash(ezUInt64 uiValue)
+    : m_uiValue(uiValue)
+  {
+  }
+
+  EZ_ALWAYS_INLINE constexpr operator ezUInt64() const { return m_uiValue; }
+  EZ_ALWAYS_INLINE constexpr operator ezInt64() const { return m_uiValue; }
+
```
jankrassnigg: You could add a "To32Bit" utility function.

Code/Engine/Foundation/Strings/Implementation/FormatString.cpp
```diff
@@ -341,6 +341,11 @@ ezStringView ezArgSensitive::BuildString_SensitiveUserData_Hash(char* tmp, ezUIn
   return tmp;
 }
 
+ezStringView BuildString(char* tmp, ezUInt32 uiLength, ezStringHash arg)
+{
+  return BuildString(tmp, uiLength, (ezUInt64)arg);
```
jankrassnigg: Format as HEX ?

Code/Engine/GameEngine/Gameplay/Implementation/InputComponent.cpp
```diff
@@ -19,7 +19,7 @@ EZ_BEGIN_DYNAMIC_REFLECTED_TYPE(ezMsgInputActionTriggered, 1, ezRTTIDefaultAlloc
 {
   EZ_BEGIN_PROPERTIES
   {
-    EZ_MEMBER_PROPERTY("InputActionHash", m_uiInputActionHash),
```
jankrassnigg, line 22: Same problem here, this will immediately break the input in the Testing Chamber sample.

jankrassnigg (changes requested):

Code/Editor/EditorEngineProcess/EngineProcGameApp.cpp
```diff
@@ -349,11 +349,7 @@ void ezEngineProcessGameApplication::EventHandlerIPC(const ezEngineProcessCommun
     if (pMsg->m_bDocumentOpen)
     {
       pDocumentContext = CreateDocumentContext(pMsg);
-
-      if (pDocumentContext == nullptr)
-      {
-        ezLog::Error("Could not create a document context for document type '{0}'", pMsg->m_sDocumentType);
-      }
+      // EZ_ASSERT_DEV(pDocumentContext != nullptr, "Could not create a document context for document type '{0}'", pMsg->m_sDocumentType);
```
jankrassnigg: Why did you undo this ?

jankrassnigg (approved):

Code/Editor/EditorEngineProcess/EngineProcGameApp.cpp
C-Core: Because some assets don't have a document context (all that don't have a preview) and for them it's totally fine. I couldn't find a good way to tell whether an asset needs a context or not.

PR #386, opened 2021-03-31
property anim component robustness

C-Core:
Made property anim component a bit more robust against runtime changes of the animation range.

### DIFF
--- a/Code/Engine/GameEngine/Animation/Implementation/PropertyAnimComponent.cpp
+++ b/Code/Engine/GameEngine/Animation/Implementation/PropertyAnimComponent.cpp
@@ -122,7 +122,6 @@ void ezPropertyAnimComponent::PlayAnimationRange(ezTime RangeLow, ezTime RangeHi
   m_AnimationRangeLow = RangeLow;
   m_AnimationRangeHigh = RangeHigh;
 
-  m_AnimationTime = m_AnimationRangeLow;
   m_bPlaying = true;
 
   StartPlayback();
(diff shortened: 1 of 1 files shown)


Code/Engine/GameEngine/Animation/Implementation/PropertyAnimComponent.cpp
```diff
@@ -122,7 +122,6 @@ void ezPropertyAnimComponent::PlayAnimationRange(ezTime RangeLow, ezTime RangeHi
   m_AnimationRangeLow = RangeLow;
   m_AnimationRangeHigh = RangeHigh;
 
-  m_AnimationTime = m_AnimationRangeLow;
```
C-Core, line 125: Don't set animation time here since animation range low is not yet clamped against the actual anim duration.

Code/Engine/GameEngine/Animation/Implementation/PropertyAnimComponent.cpp
```diff
@@ -443,10 +442,16 @@ void ezPropertyAnimComponent::ApplyAnimations(const ezTime& tDiff)
 
 ezTime ezPropertyAnimComponent::ComputeAnimationLookup(ezTime tDiff)
 {
+  m_AnimationRangeLow = ezMath::Clamp(m_AnimationRangeLow, ezTime::Zero(), m_AnimDesc->m_AnimationDuration);
```
C-Core, line 445: Clamp range here again since the range can be modified at runtime and the user doesn't know the actual duration

Code/Engine/GameEngine/Animation/Implementation/PropertyAnimComponent.cpp
```diff
@@ -582,6 +587,10 @@ void ezPropertyAnimComponent::StartPlayback()
   {
     m_AnimationTime = m_AnimationRangeHigh;
   }
+  else
+  {
+    m_AnimationTime = m_AnimationRangeLow;
```
C-Core, line 592: Animation time is set here instead of in PlayAnimationRange since the range is now clamped to the correct duration.

jankrassnigg (approved):

PR #388, opened 2021-04-10
Only declare USE_TEXCOORD0, USE_NORMAL and USE_TANGENT when used

leapach:
Only declare USE_TEXCOORD0, USE_NORMAL and USE_TANGENT when the input pins are actually used.

This fixes VSE shaders for minimal models which don't have these vertex streams set.


marcappelsmeier:
This actually breaks the default input values at the moment for nodes like BaseTexture as it wouldn't define USE_TEXCOORD.

jankrassnigg (approved):

PR #401, opened 2021-05-30
Added reflection probe support

### LINKED ISSUE #383: [BUG] Mipmap more 'Kaiser' generates a warning on import
**Bug description:**
Warning: Option '-mipmaps' selected value 'Kaiser' is unknown. Using default value instead.

**How to Reproduce:**
Select 'Kaiser' for mipmaps when transforming a texture.

**Screenshots:**
If applicable, add screenshots to help explain the problem.

image

**Dev-branch commit hash:**
475a918bd80191cf314b34fdbe3c4792f620a34f

Sanakan8472:
**Reflection probe support**
image

- Only the global fallback reflection probe is supported per world right now.
- Reflection probe is set up via ezSkyLightComponent. Both dynamic updates and static cube maps are supported.
- Static cube maps must have mipmaps defined, roughness filtering is done on the fly at runtime for now.
- If no ezSkyLightComponent exists in a world, a black fallback probe is used.

**Generalized lighting settings for simple asset documents**
image

- ezEngineViewLightSettings in a sync object that controls light settings of the engine side of the document.
- Settings can be changed via the actions in the engine view toolbar.
- Settings are persisted globally. These can be changed in the  and can be changed in the ezEditorPreferencesUser or copied from an engine view via the action in the toolbar.
- Settings will affect thumbnail generation except for skybox which is hidden in thumbnails.

**Rendering changes**
- Support for texture readback of mipmaps and cube maps.
- Added ezSimplifiedDataExtractor / ezSimplifiedDataProvider to support the reflection probe / ambient cube in the simple render pipeline.
- If no texture is set in the render target setup, the render pipeline will suplement this via the texture pool.
- Expose ezRenderContext::GetAllowAsyncShaderLoading to allow swapping the state conditionally. Used in the reflection filter pass to make sure it can never fail.
- Reflection probe debug info now shows the filtered mip maps as well.
- Updated all shader references to distinguish 'roughness' from 'perceptual roughness'. Roughness is the physical property while perceptual roughness is what is set by the user in the editor and also maps to the mip levels of the reflection probes. perceptualRoughness^2 == roughness.

**Misc**
- Bug fix: Cloning materials or other prefab documents now preserves the prefab references / seeds.
- Bug fix: Mesh asset now listens for resource changes to update the AABB to make sure thumbnail generation works correctly after size changes.
- Bug fix: staging texture has D3D11_RESOURCE_MISC_GENERATE_MIPS removed as that prevents texture readback.


Data/Base/Shaders/Common/BRDF.h
```diff
@@ -34,36 +34,56 @@ void AccumulateLight(inout AccumulatedLight result, AccumulatedLight light, floa
...
+  float f = ( NdotH * a2 - NdotH ) * NdotH + 1.0f;
+  return a2 / ( f * f );
 }
 
 float VisibilitySmithCorrelated( float roughness, float NdotV, float NdotL )
 {
-  float a = roughness * roughness;
-  float lambdaV = NdotL * sqrt((-NdotV * a + NdotV) * NdotV + a);
-  float lambdaL = NdotV * sqrt((-NdotL * a + NdotL) * NdotL + a);
+  float a2 = roughness * roughness;
```
C-Core, line 78: This looks wrong, this should be a = roughness; I guess?

Code/Engine/RendererCore/Pipeline/Implementation/Passes/ReflectionFilterPass.cpp
```diff
@@ -141,13 +179,47 @@ void ezReflectionFilterPass::SetInputCubemap(ezUInt32 uiCubemapHandle)
   m_hInputCubemap = ezGALTextureHandle(ezGAL::ez18_14Id(uiCubemapHandle));
 }
 
+ezVec4 GetV4(ezVec3 data)
```
C-Core, line 182: ezVec3 has a .GetAsVec4 method

jankrassnigg:
"Enable Fog" makes the skybox disappear. I guess you'd need to remove the flag to apply fog to the sky.

jankrassnigg:
By default the sky light is ON but the skybox is OFF. IMO the default should be for both to be ON. However, I'm wondering whether you'd maybe want different defaults per document type?

jankrassnigg:
> 
> 
> By default the sky light is ON but the skybox is OFF. IMO the default should be for both to be ON. However, I'm wondering whether you'd maybe want different defaults per document type?

Well, I see it is stored per document. In that case I'm definitely missing a "reset to defaults" action.

jankrassnigg:
The directional light comes from the back (along positive X). It should come from the front, from where the camera is looking at the meshes:

image

jankrassnigg:
> 
> 
> The directional light comes from the back (along positive X). It should come from the front, from where the camera is looking at the meshes:
> 
> image

This is different in the material assets, there the default camera angle looks at the lit side. I think the material camera needs to be adjusted, the camera angle in the meshes is the correct one.

jankrassnigg:
Scenes and Prefabs have a similar feature (View > Add Ambient Light) which is ON by default for prefabs and OFF by default for scenes (and this state is reset when the document is closed). I wonder whether we should replace that with the new option. It would need to have different behavior though.

jankrassnigg:
Switching the mode on the skylight component gave me an error:

image

jankrassnigg:
The 'Include/Exclude Tags' are only needed for the dynamic mode right? What about the cubemap option, is that static only? Would be nice if the options only show up depending on the selected mode.

Also I guess the "Show Debug" option shouldn't be in the sub group, since it is always viable, no (enabling that also consistently crashes for me).

Data/Base/Shaders/Common/BRDF.h
Sanakan8472: No this was previously wrong already, see: https://google.github.io/filament/Filament.html#listing_specularv

Sanakan8472:
> 
> 
> "Enable Fog" makes the skybox disappear. I guess you'd need to remove the flag to apply fog to the sky.

I moved all the light settings out of the engine context. The fog settings come from the particle context, the skybox from the kraut context. In combination they don't make much sense but I just wanted to preserve the current functionality for now. We could extend the fog settings?

Sanakan8472:
> 
> 
> By default the sky light is ON but the skybox is OFF. IMO the default should be for both to be ON. However, I'm wondering whether you'd maybe want different defaults per document type?

For now I kept it close to the previous state. Settings can currently be saved globally so it's easy to change the default. I would like to make the defaults per document but that requires adding a new option for it to the preferences system.

Code/Engine/RendererCore/Lights/Implementation/SkyLightComponent.cpp
```diff
@@ -8,14 +8,17 @@
...
 
 // clang-format off
-EZ_BEGIN_COMPONENT_TYPE(ezSkyLightComponent, 1, ezComponentMode::Dynamic)
+EZ_BEGIN_COMPONENT_TYPE(ezSkyLightComponent, 2, ezComponentMode::Dynamic)
 {
   EZ_BEGIN_PROPERTIES
   {
     EZ_ACCESSOR_PROPERTY("Intensity", GetIntensity, SetIntensity)->AddAttributes(new ezClampValueAttribute(0.0f, ezVariant()), new ezDefaultValueAttribute(1.0f)),
     EZ_ACCESSOR_PROPERTY("Saturation", GetSaturation, SetSaturation)->AddAttributes(new ezClampValueAttribute(0.0f, ezVariant()), new ezDefaultValueAttribute(1.0f)),
+    EZ_ENUM_ACCESSOR_PROPERTY("ReflectionProbeMode", ezReflectionProbeMode, GetReflectionProbeMode, SetReflectionProbeMode),
```
C-Core: Default value here should be dynamic, it is currently static thus skylight doesn't work anymore in existing scenes.

ChristopherManthei: Thanks! Fixed, that also caused the test to fail.

jankrassnigg:
This PR also fixes #383

PR #402, opened 2021-06-01
Add a splash screen using our new logo :)

leapach:


Code/Editor/EditorFramework/EditorApp/EditorApp.moc.h
```diff
@@ -269,7 +273,9 @@ private Q_SLOTS:
   QApplication* s_pQtApplication;
   ezLongOpControllerManager m_LongOpControllerManager;
   ezEditorEngineProcessConnection* s_pEngineViewProcess;
-  QTimer* m_pTimer;
+  QTimer* m_pTimer;
```
jankrassnigg: We should really just initialize the variables here. The existing code ist just old and therefore doesn't use that feature yet.

Code/Editor/EditorFramework/EditorApp/EditorApp.cpp
```diff
@@ -53,6 +57,9 @@ void ezQtEditorApp::SlotTimedUpdate()
   }
   ezTaskSystem::FinishFrameTasks();
 
+  // Close the splash screen when we get to the first idle event
+  CloseSplashScreen();
```
jankrassnigg, line 55: Shouldn't it get closed when the first idle event arrives, not when we send it?

Code/Editor/EditorFramework/EditorApp/Startup.cpp
```diff
@@ -414,3 +417,26 @@ void ezQtEditorApp::CreatePanels()
 
   pAssetBrowserPanel->raise();
 }
+
+
+void ezQtEditorApp::SetupAndShowSplashScreen()
+{
+  EZ_ASSERT_DEV(m_pSplashScreen == nullptr, "Splash screen shouldn't exist already.");
+
+  QPixmap splashPixmap(":/Splash/Splash/splash.png");
```
jankrassnigg: I assume we can't use the SVG directly?

jankrassnigg (approved):

jankrassnigg:
Can you change the logo to use the same color as the icons ?

Code/Editor/EditorFramework/EditorApp/Startup.cpp
jayrulez: https://stackoverflow.com/questions/10079011/qpixmap-and-svg

Code/Editor/EditorFramework/EditorApp/EditorApp.cpp
leapach: The SlotTimedUpdate() function is connected to the idle timer event and thus is called when the message loop is processed.
IdleEvent is then just our relay of that information / timer event happening. So this seemed like a good place to have the quick function call.

Code/Editor/EditorFramework/EditorApp/Startup.cpp
leapach: The problem is that we're on Qt 5.13 - that doesn't have the capability to keep the aspect ratio during SVG rendering. I'd need to hard code the aspect ratio into the code loading the SVG so that the image is not stretched. Not an ideal solution :(

leapach: I switched to a the SVG now and added a TODO to go from the awkward size to a square size and using the aspect ratio mode of the SVG renderer when we migrate to a newer Qt verison.

PR #404, opened 2021-06-03
Asset library improvements

Sanakan8472:
- GameEngineTest now fail on transform failures.
- Removed asset references inside Content that lead outside of it and thus could not be resolved if Content data dir was added to other projects. These have been replaced with material overrides in projects that relied on this.
- Scene thumbnails again contain the sky. Each asset type can configure this by setting the view exclusion tags on the RemoteCreateThumbnail call.
- Added asset browser context menu entries to find all direct and indirect uses of an asset. This is done by a custom syntax in the search field.
image


Code/Editor/EditorFramework/Assets/Implementation/AssetBrowserFilter.cpp
```diff
@@ -61,6 +61,25 @@ void ezQtAssetBrowserFilter::SetTextFilter(const char* szText)
     return;
 
   m_sTextFilter = sCleanText;
+  // Clear uses search cache
+  m_bUsesSearchActive = false;
+  m_bTransitive = false;
+  m_uses.Clear();
+
+  const char* szUsesGuid = ezStringUtils::FindSubString_NoCase(szText, "uses:");
+  const char* szUses2Guid = ezStringUtils::FindSubString_NoCase(szText, "uses2:");
```
jankrassnigg: How about "use:" and "ref:" instead of "uses" and "uses2" ?

Code/Engine/RendererCore/Pipeline/Implementation/Passes/ReflectionFilterPass.cpp
```diff
@@ -205,8 +200,8 @@ void ezReflectionFilterPass::UpdateFilteredSpecularConstantBuffer(ezUInt32 uiMip
   };
 
   auto constants = ezRenderContext::GetConstantBufferData<ezReflectionFilteredSpecularConstants>(m_hFilteredSpecularConstantBuffer);
-  constants->Forward = GetV4(vForward[outputIndex % 6]);
-  constants->Up2 = GetV4(vUp[outputIndex % 6]);
+  constants->Forward = vForward[outputIndex % 6].GetAsVec4(0.0f);
```
jankrassnigg: There's even "GetAsDirectionVec4" IIRC.

Data/UnitTests/GameEngineTest/TypeScript/DataDirectories.ddl
```diff
@@ -22,3 +22,9 @@ DataDir
 	string %RootName{""}
 	bool %Writable{false}
 }
+DataDir
+{
+	string %Path{">sdk/Data/Content"}
+	string %RootName{""}
+	bool %Writable{false}
+}
```
jankrassnigg: These projects don't use the Content data dir, do they? By adding the data directory, the asset transform for these projects may take longer (I guess all assets will only be transformed once, but the redundant projects will still scan the additional assets for changes).

jankrassnigg (approved):
See my comments.

Data/UnitTests/GameEngineTest/PlatformWin/Kraut/cgb-MossyBark01.ezMaterialAsset
```diff
@@ -1,878 +0,0 @@
...
-{
-o
-{
-	Uuid %id{u4{12393651040102028521,4891228368794462065}}
-	s %t{"Kraut/Kraut.ezShader"}
-	u3 %v{2}
-	p
-	{
-		s %BLEND_MODE{"BLEND_MODE::BLEND_MODE_OPAQUE"}
-		s %BaseTexture{"Grey.color"}
```
jankrassnigg: This material was copied over 1:1 and then adjusted to not use a texture. Now you get the original asset and that's why the tree appears textured again in the test and therefore fails now.

Sanakan8472: Ah that explains the guid collision, thanks.

Data/UnitTests/GameEngineTest/TypeScript/DataDirectories.ddl
Sanakan8472: They all do use the >sdk/Data/UnitTests/GameEngineTest/SharedAssets data dir which takes all its textures for its materials from the >sdk/Data/Content directory. Removing that dependency looks a bit more difficult without splitting the data dir in two.

Data/UnitTests/GameEngineTest/PlatformWin/Kraut/cgb-MossyBark01.ezMaterialAsset
Sanakan8472: Ok I've restored it with a new guid and changed the material references fo the tree to the new asset.

Data/UnitTests/GameEngineTest/TypeScript/DataDirectories.ddl
jankrassnigg: That's fine then, wasn't aware of that.

Code/Editor/EditorFramework/Assets/Implementation/AssetBrowserFilter.cpp
Sanakan8472: I was struggling with a good naming pattern here. The ezAssetCurator::FindAllUses finds all assets that have this asset as a transform dependency or reference it in addition to allowing to find indirect uses by creating the transitive hull.
I thought 'use' nicely generalized both refs and deps to the asset but couldn't think of a short form for the indirect case. I think ref would be misleading and wouldn't convey the indirect nuture. Granted, neither does uses2 but at least it isn't missleading.
Only in very rare cases will an asset be a transform dependency of another asset so we could get away with just calling the direct case ref: and the indirect case use: ?

jankrassnigg: In my opinion "use" is the more direct thing. As in "this asset directly uses that asset", whereas "ref" is the more indirect thing, because asset A doesn't itself USE asset B, but still references it through another asset C.

Frankly I don't understand "Only in very rare cases will an asset be a transform dependency of another asset "

Sanakan8472: ezAssetCurator understands and tracks **dependencies** as links between assets that are required to transform an asset. E.g. if A has a dependency on B, B needs to be successfully transformed for A to be able to be transformed.
**References** in the ezAssetCurator on the other hand are links between assets that are only relevant at runtime for the assest to function. E.g. if A has a reference to B, both A and B need to be successfully transformed in order for A to work at runtime.
Thus, I didn't want to reuse the same word for a different meaning which would be confusing.

PR #407, opened 2021-06-04
Added option to modulate fog with sky color

C-Core:
Uses the filtered sky reflection cubemap to modulate the fog color. Sharper mipmaps are used at sky distance and blurrier mipmaps are used towards the camera. This is a very cheap method to emulate scattering and also ensures that fogged geometry blends nicely into the skybox.


jankrassnigg:
Sounds interesting! Can you attach a screenshot ?

C-Core:
ModulateWithSkyColor

PR #427, opened 2021-07-11
Clean up empty objects that were created by PrefabReferenceComponents

### LINKED ISSUE #296: PrefabReferenceComponents never get cleaned up
When objects below are deleted, the ref component stays in the scene graph. Garbage collect empty objects?

jankrassnigg:
Fixed #296

### DIFF
--- a/Code/Engine/Core/Prefabs/Implementation/PrefabReferenceComponent.cpp
+++ b/Code/Engine/Core/Prefabs/Implementation/PrefabReferenceComponent.cpp
@@ -446,33 +446,6 @@ void ezPrefabReferenceComponentManager::Update(const ezWorldModule::UpdateContex
   }
 
   m_ComponentsToUpdate.Clear();
-
-  auto pWorld = GetWorld();
-
-  ezUInt32 uiGcSteps = 0;
-  auto it = GetComponents(m_uiGcStartIndex);
-
-  for (; it.IsValid() && uiGcSteps < 10; it.Next())
-  {
-    ++uiGcSteps;
-
-    if (!it->IsActiveAndSimulating())
-      continue;
-
-    auto pOwner = it->GetOwner();
-    if (pOwner == nullptr)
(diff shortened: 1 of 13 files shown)


Code/Engine/Core/Prefabs/Implementation/PrefabReferenceComponent.cpp
```diff
@@ -446,6 +446,33 @@ void ezPrefabReferenceComponentManager::Update(const ezWorldModule::UpdateContex
   }
 
   m_ComponentsToUpdate.Clear();
+
+  auto pWorld = GetWorld();
+
+  ezUInt32 uiGcSteps = 0;
+  auto it = GetComponents(m_uiGcStartIndex);
+
+  for (; it.IsValid() && uiGcSteps < 10; it.Next())
```
C-Core: I don't think this will work at runtime, because the prefab reference component will delete itself after it has instantiated the prefab, so you will end up with an empty game object without any components.

jankrassnigg: I've tested it in play the game mode and there it worked, but it looks like you are right and the behavior is different in a standalone app. Have to think about that.

jankrassnigg: Another idea would be to set a flag on the game object that this can be garbage collected when empty, and then do this collection on ALL objects.

jankrassnigg: I was thinking about adding a component manager that does some debug vis, anyway, like rendering a marker at the location of every (empty) object, so that you can see when a prefab doesn't clean itself up.  Eg most effects at runtime are supposed to delete themselves, but might be misconfigured.

C-Core (approved):

Issue #430, opened 2021-07-18
[FEATURE] Physical Rope Simulation


jankrassnigg:
Simulate a rope or chain with the physics engine.

Functionality that should work:
* Collide with the scene and other physical objects
* Hang / swing properly
* Connect to static walls
* Connect dynamic objects
* Simulation should be reasonable stable

Functionality that is not critical or not achievable:
* Perfect simulation stability under all circumstances. Masses and forces need to be tweaked by the user.
* Rendering is a separate topic.
* Definition of the rope shape at start.
* No physics tunneling (ropes through other ropes or thin geometry).

PR #432, opened 2021-07-18
Physical Rope Simulation

jankrassnigg:
See #430.

This PR adds two components: One for simulating a rope and one for basic (line) rendering of the rope.
The rendering aspect is only preliminary and needs to be improved a lot.

The physical simulation uses PhysX to achieve a decent simulation and interaction with the rest of the scene. Therefore it is not meant as a super simple and cheap simulation for decorative purposes only. Rather it can be used for game-play purposes (think of the power-cable puzzles in Half-Life 2). However, care must be taken to not choose configurations that make PhysX go completely bonkers. With the default settings one should be fine, though.

Ropes are limited to 64 links (a PhysX restriction).


Code/EnginePlugins/PhysXPlugin/Components/Implementation/PxRopeComponent.cpp
```diff
@@ -0,0 +1,714 @@
...
+#include <foundation/Px.h>
+
+using namespace physx;
+
+// clang-format off
+EZ_BEGIN_COMPONENT_TYPE(ezPxRopeComponent, 1, ezComponentMode::Dynamic)
+  {
+    EZ_BEGIN_PROPERTIES
+    {
+      EZ_ACCESSOR_PROPERTY("AnchorA", DummyGetter, SetAnchorAReference)->AddAttributes(new ezGameObjectReferenceAttribute()),
```
leapach, line 22: Should we make it so that the first anchor is always the game object the rope component is attached to? That would make it coherent with the ray cast and beam component for example.

jankrassnigg: The third position is needed to set up a loose rope. Originally I only had a "slack" value and did it how you suggested, but being able to also control the rotation of the where the rope curves can be useful, especially for ropes that hang mostly downwards.

The documentation shows this: http://ezengine.net/pages/docs/physics/special/physx-rope-component.html#setting-up-a-rope

For many more examples, see the demo level.

jankrassnigg: One could turn this around, and make the center position an optional reference, though. Not sure whether that would make much of a difference in usability. Would be best if you try the demo scene to get a feeling for it.

PR #434, opened 2021-07-23
Use ID3D11Fence for fence implementation instead of ID3D11Query

jayrulez:


Code/Engine/RendererDX11/Device/DeviceDX11.h
```diff
@@ -187,6 +187,7 @@ class EZ_RENDERERDX11_DLL ezGALDeviceDX11 : public ezGALDevice
   };
 
   PerFrameData m_PerFrameData[4];
+  ezUInt64 m_FenceValues[4] = {};
```
C-Core, line 190: Is this really needed? We should be able to use the framecounter instead.

Code/Engine/RendererDX11/CommandEncoder/Implementation/CommandEncoderImplDX11.cpp
```diff
@@ -129,32 +129,37 @@ void ezGALCommandEncoderImplDX11::SetUnorderedAccessViewPlatform(ezUInt32 uiSlot
...
+  //if (m_pDXContext->GetData(static_cast<const ezGALFenceDX11*>(pFence)->GetDXFence(), &data, sizeof(data), 0) == S_OK)
+  //{
+  //  EZ_ASSERT_DEV(data == TRUE, "Implementation error");
+  //  return true;
+  //}
+
+  //return false;
 
-  return false;
+  return pFence->GetCompletedValue() == value; // >=?(but what if value overflows?),
```
C-Core, line 149: This has to be >= otherwise is won't work if GPU is too fast or CPU too slow. Overflow is not a problem for a 64bit value which is typically increased once per frame.

PR #439, opened 2021-07-29
Added a function to get number of frame since on object was last visible

C-Core:
This can be used to skip expensive update functions for invisible objects.

Also some minor visibility culling performance optimizations.

### DIFF
--- a/Code/Engine/Core/World/Implementation/GameObject.cpp
+++ b/Code/Engine/Core/World/Implementation/GameObject.cpp
@@ -721,7 +721,7 @@ ezUInt64 ezGameObject::GetNumFramesSinceVisible() const
     return pSpatialSystem->GetNumFramesSinceVisible(m_pTransformationData->m_hSpatialData);
   }
 
-  return -1;
+  return 0;
 }
 
 void ezGameObject::OnMsgDeleteGameObject(ezMsgDeleteGameObject& msg)
--- a/Code/Engine/Core/World/Implementation/GameObject.cpp
+++ b/Code/Engine/Core/World/Implementation/GameObject.cpp
@@ -714,11 +714,11 @@ void ezGameObject::SetTeamID(ezUInt16 id)
   }
 }
 
-ezUInt64 ezGameObject::GetNumFramesSinceVisible(ezUInt64 uiCurrentFrame) const
+ezUInt64 ezGameObject::GetNumFramesSinceVisible() const
 {
   if (const ezSpatialSystem* pSpatialSystem = GetWorld()->GetSpatialSystem())
   {
-    return pSpatialSystem->GetNumFramesSinceVisible(m_pTransformationData->m_hSpatialData, uiCurrentFrame);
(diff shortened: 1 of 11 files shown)


Code/Engine/Core/World/Implementation/GameObject.cpp
```diff
@@ -714,6 +704,26 @@ void ezGameObject::TryGetComponentsOfBaseType(const ezRTTI* pType, ezDynamicArra
...
+}
+
+ezUInt64 ezGameObject::GetNumFramesSinceVisible() const
+{
+  if (const ezSpatialSystem* pSpatialSystem = GetWorld()->GetSpatialSystem())
+  {
+    return pSpatialSystem->GetNumFramesSinceVisible(m_pTransformationData->m_hSpatialData);
+  }
+
+  return -1;
```
jankrassnigg: Since the return type is a uint64, I'm not sure why the compiler doesn't complain about returning -1 without a cast. Anyway, maybe just return 0? Otherwise code that relies on it would get 0xFF..FF and always think it's invisible.

Code/Engine/Core/World/Implementation/SpatialSystem.cpp
```diff
@@ -217,6 +232,19 @@ void ezSpatialSystem::FindVisibleObjects(
...
+  if (!m_DataTable.TryGetValue(hData.GetInternalID(), pData))
+    return -1;
+
+  if (pData->m_Flags.IsSet(ezSpatialData::Flags::AlwaysVisible))
+  {
+    return 0;
+  }
 
+  const ezUInt64 uiLastFrameVisible = GetLastFrameVisibleInternal(pData);
+  return (m_uiFrameCounter > uiLastFrameVisible) ? m_uiFrameCounter - uiLastFrameVisible : 0;
```
jankrassnigg, line 245: This logic seems redundant.

Code/Engine/Core/World/Implementation/SpatialSystem_RegularGrid.cpp
```diff
@@ -266,12 +272,71 @@ struct ezSpatialSystem_RegularGrid::Cell
     }
   }
 
+  EZ_FORCE_INLINE void UpdateObject(ezSpatialData* pData)
+  {
+    ezUInt32 mask = pData->m_uiCategoryBitmask;
+    ezUInt32 category = ezMath::FirstBitLow(mask);
+    mask &= mask - 1;
```
jankrassnigg, line 279: what is this for ? (the mask &= mask - 1)

jankrassnigg: Ok, I get it now. You want to iterate through all active categories. A comment would be nice, since this logic is not obvious. Potentially even a helper function to put it into.

jankrassnigg (approved):

Code/Engine/Core/World/Implementation/SpatialSystem.cpp
C-Core: It's to prevent overflow if uiLastFrameVisible is 0xFF... or something like that.

PR #443, opened 2021-08-05
Added components to create local wind effects

### LINKED ISSUE #442: [FEATURE] Local wind effects
Make it possible to create localized wind effects, rather than just one global wind value.

For that one needs a few wind volume components, which define the area in which the wind is stronger and to define the wind direction.

jankrassnigg:
Resolves #442


Code/Engine/GameEngine/Effects/Wind/Implementation/SimpleWindWorldModule.cpp
```diff
@@ -19,6 +22,27 @@ ezSimpleWindWorldModule::~ezSimpleWindWorldModule() = default;
 
 ezVec3 ezSimpleWindWorldModule::GetWindAt(const ezVec3& vPosition) const
 {
+  if (auto pSpatial = GetWorld()->GetSpatialSystem())
```
jankrassnigg, line 25: @C-Core This is really the core piece for implementing this feature. If you want to have support for the wind volume shapes in your custom wind system, you could just copy these lines.

C-Core: How about moving that into a static helper function or something like that?

jankrassnigg: Well, it is quite likely that you'd do adjustments, e.g. to integrate the volume's output into your wind field simulation. Copy and paste would just be the simplest way to get started.

PR #445, opened 2021-08-07
Added a help button for components

### LINKED ISSUE #444: [DOCS] Add help buttons to editor to open component-specific docs

jankrassnigg:
The ezQtPropertyTypeContainerWidget now adds a help button to array entries, when the translation system yields a non-empty URL string for the type. It's using the translation system, so that URLs could theoretically even be language specific (not that we would ever need that, but it kinda makes sense).

@Sanakan8472 I'm not sure the location and placement of the help button is the best solution. I wanted it to be visible, rather than have it in a context menu, because otherwise people wouldn't be aware of it. Still, it takes up a lot of space and might be a bit ugly next to the delete button. Also the code might be a bit hacked, not sure whether there is a better solution.

I'd also like it, if I had similar buttons for each panel (asset browser, logs, etc) and dialogs, but I don't have a good idea how to integrate it there consistently and without too much manual effort.

Fixes #444

### DIFF
--- a/Code/Engine/Foundation/Strings/Implementation/TranslationLookup.cpp
+++ b/Code/Engine/Foundation/Strings/Implementation/TranslationLookup.cpp
@@ -125,7 +125,9 @@ void ezTranslatorFromFiles::LoadTranslationFile(const char* szFullPath)
   ezDeque<ezString> Lines;
   sContent.Split(false, Lines, "\n");
 
-  ezStringBuilder sLine, sKey, sValue, sTooltip;
+  ezHybridArray<ezString, 4> entries;
+
+  ezStringBuilder sLine, sKey, sValue, sTooltip, sHelpUrl;
   for (const auto& line : Lines)
   {
     sLine = line;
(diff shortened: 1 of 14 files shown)


Code/Engine/Foundation/Strings/Implementation/TranslationLookup.cpp
```diff
@@ -125,7 +125,9 @@ void ezTranslatorFromFiles::LoadTranslationFile(const char* szFullPath)
   ezDeque<ezString> Lines;
   sContent.Split(false, Lines, "\n");
 
-  ezStringBuilder sLine, sKey, sValue, sTooltip;
+  ezHybridArray<ezString, 4> entries;
```
Sanakan8472: Nit: Make this an ezStringView array to save a bunch of unnecessary copies.

Code/Tools/Libs/GuiFoundation/PropertyGrid/Implementation/PropertyBaseWidget.cpp
```diff
@@ -1253,18 +1244,34 @@ void ezQtPropertyTypeContainerWidget::UpdateElement(ezUInt32 index)
...
       elem.m_pSubGroup->SetIcon(ezQtUiServices::GetCachedIconResource(sIconName.GetData()));
     }
+
+    // help URL
+    {
+      QString url = ezTranslateHelpURL(pCommonType->GetTypeName());
+
+      if (!url.isEmpty())
+      {
+        ezQtElementGroupButton* pHelp = new ezQtElementGroupButton(elem.m_pSubGroup->GetHeader(), ezQtElementGroupButton::ElementAction::Help, this);
```
Sanakan8472: Don't create any widgets in this function. This function will be called multiple times for the same instance and other instances  so you accumulate buttons constantly and trigger layout / repaint unnecessarily. Better follow the example of the delete button and create the button at the start and then just hide it depending on whether it is needed for the current selection.

Code/Tools/Libs/GuiFoundation/PropertyGrid/Implementation/ElementGroupButton.moc.h
```diff
@@ -13,6 +13,7 @@ class EZ_GUIFOUNDATION_DLL ezQtElementGroupButton : public QToolButton
     MoveElementUp,
     MoveElementDown,
     DeleteElement,
+    Help,
```
Sanakan8472, line 16: To make the help button less prominent you can set setAutoRaise(true); for the ezQtElementGroupButton. It might even be a good idea to set that for all types.

Code/Engine/Foundation/Strings/Implementation/TranslationLookup.cpp
jankrassnigg: good catch !

Code/Tools/Libs/GuiFoundation/PropertyGrid/Implementation/PropertyBaseWidget.cpp
jankrassnigg: Is this better? Working with this part of the code is a bit tricky, I'm not very familiar with the structure.

PR #453, opened 2021-08-11
Added a component to simulate simple sheets of cloth

jankrassnigg:


Code/Engine/RendererCore/Meshes/Implementation/DynamicMeshBufferResource.cpp
```diff
@@ -0,0 +1,193 @@
...
+    m_VertexDeclaration.m_VertexStreams.PushBack(si);
+
+    si.m_uiOffset += si.m_uiElementSize;
+    si.m_Format = ezGALResourceFormat::XYZFloat;
+    si.m_Semantic = ezGALVertexAttributeSemantic::Normal;
+    si.m_uiElementSize = sizeof(ezVec3);
+    m_VertexDeclaration.m_VertexStreams.PushBack(si);
+
+    si.m_uiOffset += si.m_uiElementSize;
+    si.m_Format = ezGALResourceFormat::XYZFloat;
```
C-Core: This should be ezGALResourceFormat::XYZWFloat and sizeof(ezVec4) below

Code/Engine/RendererCore/Meshes/Implementation/DynamicMeshBufferResource.cpp
```diff
@@ -0,0 +1,193 @@
...
+  res.m_uiQualityLevelsDiscardable = 0;
+  res.m_uiQualityLevelsLoadable = 0;
+  res.m_State = ezResourceState::Loaded;
+
+  return res;
+}
+
+void ezDynamicMeshBufferResource::UpdateGpuBuffer(ezGALCommandEncoder* pGALCommandEncoder, ezUInt32 uiFirstVertex, ezUInt32 uiNumVertices, ezUInt32 uiFirstIndex, ezUInt32 uiNumIndices)
+{
+  // TODO: can ezGALUpdateMode::Discard be used, if only a subset is updated (without destroying existing data) ?
```
C-Core: No discard literally means "discard existing data". If you update only a subset but don't want to destroy existing data you need to use CopyToTempStorage

Code/Engine/RendererCore/Meshes/Implementation/DynamicMeshBufferResource.cpp
```diff
@@ -0,0 +1,193 @@
...
+    {
+      EZ_ASSERT_DEV(uiFirstIndex < m_Index32Data.GetCount(), "Invalid first index value {}", uiFirstIndex);
+      EZ_ASSERT_DEV(uiNumIndices <= m_Index32Data.GetCount(), "Can't upload {} indices, the buffer was allocated to hold a maximum of {} indices.", uiNumIndices, m_Index32Data.GetCount());
+
+      pGALCommandEncoder->UpdateBuffer(m_hIndexBuffer, sizeof(ezUInt32) * uiFirstIndex, m_Index32Data.GetArrayPtr().GetSubArray(uiFirstIndex, uiNumIndices).ToByteArray(), ezGALUpdateMode::Discard);
+    }
+  }
+}
+
+void ezDynamicMeshVertex::EncodeNormal(const ezVec3& normal)
```
C-Core: Maybe make these inline to save the function call. Or you make this a batch processing function.

C-Core (changes requested):

Code/Engine/RendererCore/Meshes/Implementation/DynamicMeshBufferResource.cpp
jankrassnigg: batch processing isn't easy here, though you do get public access to the actual data, in case the calling code can implement that itself

C-Core (approved):

PR #456, opened 2021-08-13
Added locking flags to the dynamic PhysX actors

### LINKED ISSUE #451: [FEATURE] Allow setting PxRigidDynamicLockFlag on the PxDynamicActorComponent
**Is your feature request related to a problem? Please describe:**
I want to implement certain mechanics using physics e.g.: Collision between rigid body's with some simulation properties disabled like being able to bump into a capsule which is a rigid body without tipping it over or affecting its rotation. 

**Describe the solution you'd like:**
Allow locking the actor's position or individual components of the position.
Allow locking the actor's rotation or individual components of the rotation.

**Additional context:**
Expose this through component properties that can be changed from code or from the editor UI.

Example screenshot from Unity:
image

jankrassnigg:
Fixes #451

### DIFF
--- a/Code/EnginePlugins/PhysXPlugin/Components/Implementation/PxDynamicActorComponent.cpp
+++ b/Code/EnginePlugins/PhysXPlugin/Components/Implementation/PxDynamicActorComponent.cpp
@@ -106,7 +106,16 @@ void ezPxDynamicActorComponentManager::UpdateMaxDepenetrationVelocity(float fMax
 //////////////////////////////////////////////////////////////////////////
 
 // clang-format off
-EZ_BEGIN_COMPONENT_TYPE(ezPxDynamicActorComponent, 3, ezComponentMode::Dynamic)
+EZ_BEGIN_STATIC_REFLECTED_BITFLAGS(ezPxActorLockingFlags, 1)
+  EZ_ENUM_CONSTANT(ezPxActorLockingFlags::FreezePositionX),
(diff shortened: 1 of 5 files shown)


Code/EnginePlugins/PhysXPlugin/Components/Implementation/PxDynamicActorComponent.cpp
```diff
@@ -428,6 +451,23 @@ void ezPxDynamicActorComponent::AddAngularImpulse(const ezVec3& vImpulse)
...
 
+void ezPxDynamicActorComponent::SetLockingFlags(ezBitflags<ezPxActorLockingFlags> flags)
+{
+  m_LockingFlags = flags;
+
+  if (m_pActor != nullptr)
+  {
+    EZ_PX_WRITE_LOCK(*(m_pActor->getScene()));
+
+    m_pActor->setRigidDynamicLockFlag(PxRigidDynamicLockFlag::eLOCK_LINEAR_X, m_LockingFlags.IsSet(ezPxActorLockingFlags::FreezePositionX));
```
leapach: There is setRigidDynamicLockFlags() which takes all flags in one go. If we mirror the flag values in our enum we could just cast it and do one function call.

jankrassnigg: done

PR #461, opened 2021-08-20
Improved physics surface interactions

### LINKED ISSUE #231: Surface 'slide' and 'roll' prefabs are hard coded
See ezPhysXWorldModule::FetchResults

roll interactions is hard coded to prefab { 4d306cc5-c1e6-4ec9-a04d-b804e3755210 }
and slide to { c2d8d66d-b123-4cf1-b123-4d015fc69fb0 }

jankrassnigg:
Fixes #231 

Surface interactions are when contact points from the physics engine are taken to spawn prefabs. There are three cases:
1. Impacts. Those are spawned once in case the impact force is large enough.
2. Slides. These are started and stopped when an object slides perpendicular to another surface.
3. Rolls. These are started and stopped when an object touches another surface and rotates.

All of these have to be enabled via flags on the physics shapes, as it is difficult to know whether a rotation around a certain axis should be considered a "roll" interaction.

So far the prefabs to use for slide and roll interactions were hardcoded. This is now configurable in the surface assets. So you can select a different slide or roll effect by material type.

Additionally, so far ALL contact points were used to spawn interactions. With large numbers of rigid bodies colliding (1000+) this could become a major performance bottleneck (seconds per frame...).

Now there is a maximum number of interactions that may be active at any time (slide or roll) or get spawned each frame (impacts). Those numbers are pretty low (default is 4, can be configured via CVar). With a simple test the best contact points are chosen, and the result is still pretty good. Performance-wise it is now so fast, that it doesn't even show up in a profiling capture anymore, not even with 3000+ physics objects in a pile.

Additionally, there are now a few cvars to visualize the used and discarded contact points.

And finally, this PR adds 'persistent' debug rendering shapes to the ezDebugRenderer, which makes it easy to render a cross, sphere or box in a place for a few seconds, without having to reissue the same draw command manually every frame.


Code/Engine/RendererCore/Debug/DebugRenderer.h
```diff
@@ -70,37 +71,59 @@ class EZ_RENDERERCORE_DLL ezDebugRenderer
...
+  /// \brief Renders a filled 2D rectangle in screenspace for one frame.
   static void Draw2DRectangle(const ezDebugRendererContext& context, const ezRectFloat& rectInPixel, float fDepth, const ezColor& color);
 
+  /// \brief Renders a textured 2D rectangle in screenspace for one frame.
   static void Draw2DRectangle(const ezDebugRendererContext& context, const ezRectFloat& rectInPixel, float fDepth, const ezColor& color, const ezTexture2DResourceHandle& hTexture);
 
-  static void Draw2DText(
-    const ezDebugRendererContext& context, const ezStringView& text, const ezVec2I32& positionInPixel, const ezColor& color, ezUInt32 uiSizeInPixel = 16, HorizontalAlignment::Enum horizontalAlignment = HorizontalAlignment::Left, VerticalAlignment::Enum verticalAlignment = VerticalAlignment::Top);
+  /// \brief Displays a string in screenspace for one frame.
+  static void Draw2DText(const ezDebugRendererContext& context, const ezFormatString& text, const ezVec2I32& positionInPixel, const ezColor& color, ezUInt32 uiSizeInPixel = 16, HorizontalAlignment::Enum horizontalAlignment = HorizontalAlignment::Left, VerticalAlignment::Enum verticalAlignment = VerticalAlignment::Top);
```
C-Core, line 114: What's the benefit of formatstring here?

Code/EnginePlugins/PhysXPlugin/WorldModule/Implementation/PhysXWorldModule.cpp
```diff
@@ -1085,6 +738,16 @@ void ezPhysXWorldModule::StartSimulation(const ezWorldModule::UpdateContext& con
 
 void ezPhysXWorldModule::FetchResults(const ezWorldModule::UpdateContext& context)
 {
+  EZ_PROFILE_SCOPE("FetchResults");
+
+  if (const ezCameraComponentManager* pCompMan = GetWorld()->GetComponentManager<ezCameraComponentManager>())
```
C-Core: Use ezRenderWorld::GetViewByUsageHint with mainview and editor view and the world instead. Then get the camera and position from it. There is no guarantee that there is a camera component for the main view. And above will also work in simulation mode in the editor.

Code/Engine/RendererCore/Debug/Implementation/DebugRenderer.cpp
```diff
@@ -806,6 +848,42 @@ void ezDebugRenderer::Draw3DText(const ezDebugRendererContext& context, const ez
   });
 }
 
+void ezDebugRenderer::AddPersistentCross(const ezDebugRendererContext& context, float fSize, const ezColor& color, const ezTransform& transform, ezTime duration)
+{
+  EZ_LOCK(s_Mutex);
```
C-Core, line 853: Maybe use a separate mutex for persistent stuff to prevent contention.

C-Core (approved):

Code/Engine/RendererCore/Debug/DebugRenderer.h
jankrassnigg: We have many places where we use a stringbuilder first to format something and then pass it to Draw2DText. This way, you can now do this directly with ezFmt("bla {}", 42). It's just more convenient.

Code/Engine/RendererCore/Debug/Implementation/DebugRenderer.cpp
jankrassnigg: I considered that, but in ezDebugRenderer::RenderInternal we call all the "DrawCross" etc functions, which again go through this mutex, so I think it's more efficient to get that lock once, instead of multiple times. And then having a second mutex doesn't gain you anything.

Code/EnginePlugins/PhysXPlugin/WorldModule/Implementation/PhysXWorldModule.cpp
jankrassnigg: Works indeed better.

PR #466, opened 2021-08-26
Merge current state of probe baking

C-Core:
Not usable at the moment and still heavily work in progress


Code/Engine/Core/World/Implementation/World.cpp
```diff
@@ -632,7 +632,7 @@ ezWorldModule* ezWorld::GetModule(const ezRTTI* pRtti)
   return nullptr;
 }
 
-const ezWorldModule* ezWorld::GetModuleReadOnly(const ezRTTI* pRtti) const
+const ezWorldModule* ezWorld::GetModule(const ezRTTI* pRtti) const
```
jankrassnigg, line 635: I'd prefer to keep the other name, because the const and non const methods have different behavior.

Code/Engine/RendererCore/BakedProbes/Implementation/BakedProbesComponent.cpp
```diff
@@ -0,0 +1,472 @@
...
+
+      ezStringBuilder sb;
+      for (ezUInt32 i = 0; i < EZ_ARRAY_SIZE(indexData.m_probeIndices); ++i)
+      {
+        ezVec3 pos = pProbeTree->GetProbePositions()[indexData.m_probeIndices[i]];
+        ezDebugRenderer::DrawCross(msg.m_pView->GetHandle(), pos, 0.5f, ezColor::Yellow);
+
+        sb.Format("Weight: {}", indexData.m_probeWeights[i]);
+        pos.z += 0.5f;
+        ezDebugRenderer::Draw3DText(msg.m_pView->GetHandle(), sb, pos, ezColor::Yellow);
```
jankrassnigg: You could now use formatted printing directly with ezFmt

Code/Engine/RendererCore/BakedProbes/Implementation/BakingUtils.cpp
```diff
@@ -0,0 +1,47 @@
+#include <RendererCore/RendererCorePCH.h>
+
+#include <RendererCore/BakedProbes/BakingUtils.h>
+
+ezVec3 ezBakingUtils::FibonacciSphere(ezUInt32 sampleIndex, ezUInt32 numSamples)
+{
```
jankrassnigg, line 6: Could this be moved somewhere more accessible? Like ezGraphicsUtils maybe? Could be useful for other things as well.

Code/Engine/RendererCore/Debug/Implementation/DebugRenderer.cpp
```diff
@@ -85,8 +85,8 @@ namespace
     ezDynamicArray<Vertex, ezAlignedAllocatorWrapper> m_line2DVertices;
     ezDynamicArray<BoxData, ezAlignedAllocatorWrapper> m_lineBoxes;
     ezDynamicArray<BoxData, ezAlignedAllocatorWrapper> m_solidBoxes;
-    ezMap<ezTexture2DResourceHandle, ezDynamicArray<TexVertex, ezAlignedAllocatorWrapper>> m_texTriangle2DVertices;
-    ezMap<ezTexture2DResourceHandle, ezDynamicArray<TexVertex, ezAlignedAllocatorWrapper>> m_texTriangle3DVertices;
+    ezMap<ezGALResourceViewHandle, ezDynamicArray<TexVertex, ezAlignedAllocatorWrapper>> m_texTriangle2DVertices;
+    ezMap<ezGALResourceViewHandle, ezDynamicArray<TexVertex, ezAlignedAllocatorWrapper>> m_texTriangle3DVertices;
```
jankrassnigg, line 89: Why? Doesn't that risk unloading the texture?

Code/Engine/Core/World/Implementation/World.cpp
C-Core: There are both variants now. I've re-added the GetModule const method for consistency, since we don't use the ReadOnly suffix elsewhere. (we would need to add it to all const methods of ezWorld and ezGameObject, potentially ezComponent etc)

Code/Engine/RendererCore/Debug/Implementation/DebugRenderer.cpp
C-Core: I've added a Draw2DRectangle variant which only takes a ezGALResourceViewHandle for rendertargets so this is the common denominator. I don't think that unloading the texture is a problem since there is someone outside which keeps a texture resource handle.

Code/Engine/RendererCore/BakedProbes/Implementation/BakingUtils.cpp
C-Core: Maybe later, we don't have a good place at the moment. ezGraphicsUtils doesn't really fit. The FibonacciSphere is mostly useful for sampling and we don't have a common place for that I think.

Issue #239, opened 2020-05-21
Ability to set scene for the player app via other machanisms besides command line


Sanakan8472:
For platforms that have limited or no command line argument support it would be nice to be able to start the player and set the scene either via fileserve or inspector protocol. Ideally allowing for the same player run to be reused to load another scene afterwards.

PR #469, opened 2021-09-04
Adds a console to the CVars window (Editor & ezInspector)

### LINKED ISSUE #224: Send game console commands via ezInspector
It would be interesting, if the ezInspector had a "console" window, where you can type console commands, use the auto-completion feature and so on, just the same way as the ingame console allows you to do.

For this one should send the typed input to the connected process, and the output back to the inspector. This should probably ignore the ezLog output (contrary to the ingame console which prints that one as well).

The networking code shouldn't be hard to implement, probably the Qt UI work in the ezInspector is the most effort.

jankrassnigg:
The console can be used the same way as the in-game Quake-style console:

* Hit TAB to auto-complete
* Enter to run a command
* Arrow up/down for input history
* F2, F3 to repeat last and second-to-last command

Commands are sent to the engine process to be executed there. Auto-complete as well. That means all CVars and console functions that are known to the engine process are available, not only the ones known by the local process.

Fixes #224 

image

### DIFF
--- a/Code/Editor/EditorEngineProcess/EngineProcGameApp.cpp
+++ b/Code/Editor/EditorEngineProcess/EngineProcGameApp.cpp
@@ -336,7 +336,7 @@ void ezEngineProcessGameApplication::EventHandlerIPC(const ezEngineProcessCommun
       ezCommandInterpreterState s;
       s.m_sInput = pMsg->m_sCommand;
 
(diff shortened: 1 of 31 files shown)


jankrassnigg:
@Sanakan8472 I'll also take a look at #239, this is the basis for that feature. Now we're able to add a "LoadScene" command that can be executed remotely while the engine is already running.

Issue #470, opened 2021-09-04
[FEATURE] DebugRenderer ordered text rendering


jankrassnigg:
The DebugRenderer should have a mode to add text to the screen, which automatically places the text in a non-overlapping position with other text.

E.g. append text, maybe structured in a hierarchy.

C-Core:
What would be the use case? This would require full text layouting which is beyond the scope of the debug renderer I think.

jankrassnigg:
The use case would be to enable multiple debug outputs and have them automatically show up below each other, daher than having to guess a good position for each. Also it would be resolution independent. 

I don't intend to do layouting, but rather a very simply mechanism of specifying a screen corner as the anchor and then draw the strings beneath each other there.  There might still be overlap, but printing 10 different things on screen would still be easier. 

For example the camera stats would be a use case. Actually basically everything that currently does Draw2DText.

jankrassnigg:
Fixed by #473

PR #473, opened 2021-09-05
Adds DebugRenderer::DrawInfoText for automatic placement

jankrassnigg:
The new function places text in one of the 4 screen corners or top/center or bottom/center.

Position is determined automatically by how much other text is already placed there.
Size is fixed.

This is useful for outputting stats where the exact placement is not important, but it should always be readable.

Example:

image

These are 4 DrawInfoText calls. One uses a single multi-line string. The other three are separate calls. There are two "groups" of strings, and there is an empty line between groups.


jankrassnigg:
@C-Core the code (and comments) are not 100% finished yet, but this is what I meant with #470.

Code/Engine/RendererCore/Debug/Implementation/DebugRenderer.cpp
```diff
@@ -1252,6 +1281,58 @@ void ezDebugRenderer::RenderInternal(const ezDebugRendererContext& context, cons
     }
   }
 
+  // info text
```
C-Core: You have to do this whole code block before the framecounter is set to the current frame:
pDoubleBufferedContextData->m_uiLastRenderedFrame = ezRenderWorld::GetFrameCounter();
otherwise you introduce a frame delay when you call Draw2DText here

Code/Engine/RendererCore/Debug/Implementation/DebugRenderer.cpp
```diff
@@ -1252,6 +1281,58 @@ void ezDebugRenderer::RenderInternal(const ezDebugRendererContext& context, cons
...
+      ezVec2I32(resX - 10, 10),
+      ezVec2I32(10, resY - 10),
+      ezVec2I32(resX / 2, resY - 10),
+      ezVec2I32(resX - 10, resY - 10)};
+
+    for (ezUInt32 corner = 0; corner < (ezUInt32)ezDebugRenderer::ScreenPlacement::ENUM_COUNT; ++corner)
+    {
+      auto& cd = pData->m_infoTextData[corner];
+
+      cd.Sort([](const InfoTextData& lhs, const InfoTextData& rhs) -> bool { return lhs.m_group < rhs.m_group; });
```
C-Core: Sort is not stable so text lines might be shuffled within one group

Code/Engine/RendererCore/Pipeline/Implementation/Passes/ForwardRenderPass.cpp
```diff
@@ -113,7 +113,7 @@ void ezForwardRenderPass::SetupPermutationVars(const ezRenderViewContext& render
   ezViewRenderMode::GetDebugText(renderViewContext.m_pViewData->m_ViewRenderMode, sDebugText);
   if (!sDebugText.IsEmpty())
   {
-    ezDebugRenderer::Draw2DText(*renderViewContext.m_pViewDebugContext, sDebugText, ezVec2I32(10, 10), ezColor::White);
+    ezDebugRenderer::DrawInfoText(*renderViewContext.m_pViewDebugContext, ezDebugRenderer::ScreenPlacement::TopLeft, "RenderPassDbg", sDebugText);
```
C-Core: I would leave this text always at a fixed place. This text typically tells what a debug render mode does, which would be odd if this would show up somewhere in the middle of the screen.

Code/Engine/RendererCore/Pipeline/Implementation/Passes/SimpleRenderPass.cpp
```diff
@@ -118,7 +118,7 @@ void ezSimpleRenderPass::Execute(const ezRenderViewContext& renderViewContext, c
 
   if (!m_sMessage.IsEmpty())
   {
-    ezDebugRenderer::Draw2DText(*renderViewContext.m_pViewDebugContext, m_sMessage.GetData(), ezVec2I32(20, 20), ezColor::OrangeRed);
+    ezDebugRenderer::DrawInfoText(*renderViewContext.m_pViewDebugContext, ezDebugRenderer::ScreenPlacement::TopLeft, "SimplePassDbg", m_sMessage.GetData(), ezColor::OrangeRed);
```
C-Core: Same here leave at fixed place.

Code/Engine/RendererCore/BakedProbes/Implementation/BakedProbesComponent.cpp
```diff
@@ -404,7 +404,7 @@ void ezBakedProbesComponent::RenderDebugOverlay()
   ezBakingInterface* pBakingInterface = ezSingletonRegistry::GetSingletonInstance<ezBakingInterface>();
   if (pBakingInterface == nullptr)
   {
-    ezDebugRenderer::Draw2DText(pView->GetHandle(), "Baking Plugin not loaded", ezVec2I32(10, 10), ezColor::OrangeRed);
+    ezDebugRenderer::DrawInfoText(pView->GetHandle(), ezDebugRenderer::ScreenPlacement::TopLeft, "BakingPlugin", "Baking Plugin not loaded", ezColor::OrangeRed);
```
C-Core: This is a warning text which should always be at a fixed place.

C-Core (approved):

Code/Engine/RendererCore/Debug/Implementation/DebugRenderer.cpp
jankrassnigg: actually the sort is not needed at all

Code/Engine/RendererCore/Debug/Implementation/DebugRenderer.cpp
jankrassnigg: I have changed this, but I sometimes see flickering, ie the text is rendered twice in a single frame. I'm wondering, whether that's because of my change here, or whether the text was always potentially inserted twice (mesh asset stats rendering).

Code/Engine/RendererCore/Debug/Implementation/DebugRenderer.cpp
C-Core: The sort is needed, otherwise there is no deterministic ordering. All debug render function can be called from multiple threads.

Code/Engine/RendererCore/Debug/Implementation/DebugRenderer.cpp
C-Core: Actually I was wrong: The code block needs to be before
PerContextData* pData = pDoubleBufferedContextData->m_pData[ezRenderWorld::GetDataIndexForRendering()].Borrow();
That should solve the flickering.

Code/Engine/RendererCore/Debug/Implementation/DebugRenderer.cpp
```diff
@@ -983,14 +1067,66 @@ void ezDebugRenderer::RenderInternal(const ezDebugRendererContext& context, cons
...
   PerContextData* pData = pDoubleBufferedContextData->m_pData[ezRenderWorld::GetDataIndexForRendering()].Borrow();
   if (pData == nullptr)
   {
     return;
   }
 
+  // update the frame counter
+  pDoubleBufferedContextData->m_uiLastRenderedFrame = ezRenderWorld::GetFrameCounter();
+
+  // draw info text
```
C-Core: This is still not the right place. This block ("draw info text") has to be before line 1064. Just like the presistent stuff.

Code/Engine/RendererCore/Debug/Implementation/DebugRenderer.cpp
jankrassnigg: Ah, right. I've switched to insertion sort. Maybe we should have a "StableSort" function on our containers ?

Code/Engine/RendererCore/Debug/Implementation/DebugRenderer.cpp
jankrassnigg: Yes, and since I need pData, I also needed to use GetDataIndexForExtraction actually.

Code/Engine/RendererCore/Debug/Implementation/DebugRenderer.cpp
jankrassnigg: Yep, my previous change was wrong.

Code/Engine/RendererCore/Debug/Implementation/DebugRenderer.cpp
C-Core: Unfortunately there is no fast and stable sort algorithm that works entirely in-place. Insertion sort is stable but also slow^^

Code/Engine/RendererCore/Debug/Implementation/DebugRenderer.cpp
C-Core: Can you add a unit test where you print a framecounter with Draw2DText and DrawInfoText and check that the frame counter is the same.

Code/Engine/RendererCore/Debug/Implementation/DebugRenderer.cpp
jankrassnigg: Yeah, I meant to add "StableSort" as a function to the containers, so that you can pick Sort or StableSort, depending on your needs. Right now it is not obvious that Sort isn't stable, and what to do, if you need it.

Code/Engine/RendererCore/Debug/Implementation/DebugRenderer.cpp
jankrassnigg: So I just noticed that using the IndexForRendering, some strings (e.g. the FPS counter) don't show up anymore. So I switched it back to the other index. But then I sometimes get flickering.

I don't understand the problem to be honest.

Code/Engine/RendererCore/Debug/Implementation/DebugRenderer.cpp
C-Core: Ok, doesn't hurt to add a StableSort function to the containers, but we should document that it is not really fast.

Code/Engine/RendererCore/Debug/Implementation/DebugRenderer.cpp
C-Core: IndexForRendering should be correct, but I can take a look later

jankrassnigg: I've added a unit test. Not sure whether it makes sense though.

C-Core: Is it possible that you tried to enable the fps counter in editor? That was not working because it only searched for a main view but not an editor view as fallback, I submitted a fix. Now everything seems to work as expected as far as I can tell.

jankrassnigg: No, that's not the problem, but interesting that you made that work.

The flickering is still there (open a mesh asset and transform something, then you'll see the text flicker). I guess I know what the problem is, but it might not be (easily) fixable. For now it's not that important and we can talk about that at a later time.

PR #475, opened 2021-09-15
Adds image volume to procecural generation

jankrassnigg:


Code/EnginePlugins/ProcGenPlugin/Tasks/Implementation/Utils.cpp
```diff
@@ -35,14 +35,32 @@ void ezProcGenExpressionFunctions::ApplyVolumes(ezExpression::Inputs inputs, ezE
 
   const ezSimdVec4f* pInitialValues = inputs[3].GetPtr();
 
+  const ezSimdVec4f* pImgMode = inputs[5].GetPtr();
```
C-Core, line 38: You can assume that imgMode and refcolor are constants and only read the .x() value, same as we do for the tagsetindex above. If you really want that the mode and refcolor can vary for each point you would need to iterate through the input array like we do for the position here.

C-Core (approved):

Code/EnginePlugins/ProcGenPlugin/Tasks/Implementation/Utils.cpp
jankrassnigg: Ok, I wasn't sure whether the VM might execute multiple pieces of the graph in the same call and thus have data change in between (ie when it switches from one volume to another).

Code/Engine/Texture/Image/Implementation/ImageUtils.cpp
```diff
@@ -1517,6 +1517,48 @@ ezResult ezImageUtils::CreateVolumeTextureFromSingleFile(ezImage& dstImg, const
...
+  return NearestSample(image.GetPixelPointer<ezColor>(), image.GetWidth(), image.GetHeight(), addressMode, uv);
+}
+
+ezColor ezImageUtils::NearestSample(const ezColor* pPixelPointer, ezUInt32 uiWidth, ezUInt32 uiHeight, ezImageAddressMode::Enum addressMode, ezVec2 uv)
+{
+  const ezInt32 w = uiWidth;
+  const ezInt32 h = uiHeight;
+
+  uv = uv.CompMul(ezVec2(static_cast<float>(w), static_cast<float>(h))) - ezVec2(0.5f);
+  const float floorX = ezMath::Round(uv.x);
```
C-Core: Is Round really correct here since we subtract 0.5 pixels already above I think it should be floor like in the bilinearSample

Code/EnginePlugins/ProcGenPlugin/Tasks/Implementation/PlacementTask.cpp
```diff
@@ -69,25 +69,39 @@ void PlacementTask::FindPlacementPoints()
...
+
+        if (!hitSurface->IsBasedOn(pOutput->m_hSurface))
+          continue;
+      }
+    }
+    else if (m_pData->m_pOutput->m_Mode == ezProcPlacementMode::Fixed)
+    {
+      ezSimdVec4f rayStart = (vXY + patternCoords * pOutput->m_fFootprint);
+      rayStart += ezSimdRandom::FloatMinMax(ezSimdVec4i(i), vMinOffset, vMaxOffset, seed);
+      rayStart.SetZ(fZStart);
```
C-Core: should be bounding box center here: fZStart + fZRange * 0.5

Code/Engine/Texture/Image/Implementation/ImageUtils.cpp
jankrassnigg: Round does +0.5 and floor, so I guess I should just remove the -0.5

PR #477, opened 2021-09-23
ProcGen improvements (asset save + color gradient)

jankrassnigg:


Code/EnginePlugins/ProcGenPlugin/Components/Implementation/PlacementTile.cpp
```diff
@@ -133,13 +135,20 @@ ezUInt32 PlacementTile::PlaceObjects(ezWorld& world, ezArrayPtr<const PlacementT
 
     pPrefab->InstantiatePrefab(world, transform, options);
 
-    for (auto pRootObject : rootObjects)
+    // only send the color message, if we actually have a custom color
+    if (objectTransform.m_ObjectColor != ezColor::ZeroColor())
```
C-Core: Since the color is sometimes used for arbitrary data it might not be uncommon to use an "everything zero" color. I would add a flag to objectTransform whether to set color or not.

jankrassnigg: ok

PR #478, opened 2021-09-26
# Scene Layer Support

### LINKED ISSUE #464: [FEATURE] Allow to separate scene into multiple layers
* One file per layer
* Locking/unlocking layers

Sanakan8472:
## User Guide
image

- Layers are separate documents (ezLayerDocument) that can be added to a scene to allow for various parts of a scene to be loaded / unloaded and help organize bigger scenes. Right now, all layers are still baked into the parent scene on export. Engine support for layers will be added later.
- A new layer panel is added that allows the user to add / remove layers, load / unload and toggle their visibility. Layers can be dragged around to change their order and game objects can be dragged onto a layer to move them between layers.
- Only one layer can be modified at any given time. By selecting the layer in the layer panel it becomes active and its content is shown in the scene tree.
- As layers are supposed to be unloadable there can't be any object / component references across layers.
- Multi-selection in the viewport will only allow selecting objects from the active layer. Single selection will automatically swtich the active layer to the one corresponding to the selected object.
- Ctrl+S is changed to save the active layer within the scene.
- Layer names map directly to their file names which are stored in the *[SceneName]_data folder*. To rename a layer, its file has to be renamed.

## Layers Implementation Details

- Before this PR, every document window mapped to exactly one document. The concept of a main / sub-document was added similiar to their asset counterparts a sub-asset is part of another document, i.e. the main document that controls it. ezDocument was extended with functions to query the type of document and the currently active sub-document that is being edited.
- To keep the affect on exisiting UI code to a minimum, not all code was changed to be aware that a document window can map to multipe documents. Instead, the ezCommandHistory, ezDocumentObjectManager, ezObjectMetaData and ezSelectionManager had their data moved into a separate storage class and whenever the active layer of a scene changes the content of the afore mentioned classes of the scene is replaced with the content of the active layer. Thus, code that is not layer ware will always work on the active layer. The side-effect of this is that changes must only be done to a layer when it is active as otherwise certain features would break.
- ezScene2Document::InitializeAfterLoadingAndSaving will swap all the document parts (object manager, selection, undo history) against new instances. Thus, everything that accessed those during load gets the 'real' document content. Afterwards, every access goes against the adapter versions that redirect to the active document. In the ezScene2Document dtor, the swap is reversed to make sure everything deints on the same object instances (removing event subs etc).
- Opening / creating a document allows for the context object to be passed into the ctor, allowing for a layer to be informed what the parent scene is.
- Document is aware now that it and its window can host other documents as clients or that is a client doc hosted inside another doc / window.

## Scene Implementation Details

- Scene has its class changed to ezScene2Document and ezSceneDocument has become a base class. Will be renamed in a later PR to keep the diff to a minimum.
- Added enum to differentiate scene, prefab, layer.
- The scene graph tree is now stackable. When the active layer changes the corresponding widget is brought to the front. This means that all view state is preserved when switching the active layer.
  - The only tricky part is the actual scene tree as the scene's content is constatly swapped out we must not rely on the scene pointer in the model. Thus, ezQtGameObjectAdapter / ezQtGameObjectModel needed extensive changes to not rely on the doc pointer to get access to various parts of the doc (selection, object manager, meta data storage).
- Added BeforeReset / AfterReset events to ezDocumentObjectManager. Used when switching the content of the scene when switching layers. So instead of firing object removed / added for every root object there are just two messages: BeforeReset / AfterReset.
- The scene has a custom ezSceneSelectionContext which allows it to switch layers automatically when selecting an element in the viewport that belongs to a different layer.
- Add support for setting a custom root node in a tree model. This allows for the tree model to be used for e.g. layers which are stored in a separate property on the root node.
- Each scene document type now has a custom ezSceneDocumentSettingsBase derived settings class (ezSceneDocumentSettings, ezPrefabDocumentSettings, ezLayerDocumentSettings, ). A patch changes ezSceneDocumentSettings to ezPrefabDocumentSettings as only prefabs use them until now. EnsureSettingsObjectExist will switch the settings object to the correct type when loading old documents.
- ezQtLayerAdapter implements the custom layer visualization. Supports layer loaded / visible state change with small buttons in the item cell.
- Layer lifetime in editor is coupled to the parent scene. I.e., closing a scene also closes all layers.
- Setting reference to object in other layer displays a status message that this is not allowed.
- RMB -> attach to this will show an error if the target object is in another layer.

## Engine Context Implementation Details

- Layers have runtime contexts to mirror their content.
- A DocumentMetaData object can be sent to the engine process on context creation. This allows for the layer context to be informed which scene it belongs to so that all its content is created in the same ezWorld.
- Differentiating between objects of different layers in the ezWorld is implemented by tagging each ezGameObject with a tag in the form of "Layer_1". Layer guids are remapped to indices to prevent tag regstry polution. Thus, all documents share the same tags.
- Hiding layers is done by adding the appropriate exclude tags to the view.
- Allow a layer to be opened individually by posing as a scene. To do this the engine context needs to be redirected to ezSceneContext instead of ezLayerContext which is done by a custom function "AllocateContext" which can be implemented per engine context and gets passed in the ezDocumentOpenMsgToEngine.
- Pick IDs are unique across the scene and its layers as they all share the same m_uiNextComponentPickingID.
- ezEngineProcessDocumentContext::m_pWorld is no longer a unique ptr to support sharing the world between scene and layers.

## Misc changes and fixes

- ezDragDropInfo now stores the adapter of the target parent object to allow querying under which property the dragged object can be stored.
- Fixed the Qt selection so that it inherits the selection state from the underlying selection manager on creation.
- Added Edit > PasteAtOriginalLocation to place objects at the exact same global position they were copied from.
- Better detection for AMD graphics cards to switch some image compare sets.
- Added **EZ_TEST_STATUS** macro to verify ezStatus success and output its error message on failure.
- Fixed bug when calling a function via reflection that returns a pointer.
- Fixed wrong position dragging objects to other layers which also affected copy&paste in general. This is done now by storing the global transform in the serialized nodes. ezSceneDocument::PasteAtOrignalPosition will restore global position if present.
- Fixed icons in the scene tree not being visible at the start.
- Fixed random crash on shadow pool (shadow pool views can contain dead ezWorld ptrs.)
- TypeScript asset was accidentally marked as having an engine connection.
- Fixed relative item order to remain the same after copy&paste or moving between layers.
- Fixed item position under the parent's children when switchting between editor and engine prefabs types.
- Added support to unmap actions from an action mapping. Used to e.g. remove the standard save action again in the scene and replace it with the save active layer action.
- Shadow pool views inherit the exclusion tags of the view that spawned them. This is e.g. used to hide the shadows of hidden layers in the scene.
- Added helper function ezGraphPatchContext::RenameClass for easier patch writing when renaming classes.
- Physics scene now uses layers.

## Unit tests

- ezEditorTest class was extended with various helper functions to test various drag&drop operations.
- Expanded GameObjectReferences test to also test the functionality within a layer.
- Added scene layer operations test.
- Added basic prefab test.


Code/Editor/EditorEngineProcessFramework/EngineProcess/EngineProcessDocumentContext.h
```diff
@@ -100,7 +121,7 @@ class EZ_EDITORENGINEPROCESSFRAMEWORK_DLL ezEngineProcessDocumentContext : publi
   /// \brief Called before a thumbnail context is destroyed. Used for cleanup of what was done in OnThumbnailViewContextCreated()
   virtual void OnDestroyThumbnailViewContext();
 
-  ezUniquePtr<ezWorld> m_pWorld;
+  ezWorld* m_pWorld;
```
jankrassnigg: = nullptr

Code/Editor/EditorEngineProcessFramework/EngineProcess/EngineProcessDocumentContext.h
```diff
@@ -15,6 +15,24 @@ class ezExportDocumentMsgToEngine;
...
 
+struct ezEngineProcessDocumentContextFlags
+{
+  typedef ezUInt8 StorageType;
+
+  enum Enum
+  {
+    None = 0,
+    CreateWorld = EZ_BIT(0),
+    Default = CreateWorld
```
jankrassnigg: Maybe I would make None the default, to make this more explicit everywhere ?

Code/Editor/EditorEngineProcessFramework/EngineProcess/EngineProcessDocumentContext.h
```diff
@@ -24,17 +42,17 @@ class EZ_EDITORENGINEPROCESSFRAMEWORK_DLL ezEngineProcessDocumentContext : publi
   EZ_ADD_DYNAMIC_REFLECTION(ezEngineProcessDocumentContext, ezReflectedClass);
 
 public:
-  ezEngineProcessDocumentContext();
+  ezEngineProcessDocumentContext(ezBitflags<ezEngineProcessDocumentContextFlags> flags = ezEngineProcessDocumentContextFlags::Default);
   virtual ~ezEngineProcessDocumentContext();
 
-  void Initialize(const ezUuid& DocumentGuid, ezEngineProcessCommunicationChannel* pIPC);
+  virtual void Initialize(const ezUuid& DocumentGuid, const ezVariant& metaData, ezEngineProcessCommunicationChannel* pIPC);
```
jankrassnigg, line 48: make sure this isn't called from constructors, because you know how they like virtual functions

Code/Editor/EditorEngineProcessFramework/EngineProcess/EngineProcessDocumentContext.h
Sanakan8472: Fixed.

Code/Editor/EditorEngineProcessFramework/EngineProcess/EngineProcessDocumentContext.h
Sanakan8472: Not possible as these are created via reflection and the void ctor. Should be safe.

Code/Editor/EditorEngineProcessFramework/EngineProcess/EngineProcessDocumentContext.h
Sanakan8472: Done

PR #480, opened 2021-09-27
Refactored geometry extraction

C-Core:
The main purpose of this refactoring is to make the extraction pass as fast as possible so that the rest can be done asynchronously. The extraction message now only holds a cpu mesh reference and a transform for each object which makes it as fast as it can get for regular meshes. This will allow things like navmesh generation or light baking at runtime in the future.

The light baking now also uses this message to extract the geometry for its tracing, thus it now also includes greybox geometry etc.


Code/Engine/Core/World/Implementation/GameObject.cpp
```diff
@@ -13,7 +13,6 @@ namespace
   {
     ezVariantArray value(ezStaticAllocatorWrapper::GetAllocator());
     value.PushBack(ezStringView("CastShadow"));
-    value.PushBack(ezStringView("AutoColMesh")); // TODO: keep this ?
```
jankrassnigg, line 16: This IS used by the greyboxing geometry to decide whether it should generate a collider or not.

jankrassnigg: Ok, saw the change below .... and who updates the documentation? http://ezengine.net/pages/docs/scenes/greyboxing.html

Code/EnginePlugins/RecastPlugin/NavMeshBuilder/NavMeshBuilder.cpp
```diff
@@ -1,15 +1,18 @@
...
+#include <Foundation/Utilities/GraphicsUtils.h>
 #include <Foundation/Utilities/Progress.h>
 #include <Recast/DetourNavMesh.h>
 #include <Recast/DetourNavMeshBuilder.h>
 #include <Recast/Recast.h>
 #include <RecastPlugin/NavMeshBuilder/NavMeshBuilder.h>
 #include <RecastPlugin/Resources/RecastNavMeshResource.h>
+#include <RendererCore/Meshes/CpuMeshResource.h>
+#include <RendererCore/Meshes/MeshBufferUtils.h>
+#include <RendererCore/Utils/WorldGeoExtractionUtil.h>
```
jankrassnigg, line 15: Did this plugin already have a link dependency on RendererCore ?

jankrassnigg (approved):

Code/EnginePlugins/RecastPlugin/NavMeshBuilder/NavMeshBuilder.cpp
C-Core: Yes, it is already used for the navmesh visualization.

jayrulez: Would it make sense to move Mesh stuff to core since meshes are not just for rendering?

Code/Engine/Core/World/Implementation/GameObject.cpp
C-Core: https://github.com/ezEngine/docs-src/pull/2

PR #492, opened 2021-10-31
Added a component to render custom geometry

jankrassnigg:
This component makes it easier to render custom meshes. The ezDynamicMeshBufferResource is already meant for building geometry that works with standard materials, but you need a custom renderer to use it. This component provides exactly that.

It only uses a single material, but you can let it render only a subset of the mesh, so you could have multiple submeshes in a single resource and use multiple components to render them.

### DIFF
--- a/Code/Engine/RendererCore/Meshes/Implementation/CustomMeshComponent.cpp
+++ b/Code/Engine/RendererCore/Meshes/Implementation/CustomMeshComponent.cpp
@@ -93,6 +93,7 @@ ezDynamicMeshBufferResourceHandle ezCustomMeshComponent::CreateMeshResource(ezGA
   desc.m_uiMaxVertices = uiMaxVertices;
   desc.m_uiMaxPrimitives = uiMaxPrimitives;
   desc.m_IndexType = indexType;
+  desc.m_bColorStream = true;
 
   ezStringBuilder sGuid;
   sGuid.Format("CustomMesh_{}", s_iCustomMeshResources.Increment());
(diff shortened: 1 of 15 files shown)


Code/Engine/RendererCore/Meshes/Implementation/CustomMeshComponent.cpp
```diff
@@ -0,0 +1,379 @@
...
+void ezCustomMeshComponent::OnMsgExtractRenderData(ezMsgExtractRenderData& msg) const
+{
+  if (!m_hDynamicMesh.IsValid() || !m_hMaterial.IsValid())
+    return;
+
+  ezResourceLock<ezDynamicMeshBufferResource> pMesh(m_hDynamicMesh, ezResourceAcquireMode::BlockTillLoaded);
+
+  ezCustomMeshRenderData* pRenderData = ezCreateRenderDataForThisFrame<ezCustomMeshRenderData>(GetOwner());
+  {
+    pRenderData->m_GlobalTransform = GetOwner()->GetGlobalTransform() * pRenderData->m_GlobalTransform;
```
C-Core: pRenderData->m_GlobalTransform is identity at this point so it doesn't make sense to multiply with it.

C-Core (approved):

Code/Engine/RendererCore/RenderContext/Implementation/RenderContext.cpp
```diff
@@ -440,10 +440,9 @@ void ezRenderContext::BindMeshBuffer(const ezMeshBufferResourceHandle& hMeshBuff
 }
 
 void ezRenderContext::BindMeshBuffer(ezGALBufferHandle hVertexBuffer, ezGALBufferHandle hIndexBuffer,
-  const ezVertexDeclarationInfo* pVertexDeclarationInfo, ezGALPrimitiveTopology::Enum topology, ezUInt32 uiPrimitiveCount)
+  const ezVertexDeclarationInfo* pVertexDeclarationInfo, ezGALPrimitiveTopology::Enum topology, ezUInt32 uiPrimitiveCount, ezGALBufferHandle hVertexBuffer2, ezGALBufferHandle hVertexBuffer3, ezGALBufferHandle hVertexBuffer4)
```
C-Core, line 443: This certainly needs some cleanup (not necessarily in this PR)

Code/Engine/RendererFoundation/RendererFoundationDLL.h
```diff
@@ -81,8 +81,9 @@ struct EZ_RENDERERFOUNDATION_DLL ezGALIndexType
 {
   enum Enum
   {
-    UShort,
-    UInt,
+    None,   // indices are not used, vertices are just used in order to form primitives
```
C-Core, line 84: Why is the 'None' option needed here? Typically we just use an invalid index buffer if we don't want indexed rendering.

jankrassnigg: Because the enum is used in the creation of the dynamic buffer resource (and thus also the component) and I didn't want to add another enum that would be mostly redundant.
'None' is used to set up meshes that don't use indexed rendering.

Code/Engine/RendererCore/RenderContext/Implementation/RenderContext.cpp
jankrassnigg: Yeah, not the greatest interface, but since it's pretty much only used internally, I thought it's good enough :D

PR #511, opened 2021-11-26
Reflection Probes

### LINKED ISSUE #273: Skylight component far plane should be infinite
Because 100.

Sanakan8472:
Adds support for sphere and box reflection probes and extended the existing fallback probe (sky light).
image

Rendered probe textures have an alpha channel of zero if nothing was rendered during capture at that pixel. This can be used to trace through multiple cube maps until eventually reaching the fallback probe generated by the sky light component.
Probes are rasterized into the clustered data grid in order of increasing volume. At runtime, the cube maps are sampled in that order until a weight of 1 is accumulated or the fallback probe is reached.
Besides the already existing sky light component, two new probe types have been added:
- Sphere reflection probe (used on the right in the above image)
  - The reflection is projected to infinity.
  - The radius and the parent game object's scale define the area of influence. Ellipsoids are supported, it is rasterized either as a cube or sphere depending on which volume is smaller. This is done as rasterizing an ellipsoid is more costly.
  - A falloff can be set to smoothly blend the influence of the probe with the surrounding probes.
- Box reflection probe (used on the left in the above image)
  - The reflection is projected to a box defined by its extents and the parent game objects scale.
  - The influence volume is a subset of the projection volume, defined by the influence scale and shift properties.
  - A falloff is defined for each face of the box.
- Common properties
  - Both probes have a capture offset, allowing for the capture position to be decoupled from the game object position.
  - All probes now have a near and far plane setting. Anything outside the far range will result in an alpha value of 0 in the cube map, which will cause the next probe to be sampled.
  - Eventually static probes will be written to disk. Right now, static probes are updated dynamically once they get mapped to the atlas or whenever the user modifies the probe settings.
  - Dynamic probes currently are update in a round robin fashion across all probes of all worlds.

### Reflection Pool Changes
- Extracted ezReflectionProbeMapping which manages the mapping of probes into its atlas texture.
  - Probes are ordered by their weight which is currently defined as the solid angle of projection to the camera. If the probe is visible in multiple views the max weight is taken.
  - Probes are mapped or evicted from the atlas if more than 32 (s_uiNumReflectionProbeCubeMaps) exists in a single world according to their weight.
- Extracted ezReflectionProbeUpdater which is responsible for rendering / filtering probes into an atlas slot.
- Extracted ezReflectionPool::Data into a separate file. Handles internal probe management.
- The reflection filter pass now uses a compute shader.
- The specular fallback probe is now cleared if a sky light is deactivated.
- Use black fallback texture for specular atlas if rendering reflection probes. To do this the camera usage hint is passed through the extractor.

### Misc changes
- ezReflectionUtils::EnumerationToString overload for ezBitflags
- reflection support for ezTagSet as an accessor property
- Fix ordering problem when disabling multi-threaded rendering in ezGameApplication::Run_WorldUpdateAndRender
- Added support for cube map array textures as UAVs.
- Expose alpha clear value of the source pipeline pass. Used by reflection render pipeline to clear alpha to 0.
- ReflectionProbeVisualization.ezShader now rotates the the reflection according to the source game object and supports setting the ReflectionProbeIndex.


Data/Tools/ezEditor/Localization/en/ScenePlugin.txt
```diff
@@ -405,4 +405,5 @@ ezClothSheetFlags::FixedEdgeRight;Edge: Right
 ezClothSheetFlags::FixedEdgeBottom;Edge: Bottom
 ezClothSheetFlags::FixedEdgeLeft;Edge: Left
 ezClothSheetFlags::FixedEdgeLeft;Edge: Left
-ezCustomMeshComponent;Custom Mesh Component
\ No newline at end of file
+ezBoxReflectionProbeComponent;Box Reflection Probe
+ezSphereReflectionProbeComponent;Sphere Reflection ProbeezCustomMeshComponent;Custom Mesh Component
```
jankrassnigg: something got mixed up here

jankrassnigg: btw, you can insert a link to your new component documentation here

Sanakan8472: Done!

Code/Engine/RendererCore/Lights/ClusteredDataExtractor.h
```diff
@@ -61,8 +65,10 @@ class EZ_RENDERERCORE_DLL ezClusteredDataExtractor : public ezExtractor
 
   ezDynamicArray<ezPerLightData, ezAlignedAllocatorWrapper> m_TempLightData;
   ezDynamicArray<ezPerDecalData, ezAlignedAllocatorWrapper> m_TempDecalData;
+  ezDynamicArray<ezPerReflectionProbeData, ezAlignedAllocatorWrapper> m_TempReflectionProbeData;
   ezDynamicArray<TempCluster<ezClusteredDataCPU::MAX_LIGHT_DATA>> m_TempLightsClusters;
   ezDynamicArray<TempCluster<ezClusteredDataCPU::MAX_DECAL_DATA>> m_TempDecalsClusters;
+  ezDynamicArray<TempCluster<ezClusteredDataCPU::MAX_DECAL_DATA>> m_TempReflectionProbeClusters;
```
C-Core: Should be MAX_REFLECTION_PROBE_DATA

Code/Engine/RendererCore/Lights/Implementation/ClusteredDataExtractor.cpp
```diff
@@ -306,6 +308,78 @@ void ezClusteredDataExtractor::PostSortAndBatch(
...
+          else
+          {
+            ezTransform transform = pReflectionProbeRenderData->m_GlobalTransform;
+            transform.m_vScale = vFullScale.CompMul(probeData.InfluenceScale.GetAsVec3());
+            transform.m_vPosition += transform.m_qRotation * vFullScale.CompMul(probeData.InfluenceShift.GetAsVec3());
+
+            //const ezBoundingBox aabb(ezVec3(-1.0f), ezVec3(1.0f));
+            //ezDebugRenderer::DrawLineBox(view.GetHandle(), aabb, ezColor::DarkBlue, transform);
+
+            RasterizeDecal(transform, uiProbeIndex, viewProjectionMatrix, m_TempReflectionProbeClusters.GetData(), m_ClusterBoundingSpheres.GetData());
```
C-Core: Rename to RasterizeBox and RasterizeSphere

C-Core (approved):

Code/Engine/RendererCore/Lights/ClusteredDataExtractor.h
Sanakan8472: Fixed!

Code/Engine/RendererCore/Lights/Implementation/ClusteredDataExtractor.cpp
Sanakan8472: Done!

PR #516, opened 2021-12-11
Various bug fixes

### LINKED ISSUE #218: 'Global' ezActions don't work as expected
1. They depend on the 'Settings' tab to work (not assert), because that 'fixes' that when the same action is mapped twice, the ezDocument that they are bound to is null. That happens because the Settings tab is always the first document being set up, and its ezDocument pointer is null.
1. Mapping them twice does not work. So for example having the 'Transform All' button in two places would not work.
1. Unless the window in which the action is mapped has focus, any shortcut for global actions does not work.
1. Closing the only window that mapped the global action, will remove the action, making it inaccessible via shortcut. Arguable whether this is wanted, but for a 'global' action it may be nice to not have it tied to any single UI element.

### LINKED ISSUE #226: Ambiguous Qt Action shortcuts when docking documents in a specific way
Dock two documents like this:

image

Then try to execute a shortcut (e.g. CTRL+E). Qt will complain (see screenshot) and the shortcut does not work.

### LINKED ISSUE #297: [BUG] Editor: deleting objects across hierarchies doesn't work
Select many objects via marquee selection, e.g. in the Physics scene. Try to delete them. Depending on which objects are selected, deletion may do absolutely nothing.

This appears to happen when the selected objects span across many hierarchy levels. If you deselect all objects from other hierarchies, deletion works.

This has nothing to do with marquee selection, but is easier to reproduce this way.

Sanakan8472:
- Adding a camera component and setting it to editor usage hint will now allow you to control the editor's fov settings with it.
- Reworked editor redraw code.
  - ezSyncWithProcessMsgToEngine / ezSyncWithProcessMsgToEditor are now global instead of per document and handled by the ezEngineProcessConnection.
  - The engine will now block until the next ezSyncWithProcessMsgToEngine is received to ensure lock stepping between the two processes. To override this, ezEngineProcessDocumentContext::PendingOperationInProgress must be implemented.
  - ezQtUiServices::TickEvent is the global editor tick that fires at the main window's refresh rate. Each ezQtDocumentWindow listens to this event and decides whether to request a redraw or not.
- Added lots of profiling scopes for general engine steps
- Fixed race condition when multiple threads went through a !GetResource => CreateResource code path. CreateResource is replaced with GetOrCreateResource in those cases so if another thread already created the resource between the Get and Create calls the GetOrCreateResource  will not assert but just give out the resource instead.
- Reflection code now uses ezDynamicArray<ezVariant> instead of ezHybridArray<ezVariant, 16> since any ezHybridArray can now be downcast into a ezDynamicArray.
- Added ezReflectionUtils::ClampValue function which applies a ezClampValueAttribute if it can.
- GPUResourcePool now stores a last used frame index so that garbage collection only collects resources of a certain age. This prevents memory thrasing if a GC is forced due to memory pressure.
- Reduced time of ezClusteredDataExtractor::FillItemListAndClusterData to around 60% of the previous implementation.
- Sprinkled some QMenu::setToolTipsVisible(true) in the editor code base.
- Added property copy and paste support to the property grid.
  - Paste works as long as the copied value can be converted to the target property and it can be successfully clamped to the target property's contraints.
- Undo / Redo shortcuts work now while a spin box has focus.
- Added helper function ezObjectAccessorBase::Clear to clear a container property.
- Fixed merge problem in ScenePlugin.txt
- Fixes #297 : Deleting objects across hierarchy levels fails. Caused by using GetSelection instead of GetTopLevelSelection.
- Fixes #226 : Ambiguous Qt Action shortcuts when docking two documents besides each other in the same container window.
  - Document actions are not mapped as ShortcutContext::WindowShortcut because docking allows for multiple documents to be mapped into the same window. Instead, ShortcutContext::WidgetWithChildrenShortcut is used to prevent ambiguous action shortcuts and the actions are executed manually via filtering QEvent::ShortcutOverride at the dock widget level.
- Fixes #218 : 'Global' ezActions are now accessible from anywhere as long as there is a single GUI element that maps the action.
- Fixed focus issues after calling various actions. Now the previous focus is restored after an action has executed.


Code/Engine/RendererCore/Lights/Implementation/ClusteredDataExtractor.cpp
```diff
@@ -409,11 +415,16 @@ void ezClusteredDataExtractor::FillItemListAndClusterData(ezClusteredDataCPU* pD
...
   const ezUInt32 uiMaxReflectionProbeBlockIndex = (uiNumReflectionProbes + 31) / 32;
 
+  const ezUInt32 uiWorstCase = ezMath::Max(uiNumLights, uiNumDecals, uiNumReflectionProbes);
   for (ezUInt32 i = 0; i < NUM_CLUSTERS; ++i)
   {
-    ezUInt32 uiOffset = m_TempClusterItemList.GetCount();
+    const ezUInt32 uiOffset = m_TempClusterItemList.GetCount();
     ezUInt32 uiLightCount = 0;
 
+    // We expand m_TempClusterItemList by the worst case this loop can produce and then cut it down again to the actual size once we have filled the data. This makes sure we do not waste time on boundary checks or out of line calls compared to PushBack or PushBackUnchecked.
```
C-Core: Nice! The comment is a bit confusing though, especially the last part. Maybe: "... or out of line calls **like** PushBack ..."

C-Core (approved):

Code/Engine/RendererCore/Lights/Implementation/ClusteredDataExtractor.cpp
Sanakan8472: Thanks, fixed!

jankrassnigg (approved):

jankrassnigg:
Awesome improvements!

Issue #208, opened 2020-04-29
Input: Mouse Double Click detection does not work


jankrassnigg:
ezInputSlot_MouseDblClick0-2 exist, but currently rely on the OS to provide the information. At least on Windows, but probably other OSes as well, this information is not available, unless specifically asked for. See https://docs.microsoft.com/en-us/windows/win32/inputdev/wm-lbuttondblclk?redirectedfrom=MSDN -> CS_DBLCLKS window flag. This is not desirable, as it hides single clicks in quick succession.

Instead, it would be best to have a ezTime values for when the last click (left, middle, right) arrived (using ezTime::Now()) and check whether a new click is within a short duration. If so, we can inject ezInputSlot_MouseDblClick. This way code that is interested in double clicks can get this information, and code that does not care can just read the single click information. No single click would be removed.

This would then also work the same way on all platforms. One could theoretically even read the double click duration from the OS (meaning how many milliseconds of delay should still be considered a double click).

This check should probably be implemented in ezInputDeviceMouseKeyboard, which should override UpdateInputSlotValues(). Make sure that derived classes also call the base class implementation.

PR #527, opened 2021-12-29
Improved mouse click handling

jankrassnigg:
* Fixes issues at low framerates
* Fixes issues on devices with weird behavior (some trackpads)


jankrassnigg:
Related to #208

PR #528, opened 2021-12-29
Visualizers are now visible in ortho mode

### LINKED ISSUE #517: [FEATURE] It would be nice to be able to see Skylight directional arrow represented in all the viewports.
It would be nice to be able to see Skylight directional arrow represented in all the viewports. So you can see the direction the Skylight is pointing from additional directions.

SkyLightDirectionalArrow

jankrassnigg:
Fixes #517


antonyjones67:
Thank you for adding this! Yay!

PR #530, opened 2021-12-30
Created a MaterialTest project to present EZ materials.

antonyjones67:
image


jankrassnigg:
Hey Antony, I just took a look. The model looks really great, I'd love to replace our material preview sphere with it (or have an option to switch between multiple models, actually).

The main problem is, that GitHub has very nebulous rules about repository size. They don't mention anywhere how much data you can upload. That's why we are very careful to keep the main repository small, and we put most data that's not critical into separate repositories, which we can switch out, if size ever becomes a problem.

Therefore I wouldn't take this PR as is, because it adds a lot of data. I'd want to compress the textures quite a bit more and put the data into the Content sub-module.
Also the OBJ/MTL files contain absolute paths. These days I prefer GLTF (or probably GLB) files.
I don't know about the textures, regarding the license, yet. Will have to take think about that a bit more. If you would find an equally good texture on https://polyhaven.com or https://www.cgbookcase.com I would feel a bit better, since their licenses are definitely fine for us.

I can take care of all that, I don't know exactly where everything would end up myself yet. But I would probably do that next week, since I only have my laptop at hand at the moment and working with Blender is not so great on that.

But for preview purposes only, feel free to update the PR. The screenshot looks fabulous, really awesome work on that model!

antonyjones67:
These are all good suggestions. I will do all the things you recommended. Let's go ahead and stop this pull request for now. Maybe I can send you stuff in a zip file on Dropbox and you can pick and choose what you want. Thank you.

PR #533, opened 2021-12-30
Implemented InsertRange on arrays

jayrulez:


Code/Engine/Foundation/Memory/Implementation/MemoryUtils_inl.h
```diff
@@ -590,6 +596,42 @@ inline void ezMemoryUtils::Prepend(T* pDestination, T&& source, size_t uiCount,
...
+EZ_ALWAYS_INLINE void ezMemoryUtils::Prepend(T* pDestination, const T* source, size_t uiSourceCount, size_t uiCount, ezTypeIsPod)
+{
+  memmove(pDestination + uiSourceCount, pDestination, uiCount * sizeof(T));
+  CopyConstructArray(pDestination, source, uiSourceCount, ezTypeIsPod());
+}
+
+template <typename T>
+EZ_ALWAYS_INLINE void ezMemoryUtils::Prepend(T* pDestination, const T* source, size_t uiSourceCount, size_t uiCount, ezTypeIsMemRelocatable)
+{
+  memmove(pDestination + 1, pDestination, uiCount * sizeof(T));
```
jankrassnigg: The +1 should be +uiSourceCount

Code/Engine/Foundation/Memory/Implementation/MemoryUtils_inl.h
```diff
@@ -590,6 +596,42 @@ inline void ezMemoryUtils::Prepend(T* pDestination, T&& source, size_t uiCount,
...
+{
+  EZ_CHECK_CLASS(T);
+
+  if (uiCount > 0)
+  {
+    MoveConstruct(pDestination + uiCount, std::move(pDestination[uiCount - uiSourceCount]), uiCount);
+
+    for (size_t i = uiCount - 1; i-- > 0;)
+    {
+      pDestination[i + 1] = std::move(pDestination[i]);
```
jankrassnigg: this shifts each element only by one(from i to i+1), but it should shift it by uiSourceCount

Code/Engine/Foundation/Memory/Implementation/MemoryUtils_inl.h
jayrulez: Thanks. Missed that.

Code/Engine/Foundation/Memory/Implementation/MemoryUtils_inl.h
jayrulez: Fixed

Code/Engine/Foundation/Memory/Implementation/MemoryUtils_inl.h
```diff
@@ -590,6 +596,42 @@ inline void ezMemoryUtils::Prepend(T* pDestination, T&& source, size_t uiCount,
...
+}
+
+template <typename T>
+inline void ezMemoryUtils::Prepend(T* pDestination, const T* source, size_t uiSourceCount, size_t uiCount, ezTypeIsClass)
+{
+  EZ_CHECK_CLASS(T);
+
+  if (uiCount > 0)
+  {
+    MoveConstruct(pDestination + uiCount, std::move(pDestination[uiCount - uiSourceCount]), uiCount);
```
jankrassnigg: This line exists in the original source as well, and it doesn't make sense to me. @C-Core or @Ingrater any idea why that exists there? I think it can just be removed.

jankrassnigg: Hmm, if I understand this correctly, this CREATES the N new elements at the end, so that we can then MOVE the elements over them (you can only move after an element has been created already).

In that case the third argument to MoveConstruct would need to be uiSourceCount (your N here).

Frankly, I would feel a bit better, if we had a unit test to cover this code path as well.

jankrassnigg: Also I think you can't just use MoveConstruct here, because uiSourceCount can be larger than uiCount and thus you may not have enough elements to move from. 
It would be easier to use CopyConstruct(dst+count, dst[count-1], srcCount)

That's not as efficient, but works by just replicating a single element across the target array.

jankrassnigg (commented):
Overall it looks good to me, except for the class case. A test for the non-POD case seems necessary to be certain. Could just use ezStrings for that (and otherwise the same logic as the int tests).

Code/Engine/Foundation/Memory/Implementation/MemoryUtils_inl.h
jayrulez: Fixed and added test.

I did come across the issue you highlight as I didn't read your latest feedback yet.

I was able to solve it by first doing move construct - move constructing the elements that should be shifted,
Then doing CopyConstruct for the elements from the source in the slots to be filled. 

I don't know if this is more elegant or efficient but at least it solves the problem.

jankrassnigg:
Awesome, thanks a lot! :-)

PR #538, opened 2022-01-02
Added a mutex to skeleton caching

jankrassnigg:
There is a race condition in the ozz skeleton caching in ezSkeleton::GetOzzSkeleton(). Using a global mutex, all skeleton caching is now serialized.


jankrassnigg:
Thanks to @antonyjones67 for finding the bug.

antonyjones67:
I just verified the fix. Thank you!

PR #552, opened 2022-01-12
Experimental new gizmo meshes

jankrassnigg:
* Under Editor>Preferences you can now enable experimental new gizmos.
  * This will use different gizmo meshes, that are less clunky
  * Meshes are still work in progress
* Gizmos can now be easier picked even with the thinner meshes
* Cleaned up the ezGeometry interface a bit
* Improved the rotation behavior of the drag-to-position gizmo, it now keeps the original rotation as closely as possible


Code/Engine/Core/Graphics/Geometry.h
```diff
@@ -145,29 +164,26 @@ class EZ_CORE_DLL ezGeometry
...
-  void AddTesselatedRectXY(const ezVec2& size, const ezColor& color, ezUInt32 uiTesselationX, ezUInt32 uiTesselationY, const ezMat4& mTransform = ezMat4::IdentityMatrix(), ezUInt16 boneIndex = 0);
+  /// Optionally tessellates the rectangle for more detail.
+  void AddRectXY(const ezVec2& size, ezUInt32 uiTesselationX = 1, ezUInt32 uiTesselationY = 1, const GeoOptions& options = GeoOptions());
 
-  /// \brief Adds an untextured box (8 vertices).
-  void AddBox(const ezVec3& size, const ezColor& color, const ezMat4& mTransform = ezMat4::IdentityMatrix(), ezUInt16 boneIndex = 0);
+  /// \brief Adds a box.
+  /// If bExtraVerticesForTexturing is false, 8 shared vertices are added.
+  /// If bExtraVerticesForTexturing is true, 24 separate vertices with UV coordinates are added.
+  void AddBox(const ezVec3& size, bool bExtraVerticesForTexturing, const GeoOptions& options = GeoOptions());
```
C-Core, line 173: I found AddBox and AddTexturedBox easier to understand than AddBox(..., false) and AddBox(..., true)

jankrassnigg: Yeah, interfaces are hard. I've merged them together, because this way it is easier to discover that there even are different options. I might do the same with cylinder/solid-cylinder.

PR #553, opened 2022-01-14
Ragdoll improvements

jankrassnigg:
* Improved skeleton collider setup
* Added joint limit visualization
* Fixed physics shape setup across skeleton component, collider component and ragdoll component
* Improved the ragdoll demo scene
* Other minor fixes

Part of #418


Code/Editor/EditorFramework/Assets/Implementation/AssetDocument.cpp
```diff
@@ -115,7 +116,7 @@ void ezAssetDocument::InternalAfterSaveDocument()
   {
     // If we request an engine connection but the mirror is not set up yet we are still
     // creating the document and TransformAsset will most likely fail.
-    if (m_EngineConnectionType == ezAssetDocEngineConnection::None || m_Mirror.GetIPC())
+    if (m_EngineConnectionType != ezAssetDocEngineConnection::FullObjectMirroring || m_Mirror.GetIPC())
```
jankrassnigg: @Sanakan8472 I changed this, so that transform on save would work for skeleton assets (it uses a "simple" engine connection). However, then the unit tests failed, because they save all assets, thus now want to transform skeleton and particle assets and then complained that they can't generate thumbnails because there is no engine connection.
I then disabled thumbnail generation during tests, because I don't see a better fix.

Any idea whether that is the right thing to do?

Code/Editor/EditorFramework/Assets/Implementation/AssetDocument.cpp
```diff
@@ -617,6 +621,11 @@ ezStatus ezAssetDocument::RemoteCreateThumbnail(const ThumbnailInfo& thumbnailIn
 
   if (GetEngineStatus() == ezAssetDocument::EngineStatus::Disconnected)
   {
+    if (ezQtEditorApp::GetSingleton()->IsInUnitTestMode())
```
jankrassnigg: Claim thumbnail generation was successful, otherwise saving some assets will always fail.

Code/Editor/EditorFramework/Assets/Implementation/AssetDocument.cpp
Sanakan8472: The line should be
```
if (m_EngineConnectionType == ezAssetDocEngineConnection::None || m_pEngineConnection)
```
Either we don't need an engine connection (ezAssetDocEngineConnection::None) or it is present (m_pEngineConnection). Otherwise the inevitable access to the engine connection during transform / thumbnail generation will fail.

The problem is that during document creation this check will always fail because the engine connection is generated in `InitializeAfterLoadingAndSaving` which obviously will be right after this call.
If you fix the line above, it will obviously still not transform on create document but it will work for all saves after that so you can remove the unit test check again.

PR #554, opened 2022-01-16
XR Stereo rendering improvements

Sanakan8472:
- Dummy XR system added. If no XR plugin is loaded, this dummy system activates via the default game state. It renders both eyes at half 720p resolution next to each other into the window. It's settings can be modified via the main camera component. Can be used to debug stereo rendering in shaders.
- Stereo support for all pipeline, material, sprite and particle shaders for both default and hololens2 render pipeline.
  - Added support for VPAndRTArrayIndexFromAnyShaderFeedingRasterizer in the device capabilities (d3d 11.3+). This is used to inject the permutation variable VERTEX_SHADER_RENDER_TARGET_ARRAY_INDEX into shaders. If available, SV_RenderTargetArrayIndex will be set in the vertex shader for single pass stereo rendering. Otherwise a geometry shader needs to be added to do the same at an increased performance cost.
  - Extracted the material shader in / out struct into a template file so it doesn't need to be duplicated.
  - Stereo support for visual shaders.
  - Added CopyCustomInterpolators function which needs to be defined in the [GEOMETRYSHADER] section if CUSTOM_INTERPOLATOR is used to allow the geometry shader to copy the value between stages.
  - Deduplicated fullscreen shader interpolator struct (FullscreenTriangleInterpolator.h).
  - Fixed LSAO render pass to work again.
- Stereo support for particles and kraut in the simplified shading model.
- ezGraphicsUtils::ExtractPerspectiveMatrixFieldOfView added. Inverse of ezGraphicsUtils::CreatePerspectiveProjectionMatrix using left, right, bottom, top values for the frustum.
  - This is used to fix clustered shading when using custom projection matrices.
  - Extended existing unit tests to test all input permutations.
- Stereo test and effects stress test added (the later one can also be used to profile clustered shading performance).
- Fixed a bug in the DX11 binding that prevented the rendertarget of the last render pass to be bound as input to a compute shader.


jakrams (approved):

Code/Engine/RendererCore/Lights/Implementation/ClusteredDataExtractor.cpp
```diff
@@ -59,11 +71,11 @@ namespace
             }
 
             ezVec3 cc[8];
-            GetClusterCornerPoints(*pCamera, fZf, fZn, fTanFovX, fTanFovY, x, y, z, cc);
+            GetClusterCornerPoints(*pCamera, fZf, fZn, fTanLeft, fTanRight, fTanBottom, fTanTop, x, y, z, cc);
 
             float lightCount = (float)GET_LIGHT_INDEX(clusterData.counts);
             float decalCount = (float)GET_DECAL_INDEX(clusterData.counts);
-            float r = ezMath::Clamp(lightCount / 16.0f, 0.0f, 1.0f);
```
C-Core: Why this change?

Code/Engine/RendererCore/Pipeline/Implementation/Passes/AOPass.cpp
```diff
@@ -219,7 +221,7 @@ void ezAOPass::Execute(const ezRenderViewContext& renderViewContext, const ezArr
 
       renderViewContext.m_pRenderContext->BindNullMeshBuffer(ezGALPrimitiveTopology::Triangles, 1);
 
-      renderViewContext.m_pRenderContext->DrawMeshBuffer().IgnoreResult();
+      renderViewContext.m_pRenderContext->DrawMeshBuffer(1, 0, uiRenderedInstances).IgnoreResult();
```
C-Core: Would be nice if the render context can just do that internally. Maybe we could pass a stereo flag to BeginRendering or so.

Data/Base/Shaders/Common/Lighting.h
```diff
@@ -68,20 +68,20 @@ float MicroShadow(float occlusion, float3 normal, float3 lightDir)
 
 float3 SampleSceneColor(float2 screenPosition)
 {
-  float3 sceneColor = SceneColor.SampleLevel(SceneColorSampler, screenPosition.xy * ViewportSize.zw, 0.0f).rgb;
+  float3 sceneColor = SceneColor.SampleLevel(SceneColorSampler, float3(screenPosition.xy * ViewportSize.zw, s_ActiveCameraEyeIndex), 0.0f).rgb;
```
C-Core, line 71: where does s_ActiveCameraEyeIndex come from`? I can't find the definition.

Data/Base/Shaders/Materials/MaterialInterpolatorTemplate.h
```diff
@@ -0,0 +1,64 @@
...
+  float4 Color1 : COLOR1;
+#  endif
+#endif
+
+#if defined(USE_DEBUG_INTERPOLATOR)
+  float4 DebugInterpolator : DEBUG_INTERPOLATOR;
+#endif
+
+#if defined(CUSTOM_INTERPOLATOR)
+#  ifndef CUSTOM_INTERPOLATOR_NAME
```
C-Core: This does not really work in all cases. Sometimes you need more than one custom interpolator, so previously you could do something like 
#define CUSTOM_INTERPOLATOR float3 a : BLUBB; float3 b : BLABB;
I would suggest that we have a custom function that copies the data in the geometry shader:
CopyCustomInterpolators(GS_OUT output, VS_OUT input) and the user has to implement that if CUSTOM_INTERPOLATOR is defined. We could add a macro that implements the function for different numbers of interpolators used:
CUSTOM_INTERPOLATOR_COPY1(name1)
CUSTOM_INTERPOLATOR_COPY2(name1, name2)
etc.

C-Core (changes requested):

Data/Base/Shaders/Common/Lighting.h
C-Core: My bad, it existed already.

Code/Engine/RendererCore/Lights/Implementation/ClusteredDataExtractor.cpp
Sanakan8472: Good find! Unintended debug code.

Code/Engine/RendererCore/Pipeline/Implementation/Passes/AOPass.cpp
Sanakan8472: I'll add a stereo flag, you will still need to pass in `renderViewContext.m_pCamera->IsStereoscopic()` in most cases but at least you won't need to fiddle with the DrawMeshBuffer params.

C-Core: Much better now, I like it!

Data/Base/Shaders/Materials/MaterialInterpolatorTemplate.h
Sanakan8472: Good point! I've replaced the CUSTOM_INTERPOLATOR_NAME define with the CopyCustomInterpolators function. I didn't add CUSTOM_INTERPOLATOR_COPY1 etc as that would require an include before the #include <Shaders/Materials/MaterialStereoGeometryShader.h> which we don't have right now so it is easier to just define the function instead.

C-Core: Looks good for now. We can add a macro later if needed.

C-Core (approved):

PR #558, opened 2022-01-20
Added way to create a shader from a template

jankrassnigg:
1. Open a material
2. Set its shader mode to "from file"
3. Click on the shader file "browse..." button
4. Select "Create from template"
A dialog pops up, select a template, choose a file name, select the options, click OK -> new shader is created and applied to the material.


Data/Base/Shaders/Materials/MaterialHelper.h
```diff
@@ -88,36 +88,48 @@ ezMaterialData FillMaterialData()
...
 
-  float3 worldNormal = normalize(GetNormal());
-  #if TWO_SIDED == TRUE && defined(USE_TWO_SIDED_LIGHTING)
-    #if FLIP_WINDING == TRUE
-      matData.worldNormal = G.Input.FrontFace ? -worldNormal : worldNormal;
-    #else
-      matData.worldNormal = G.Input.FrontFace ? worldNormal : -worldNormal;
-    #endif
-  #else
+#if defined(USE_NORMAL)
```
C-Core: Technically this is not correct: USE_NORMAL indicates whether you want to use the **vertex** normal stream and interpolator. You can still have a GetNormal function that procedurally generates normals but doesn't use the vertex normal stream.

C-Core (approved):

PR #568, opened 2022-03-08
Cleanup of high-level rendering interface

Sanakan8472:
- Removed concept of primary swap-chain. If was only used to enable alt+enter on a window which actually crashes so best to re-implement that manually. It also made swap-chain life-time very complicated and ownership unclear.
- Removed the ability for the device to create a swapchain. The user now has to do this manually. A auto-created swapchain causes ownership issues and makes it impossible change it later. All samples have been changed to manually create the swap-chain.
- Present is no longer a public function on the device. Instead, a swap-chain needs to be set on a view which is passed into BeginPipelinePlatform / EndPipelinePlatform by the high-level rendering code. The device implementation will then take care of acquire / present logic in these calls.
- Only a single code location (ezGameApplication::Init_SetupGraphicsDevice) sets the used graphics API, all other code locations were redundant and have been removed.
- -renderer flag is now passed into child proccesses like the editor's asset processor.
- Active renderer can be queried via ezGameApplication::GetActiveRenderer.
- ezGameState now supports resizable windows. if a callback is set, ezWindowOutputTargetGAL will check for window changes, recreates the swapchain and triggers the callback. ezGameState will set configure the view again.
- ezXRInterface now requires a ezWindowOutputTargetGAL in its interface instead of blindly casting the base class to ezWindowOutputTargetGAL.
- Fixed a bug in which the render setup in dx11 was not correct as ezGALCommandEncoderImplDX11::BeginCompute did not reset its cached render setup.
- Vsync is no longer part of the (now removed) present function but instead replaced with ezGALPresentMode, which is part of ezGALSwapChainCreationDescription. Some APIs like Vulkan require the re-creation of the swapchain to change modes. ezWindowOutputTargetGAL now takes care of re-creating the swap-chain if cvar_AppVSync changes.


Code/Engine/RendererDX11/Device/Implementation/DeviceDX11.cpp
```diff
@@ -3,6 +3,8 @@
 #include <Core/System/Window.h>
 #include <Foundation/Basics/Platform/Win/IncludeWindows.h>
 #include <Foundation/Configuration/Startup.h>
+#include <RendererCore/Pipeline/Declarations.h>
```
C-Core: Renderer implementations should only depend on RendererFoundation not on RendererCore

Code/Engine/RendererDX11/Device/Implementation/SwapChainDX11.cpp
```diff
@@ -122,6 +122,15 @@ ezResult ezGALSwapChainDX11::InitPlatform(ezGALDevice* pDevice)
   else
 #endif
   {
+#if EZ_ENABLED(EZ_PLATFORM_WINDOWS_DESKTOP)
```
C-Core: Remove or fix?

Code/Engine/RendererFoundation/Device/Device.h
```diff
@@ -247,8 +241,8 @@ class EZ_RENDERERFOUNDATION_DLL ezGALDevice
 
   // Pipeline & Pass functions
 
-  virtual void BeginPipelinePlatform(const char* szName) = 0;
-  virtual void EndPipelinePlatform() = 0;
+  virtual void BeginPipelinePlatform(const char* szName, ezGALSwapChainHandle hSwapChain) = 0;
```
C-Core: Typically the SomethingPlatform functions already get a pointer to the corresponding GAL object and not the handle anymore.

Sanakan8472: Fixed!

Code/Engine/RendererDX11/Device/Implementation/DeviceDX11.cpp
Sanakan8472: Were not in use, removed.

Code/Engine/RendererDX11/Device/Implementation/SwapChainDX11.cpp
Sanakan8472: Removed.

C-Core (approved):

PR #569, opened 2022-03-11
Vulkan swap chain support

Sanakan8472:
With this PR the Vulkan renderer can do the following things which are tested in the new SwapChain renderer tests:
  - Clear a color and or depth target.
  - Resize a swap chain.
  - Render with or without vSync
  - Create depth target of various formats with potential fallbacks if not supported.

## Basic Vulkan frame loop
  - As the renderer is still very much an immediate renderer, a command buffer is started at BeginFramePlatform and is submitted at EndPipelinePlatform at which point a new command buffer is started immediately. The final command buffer is submitted at EndFramePlatform.
  - All command buffers are daisy-chained sequentially for now via semaphores.
  - Each submit pushes a fence into the current frame data. Once the frame is about to be reused, we wait for those fences.
  - Resources that are still in use in a frame need to call ReclaimLater which will put them in a queue on the current frame data. They will be reclaimed once the frame is about to be re-used and we wait for the fences to make sure no resource is still in use.
  - Deletion function the same way, DeleteLater must be called on resources about to be deleted and they will be cleaned up once the fence of the frame is reached.

## Memory allocation
  - We use AMDs VulkanMemoryAllocator which has been added to the third-party folder. There is a thing abstraction layer on top of it to not bleed its header into the rest of the source code and to allow for further abstractions. Only features that are currently in use are exposed in the ezMemoryAllocatorVulkan.

## Helper classes
  - **ezResourceCacheVulkan**: creates persistant Vulkan objects and caches them forever. Currently vk::RenderPass and vk::Framebuffer are implemented.
  - **ezCommandBufferPoolVulkan**: basic pool to re-use command buffers.
  - **ezFencePoolVulkan**: basic pool to re-use fences.
  - **ezSemaphorePoolVulkan**: basic pool to re-use semaphores.
  - **ezConversionUtilsVulkan**: basic functions for EZ -> Vulkan conversion and simple queries.
  - **ezCommandBufferUtilsVulkan**: high level command buffer functions. Currently has a memory barrier that derives many flags from the passed-in resource view.

## Misc changes
  - Added a Resize function on the ezWindow. Currently only implemented on win32.
  - ezGALSwapChainVulkan, ezGALRenderTargetViewVulkan, ezGALResourceViewVulkan, ezGALTextureVulkan have extensive changes but only a sub-set of the functionality is currently tested.


Code/Engine/RendererVulkan/Cache/Implementation/ResourceCacheVulkan.cpp
```diff
@@ -0,0 +1,320 @@
+#include <RendererVulkan/RendererVulkanPCH.h>
+
+#include <Foundation/Basics.h>
+#include <RendererFoundation/Resources/Texture.h>
+#include <RendererVulkan/Cache/ResourceCacheVulkan.h>
+#include <RendererVulkan/Utils/ConversionUtilsVulkan.h>
+
+ezGALDeviceVulkan* ezResourceCacheVulkan::s_pDevice;
+vk::Device ezResourceCacheVulkan::s_device;
+ezHashTable<ezResourceCacheVulkan::RenderPassDesc, vk::RenderPass, ezResourceCacheVulkan::ezRenderPassHash> ezResourceCacheVulkan::s_renderPasses;
```
C-Core, line 10: I think it will be faster to directly use ezGALRenderingSetup as key and then always have a render pass and a framebuffer together in one entry. Or is there any reason why you want to separate those?

Code/Engine/RendererVulkan/CommandEncoder/CommandEncoderImplVulkan.h
```diff
@@ -39,6 +39,8 @@ class EZ_RENDERERVULKAN_DLL ezGALCommandEncoderImplVulkan : public ezGALCommandE
   virtual void InsertFencePlatform(const ezGALFence* pFence) override;
```
C-Core, line 39: Note that we would like to remove these fence functions from the high level interface altogether, so don't bother trying to implement them. If you like you can remove them.

Code/Engine/RendererVulkan/CommandEncoder/Implementation/CommandEncoderImplVulkan.cpp
```diff
@@ -491,23 +516,59 @@ void ezGALCommandEncoderImplVulkan::BeginRendering(vk::CommandBuffer& commandBuf
...
+      colorClear.color.setFloat32({col.r, col.g, col.b, col.a});
+    }
+
+    renderPassBeginInfo.clearValueCount = clearValues.GetCount();
+    renderPassBeginInfo.pClearValues = clearValues.GetData();
+  }
 
   m_pCommandBuffer->beginRenderPass(renderPassBeginInfo, vk::SubpassContents::eInline);
+
+  vk::Viewport viewport;
```
C-Core: Viewport and scissor setup are separate commands, so they should not be set here.

Code/Engine/RendererVulkan/Device/DeviceVulkan.h
```diff
@@ -149,17 +202,36 @@ class EZ_RENDERERVULKAN_DLL ezGALDeviceVulkan : public ezGALDevice
   struct PerFrameData
   {
     ezGALFence* m_pFence = nullptr;
+
+    /// \brief These are all fences passed into submit calls. For some reason waiting for the fence of the last submit is not enough. At least I can't get it to work (neither semaphores nor barriers make it past the validation layer).
+    ezHybridArray<vk::Fence, 2> m_CommandBufferFences;
```
C-Core, line 207: Fences are optional on command submit so it should be enough to insert a fence for the last one in a frame.

Code/Engine/RendererVulkan/Device/Implementation/DeviceVulkan.cpp
```diff
@@ -307,21 +467,109 @@ ezResult ezGALDeviceVulkan::ShutdownPlatform()
...
+  {
+    pSwapChainVulkan->PresentMemoryBarrier(m_PerFrameData[m_uiCurrentPerFrameData].m_currentCommandBuffer);
+  }
+  m_PerFrameData[m_uiCurrentPerFrameData].m_currentCommandBuffer.end();
+
+  vk::SubmitInfo submitInfo = {};
+  submitInfo.commandBufferCount = 1;
+  submitInfo.pCommandBuffers = &m_PerFrameData[m_uiCurrentPerFrameData].m_currentCommandBuffer;
+
+  vk::Fence renderFence = ezFencePoolVulkan::RequestFence();
```
C-Core, line 522: As mentioned in the header only insert a fence for the last command submit

Code/Engine/RendererVulkan/Device/Implementation/DeviceVulkan.cpp
```diff
@@ -307,21 +467,109 @@ ezResult ezGALDeviceVulkan::ShutdownPlatform()
   }
 
   m_pDefaultPass = nullptr;
+
+  ezMemoryAllocatorVulkan::DeInitialize();
```
C-Core, line 468: Since we have this allocator thingy, can it report live resources (buffers, images) on deinit?

Sanakan8472: Allocator and validation layer will do that for you automatically.

Code/Engine/RendererVulkan/Device/DeviceVulkan.h
Sanakan8472: This will only work as long as the command buffers have semaphores to connect them. Once we have e.g. async compute blocks that are independent of the other render data this is no longer the case.

Code/Engine/RendererVulkan/CommandEncoder/CommandEncoderImplVulkan.h
Sanakan8472: Will remove them in the next batch of changes.

Code/Engine/RendererVulkan/Cache/Implementation/ResourceCacheVulkan.cpp
Sanakan8472: ezGALDevice::ReplaceExisitingNativeObject prevents hashing of ezGALRenderingSetup as the handles will point to different things every frame. Will try to remove that functionality in my next batch of changes.

Code/Engine/RendererVulkan/CommandEncoder/Implementation/CommandEncoderImplVulkan.cpp
Sanakan8472: Yep, removed.

PR #570, opened 2022-03-16
MSFT contributions March 2022

jakrams:
* Various improvements to the build system
  * new way of setting up build filters via config files
  * Improved VS2022 support
* Much improved memory reader/writer storage, that can go beyond 32 bit sizes while also being more efficient
* Some improvements to parallel ezWorld loading
* Significantly improved performance of the default array sort implementation, especially in worst-case scenarios
* Minor additions to the profiling system (scope timeout)
* Improved performance of ezRTTI::IsDerivedFrom()
* Small additions to the ezCommandLineOptions
* ezImage can now only load the image header, which can be used to speed up loading texture info only
* Updated Lua 3rd party lib
* Updated zstd 3rd party lib
* Added StackResolver tool


Code/Editor/EditorEngineProcessFramework/EngineProcess/Implementation/WorldRttiConverterContext.cpp
```diff
@@ -199,7 +199,10 @@ void ezWorldRttiConverterContext::UnregisterObject(const ezUuid& guid)
 
 ezRttiConverterObject ezWorldRttiConverterContext::GetObjectByGUID(const ezUuid& guid) const
 {
-  ezRttiConverterObject object = ezRttiConverterContext::GetObjectByGUID(guid);
+  ezRttiConverterObject object = ezRttiConverterContext::GetObjectByGUID(guid);
+
+  if (!guid.IsValid())
+    return object;
```
jankrassnigg, line 205: Not sure why this was necessary. Apparently we get invalid GUIDs here and this could now crash a bit below.

Code/Editor/EditorFramework/LongOps/Implementation/LongOpsAdapter.cpp
```diff
@@ -139,14 +139,15 @@ void ezLongOpsAdapter::ObjectAdded(const ezDocumentObject* pObject)
...
         }
 
         pRtti = pRtti->GetParentType();
       }
     }
+
+    return;
   }
 
   if (pRtti->IsDerivedFrom<ezGameObject>() || pObject->GetParent() == nullptr /*document root object*/)
```
jankrassnigg, line 153: Again, this now crashed. It seems as if calling IsDerivedFrom on a nullptr object actually worked somehow up until now.

Code/Editor/EditorEngineProcessFramework/EngineProcess/Implementation/WorldRttiConverterContext.cpp
jankrassnigg: @Sanakan8472 any idea?

ChristopherManthei: This shouldn't happen unless the serialization was messed up. Do you have a callstack / repro case for this?

ChristopherManthei: Ah nevermind, it's all the derived type changes. Looks fine.

PR #572, opened 2022-03-26
Render pipeline changes

Sanakan8472:
## Renderer changes
* Restored DX11 GPU profiling scopes.
  * ezGALDevice::BeginFrame now has a uiRenderFrame parameter to allow implementations to correlate GPU profile data with CPU profile data.
  * ezProfilingScopeAndMarker extended with Start and Stop method to allow their use outside of scoped objects.
image
* Added **ezGALRenderTargets** class. An array of ezGALTextureHandle. Used by ezView to determine where to render to. Previously ezView used ezGALRenderTargetSetup but the view handles served no purpose except for resolving to the underlying texture handles anyways.
* Removed **ezGALFence**. Fences are now an implementation detail.
* Removed **ReplaceExisitingNativeObject**. With this change, GAL handles and native pointers have a 1:1 relationship. This allows to created caches based on handles instead of their native objects.
  * ezResourceCacheVulkan will now hash handles instead of needing to drill down to native objects.
  * To swap the textures used in a swap-chain based view, ezGALSwapChain::AcquireNextImage is called by ezRenderPipeline and the new render targets are queried via ezGALSwapChain::GetRenderTargets.
  * To easily swap the texture used in render passes, ezRenderPipeline::TextureUsageData now stores an offset into ezGALRenderTargets (slightly unsafe).
  * This change also means that any custom low level render code must not cache handles retrieved from a swap chain as these will change each frame.
* Added **WriteRenderPipelineDgml()** console function which writes a dgml file for each active render pipeline.
  * Added ezRenderPipeline::CreateDgmlGraph. Can be used to inspect flow of render targets and pool textures and render pass order.
image
* Replaced ezGALRenderTargetSetup::GetMaxRenderTargetIndex with ezGALRenderTargetSetup::GetRenderTargetCount to make it behave like any other container. Also removed HasRenderTargets and fixed GetRenderTarget to actually assert if out of bounds accesses ar done if no render targets are set.
* Added ezGALDeviceVulkan::Submit. Allows for the device not not be aware of swap chain implementation details.
* Vulkan: fallback for swap chain modes if not supported by hardware.

## Swap Chain changes
* There can now be many custom swap chain implementations. ezGALSwapChain is now a base class of base classes. Implementations need to implement AcquireNextImage, PresentNextImage and GetRenderTargets.
* Swap chains are created via each base class create function, e.g. ezGALWindowSwapChain::Create. Internally these will forward to ezGALDevice::CreateSwapChain which requires a SwapChainFactoryFunction which ensures GAL objects are created under the correct allocator.
* ezGALWindowSwapChain is the base class for window-based swap chains. ezGALWindowSwapChainCreationDescription is used to create these which replaces ezGALSwapChainCreationDescription.
* ezGALXRSwapChain is the base class for XR swapchains.
* Added a templated ezGALDevice::GetSwapChain function to safely cast a swap chain to one of its base classes.

## View changes
* Replaced SetRenderTargetSetup/GetRenderTargetSetup with SetRenderTargets/GetRenderTargets.
* SetSwapChain and SetRenderTargets are now mutually exclusive. Calling each function will reset the other.
* ezView::GetActiveRenderTargets retrieves the ezGALRenderTargets from either swap chain or SetRenderTargets, depending on what was set.
* Extraction will retrieve both swap chain and ezGALRenderTargets as these can change on every render frame.

## XR changes
* Added custom swap chain base class: ezGALXRSwapChain.
* Removed ezXRInterface::Present, present logic is done by the ezGALXRSwapChain implementations.
* Added ezXRInterface::GetCurrentTexture which is used by the new ezWindowOutputTargetXR::RenderCompanionView method to get the input for the companion view.
* Limit RenderCompanionView to 60 fps as otherwise it might block the HMD rendering at higher refresh rates.
* MainRenderPipeline now sets the depth buffer of the target pass. This allows for XR implementations that use the depth buffer for reprojection to work correctly.
* OpenXR: Do not release the view texture before the companion view has read from it.
* OpenXR: Moved lots of code into the new ezGALOpenXRSwapChain class.

## Misc changes
* Support AMD API to select dedicated GPU in multi-gpu machines.
* DGML support for connection labels.


Code/Engine/RendererCore/Pipeline/Implementation/RenderPipeline.cpp
```diff
@@ -1162,4 +1191,44 @@ ezRenderDataBatchList ezRenderPipeline::GetRenderDataBatchesWithCategory(ezRende
...
+  for (ezUInt32 i = 0; i < m_TextureUsage.GetCount(); ++i)
+  {
+    const TextureUsageData& data = m_TextureUsage[i];
+
+    for (const ezRenderPipelinePassConnection* pCon : data.m_UsedBy)
+    {
+      ezDGMLGraph::NodeDesc nd;
+      nd.m_Color = data.m_iTargetTextureIndex != -1 ? ezColor::Black : ezColor::GetPaletteColor(i, 64);
+      nd.m_Shape = ezDGMLGraph::NodeShape::RoundedRectangle;
+      sTmp.Format("{} #{}: {}x{}:{}, MSAA:{}, {}Format: {}", data.m_iTargetTextureIndex != -1 ? "RenderTarget" : "PoolTexture", i, pCon->m_Desc.m_uiWidth, pCon->m_Desc.m_uiHeight, pCon->m_Desc.m_uiArraySize, (int)pCon->m_Desc.m_SampleCount, ezGALResourceFormat::IsDepthFormat(pCon->m_Desc.m_Format) ? "Depth" : "Color", (int)pCon->m_Desc.m_Format);
```
C-Core: Use reflection to write stringified format

Code/Engine/RendererFoundation/Device/SwapChain.h
```diff
@@ -1,27 +1,55 @@
...
-  /// \brief Returns a handle to the back buffer texture.
-  ///
-  /// \attention Some RenderSystems do not support reading the swap chain's back buffer texture. Those will return an invalid handle.
-  inline ezGALTextureHandle GetBackBufferTexture() const;
+  const ezGALRenderTargets& GetRenderTargets() const { return m_RenderTargets; }
+  ezGALTextureHandle GetBackBufferTexture() const { return m_RenderTargets.m_hRTs[0]; }
 
-protected:
-  ezGALSwapChain(const ezGALSwapChainCreationDescription& Description);
+  virtual void AcquireNextImage(ezGALDevice* pDevice) = 0;
```
C-Core: Change name to something more in line with ez naming, like e.g. AcquireNextRenderTarget(s)

C-Core (approved):

Code/Engine/RendererFoundation/Device/SwapChain.h
Sanakan8472: Changed to AcquireNextRenderTarget / PresentRenderTarget.

Code/Engine/RendererCore/Pipeline/Implementation/RenderPipeline.cpp
Sanakan8472: Fixed and also corrected other instances.

PR #573, opened 2022-03-30
Added sensor components

C-Core:
* Sensor components are typically used for AI perception like vision or hearing. They have a shape like sphere, cylinder or cone and can do visibility tests via raycasts. The components store an array of all their currently visible objects and send an event msg if this array changes. What do to with these objects is then up to higher level AI logic.

image
image

* Also added a new time interval based scheduler which can be used to update generic objects at a lower rate than every frame. The scheduler tries to maintain an even workload each frame and also honor the given interval for each object. This scheduler is used for the sensor components but can be used in other places as well.


Code/Engine/Core/Utils/IntervalScheduler.h
```diff
@@ -0,0 +1,116 @@
...
+  };
+
+  static ezTime GetInterval(Enum updateRate);
+};
+
+EZ_DECLARE_REFLECTABLE_TYPE(EZ_CORE_DLL, ezUpdateRate);
+
+//////////////////////////////////////////////////////////////////////////
+
+class EZ_CORE_DLL ezIntervalSchedulerBase
```
jankrassnigg, line 34: Could use some documentation.

Code/Engine/Core/Utils/IntervalScheduler.h
```diff
@@ -0,0 +1,116 @@
...
+
+//////////////////////////////////////////////////////////////////////////
+
+template <typename T>
+class ezIntervalScheduler : public ezIntervalSchedulerBase
+{
+  using SUPER = ezIntervalSchedulerBase;
+
+public:
+  EZ_ALWAYS_INLINE ezIntervalScheduler(ezTime minInterval = ezTime::Milliseconds(1), ezTime maxInterval = ezTime::Seconds(1))
```
jankrassnigg, line 91: Does it make sense to have default values here? Aren't these values highly context specific?

Code/Engine/Core/Utils/IntervalScheduler.h
```diff
@@ -0,0 +1,116 @@
...
+    : SUPER(minInterval, maxInterval)
+  {
+    EZ_CHECK_AT_COMPILETIME_MSG(sizeof(T) <= sizeof(ezUInt64), "sizeof T must be smaller or equal to 8bytes");
```
jankrassnigg: I think we could get rid of this macro and just use static_assert directly now.

jankrassnigg: Also "8bytes" is a typo?

Code/Engine/Core/Utils/Implementation/IntervalScheduler.cpp
```diff
@@ -0,0 +1,193 @@
...
+ezTime ezUpdateRate::GetInterval(Enum updateRate)
+{
+  return s_Intervals[updateRate];
+}
+
+//////////////////////////////////////////////////////////////////////////
+
+EZ_ALWAYS_INLINE float GetRandomZeroToOne(int pos, ezUInt32& seed)
+{
+  return ezSimdRandom::FloatZeroToOne(ezSimdVec4i(pos), ezSimdVec4u(seed++)).x();
```
jankrassnigg, line 35: What's the position here? Is this 3D noise? ezSimdRandom could also use some code comments...

Code/Engine/Core/World/Declarations.h
```diff
@@ -139,6 +146,13 @@ struct ezHashHelper<ezComponentHandle>
   EZ_ALWAYS_INLINE static bool Equal(ezComponentHandle a, ezComponentHandle b) { return a == b; }
 };
 
+/// \brief Currently not implemented as it is not needed for component handles.
```
jankrassnigg, line 149: = delete ?

Code/Engine/Core/World/Implementation/Declarations.cpp
```diff
@@ -19,6 +28,28 @@ EZ_END_STATIC_REFLECTED_ENUM;
...
+
+void operator<<(ezStreamWriter& Stream, const ezComponentHandle& Value)
+{
+  EZ_ASSERT_DEV(false, "This function should not be called. Use ezWorldWriter::WriteComponentHandle instead.");
+}
+
+void operator>>(ezStreamReader& Stream, ezComponentHandle& Value)
+{
+  EZ_ASSERT_DEV(false, "This function should not be called. Use ezWorldReader::ReadComponentHandle instead.");
+}
```
jankrassnigg, line 49: = delete instead ?

Code/Engine/Core/World/Implementation/World.cpp
```diff
@@ -995,12 +995,13 @@ void ezWorld::UpdateAsynchronous()
...
+    ezWorldModule* pModule = static_cast<ezWorldModule*>(updateFunction.m_Function.GetClassInstance());
+    ezComponentManagerBase* pManager = ezDynamicCast<ezComponentManagerBase*>(pModule);
 
-    const ezUInt32 uiTotalCount = pManager->GetComponentCount();
-    ezUInt32 uiStartIndex = 0;
-    ezUInt32 uiGranularity = (updateFunction.m_uiGranularity != 0) ? updateFunction.m_uiGranularity : uiTotalCount;
+    const ezUInt32 uiTotalCount = pManager != nullptr ? pManager->GetComponentCount() : 1;
+    const ezUInt32 uiGranularity = (updateFunction.m_uiGranularity != 0) ? updateFunction.m_uiGranularity : uiTotalCount;
 
+    ezUInt32 uiStartIndex = 0;
```
jankrassnigg, line 1004: Is this just code cleanup, or is there a functional change?

Code/Engine/Foundation/Reflection/Implementation/ArrayProperty.h
```diff
@@ -207,3 +207,56 @@ class ezMemberArrayProperty : public ezTypedArrayProperty<typename ezTypeTraits<
   GetConstContainerFunc m_ConstGetter;
   GetContainerFunc m_Getter;
 };
+
+template <typename Class, typename Container, typename Type>
+class ezMemberArrayReadOnlyProperty : public ezTypedArrayProperty<typename ezTypeTraits<Type>::NonConstReferenceType>
```
jankrassnigg, line 213: @Sanakan8472 is this property needed?

Also, if yes, it should have some code comment describing what effect it has.

Code/Engine/GameEngine/AI/SensorComponent.h
```diff
@@ -0,0 +1,205 @@
...
+struct EZ_GAMEENGINE_DLL ezMsgSensorVisibleObjectsChanged : public ezEventMessage
+{
+  EZ_DECLARE_MESSAGE_TYPE(ezMsgSensorVisibleObjectsChanged, ezEventMessage);
+
+  ezArrayPtr<ezGameObjectHandle> m_VisibleObjects;
+};
+
+//////////////////////////////////////////////////////////////////////////
+
+class EZ_GAMEENGINE_DLL ezSensorComponent : public ezComponent
```
jankrassnigg, line 25: I know we often don't document components much, but it would still be nice to actually have some basic description here, what this component does and how to use it.

Code/Engine/GameEngine/AI/Implementation/SensorComponent.cpp
```diff
@@ -0,0 +1,683 @@
...
+{
+  EZ_BEGIN_PROPERTIES
+  {
+    EZ_MEMBER_PROPERTY("Radius", m_fRadius)->AddAttributes(new ezDefaultValueAttribute(10.0f), new ezClampValueAttribute(0.0f, ezVariant())),
+  }
+  EZ_END_PROPERTIES;
+
+  EZ_BEGIN_ATTRIBUTES
+  {
+    new ezCategoryAttribute("AI"),
```
jankrassnigg: The category can be set on the base class. Also maybe it should be "AI/Sensors"

jankrassnigg (approved):

Code/Engine/Core/World/Implementation/World.cpp
C-Core: It's a fix, that code crashed if an ezWorldModules (not a component manager) registered an update function for the async phase.

Code/Engine/Core/Utils/IntervalScheduler.h
C-Core: I think the defaults are good enough for typical game use cases

Code/Engine/Core/World/Implementation/Declarations.cpp
C-Core: Nope doesn't work

Code/Engine/Core/World/Declarations.h
C-Core: Nope doesn't work

Code/Engine/Foundation/Reflection/Implementation/ArrayProperty.h
C-Core: It is needed to reflect ezArrayPtr as member

PR #574, opened 2022-03-31

jankrassnigg:


Code/EditorPlugins/PhysX/EditorPluginPhysX/Actions/PhysXActions.cpp
```diff
@@ -26,7 +25,7 @@ void ezPhysXActions::UnregisterActions()
 void ezPhysXActions::MapMenuActions()
 {
   /// \todo Is there a way to integrate into ALL document types in a specific menu (ie. project settings)
-  ezActionMap* pMap = ezActionMapManager::GetActionMap("EditorPluginScene_DocumentMenuBar");
+  ezActionMap* pMap = ezActionMapManager::GetActionMap("EditorPluginScene_Scene2MenuBar");
```
C-Core, line 28: Note this change makes the option visible in scenes but not in prefabs anymore

jankrassnigg: That was the original behavior, actually. Would need a general improvement to how this is done, to have it show up everywhere. (@Sanakan8472)

PR #578, opened 2022-04-16
Vulkan support for shaders, index/vertex buffer and render pipelines

Sanakan8472:
## Index and vertex buffer support
- Added ezGALDeviceVulkan::UploadBuffer to set the intial state of a buffer object.
- Added ezResourceCacheVulkan::RequestPipelineLayout and RequestGraphicsPipeline. Pipeline layout is just a dummy for now, no inputs are supported.
- Added ezStagingBufferPoolVulkan. It doesn't cache anything and used buffers are immediately destroyed but will be extended in the future.

## Shader support
- Some glue code in ezGALShaderVulkan.
- Implemented ezGALVertexDeclarationVulkan.
- Added ezGALBlendStateVulkan accessors.

## Reworked state caching
- ezGALCommandEncoderImplVulkan was extended to do state caching and dirty flag tacking.
- ezRenderContext is reset on pipeline end. We do reset the ezGALCommandEncoderImplVulkan and ezGALCommandEncoder at the same time so that the next command buffer and/or frame does not assume stale caches.
- Whenever a new command buffer is started we only invalidate the ezGALCommandEncoderImplVulkan as the raw pointers stored in it will remain valid until the end of the frame so we can simply flush all states again in the next command buffer.

## DeviceVulkan cleanup
- Some infrastructure to support transfer queue.
- GetCurrentCommandBuffer will lazily start a new command buffer.
- Remove code duplication like submit logic and start new command buffer.

## Clip space
- ezClipSpaceDepthRange::Default and ezClipSpaceYMode::RenderToTextureDefault now start as DX11 defaults until a renderer is initialized and overrides the settings.
- Added CLIP_SPACE_FLIPPED shader permutation that maps to whether ezClipSpaceYMode == flipped.

## Renderer Unit Test
- The pipeline states test tries to use more and more features of the render pipeline.
- ST_MostBasicShader is the least amount of rendering code you need to implement in a renderer to do the first draw call. The shader runs in NDC space and has no input of any kind, nertex postions of the triangle are stores in the shader itself.
- ST_ViewportScissor uses the previous shader to verify viewport and scissor settings.
- ST_VertexBuffer renders the same triangle as ST_MostBasicShader but this time using a vertex buffer.
- ST_IndexBuffer renders a NDC space sphere using index and vertex buffer.


Code/Engine/RendererVulkan/Shader/Implementation/ShaderVulkan.cpp
```diff
@@ -56,93 +56,69 @@ void ezGALShaderVulkan::SetDebugName(const char* szName) const
...
   if (m_Description.HasByteCodeForStage(ezGALShaderStage::VertexShader))
   {
-    if (FAILED(pD3D11Device->CreateVertexShader(m_Description.m_ByteCodes[ezGALShaderStage::VertexShader]->GetByteCode(),
-                                                m_Description.m_ByteCodes[ezGALShaderStage::VertexShader]->GetSize(), nullptr,
-                                                &m_pVertexShader)))
-    {
-      ezLog::Error("Couldn't create native vertex shader from bytecode!");
-      return EZ_FAILURE;
-    }
+    createInfo.codeSize = m_Description.m_ByteCodes[ezGALShaderStage::VertexShader]->GetSize();
```
C-Core, line 64: Make that a loop to reduce code duplication?

Code/Engine/RendererVulkan/Shader/Implementation/VertexDeclarationVulkan.cpp
```diff
@@ -42,49 +25,47 @@ ezResult ezGALVertexDeclarationVulkan::InitPlatform(ezGALDevice* pDevice)
...
   {
-    return EZ_SUCCESS;
+    if ((usedBindings & EZ_BIT(i)) == 0)
+    {
+      m_bindings.RemoveAtAndCopy(i);
+    }
   }
-#endif
 
+  const ezScopedRefPointer<ezGALShaderByteCode>& pByteCode = pShader->GetDescription().m_ByteCodes[ezGALShaderStage::VertexShader];
```
C-Core: Not used?

C-Core (approved):

Code/Engine/RendererVulkan/Shader/Implementation/ShaderVulkan.cpp
Sanakan8472: Will be part of my next PR as I have to rewrite the class anyways.

Code/Engine/RendererVulkan/Shader/Implementation/VertexDeclarationVulkan.cpp
Sanakan8472: Fixed!

PR #579, opened 2022-04-23
Jolt integration improvements

jankrassnigg:
* New cone constraint
* New swing twist constraint
* Implemented contact events (impact, side, roll reactions)
* New query-shape only actor
* Optimized body creation (delayed and batched)
* New trigger component
* New bone collider component for animated meshes
* Jolt mesh resource now cleans up unused memory
* Improved performance of dynamic body updates
* Implemented constraint pair filter (joined bodies can now ignore collisions)
* Added broadphase optimization when many new bodies were added


Code/Editor/EditorFramework/Assets/Implementation/AssetUpdates.cpp
```diff
@@ -632,6 +632,11 @@ void ezAssetCurator::InvalidateAssetTransformState(const ezUuid& assetGuid)
   ezAssetInfo* pAssetInfo = nullptr;
   if (m_KnownAssets.TryGetValue(assetGuid, pAssetInfo))
   {
+    if (pAssetInfo->m_bInvalidating)
```
jankrassnigg: @Sanakan8472 I ran into a recursive crash once, after setting a particular asset dependency. I wasn't able to really fix it, but this part remained from my attempt.

Sanakan8472: This could only happen if you have circular dependencies / references between assets. While this will prevent the stackoverflow it will most likely crash at a different location afterwards as the asset system is not designed to handle cyclic dependencies. It would be useful to figure out how the circular dependency came to be and what the use case is.
We absolutely can't have circles in m_InverseDependency. If there are circles in m_InverseReferences we could probably work around it but it would require a lot of special case handling but I doubt there are any intentional use cases for this.

jankrassnigg: True, as I said, it continue to crash in other places. I think I got a cyclic dependency by having a Surface which referenced a prefab, which referenced a mesh, which referenced a material, which referenced the same surface....

The use case is real, it makes sense to have that kind of dependency chain, however, as we already brainstormed about in the past, it would be useful to have one of these links not as a fully asset dependency, but rather something else (since e.g. changing the surface on a material currently also triggers a full asset reprocessing, although it doesn't have any transitive impact).

Should I remove this change again, or keep it in?

Sanakan8472: Remove for now, I'll think about it a bit.

PR #583, opened 2022-05-01
Vulkan Resource Binding Support

Sanakan8472:
## Resource Binding Support
* ezResourceCacheVulkan::RequestDescriptorSetLayout creates a vk::DescriptorSetLayout backed by a hashtable.
* ezDescriptorSetPoolVulkan::CreateDescriptorSet creates new sets until the current pool is exhaused. Pool will return to the ring buffer once no descriptor sets in it are in use anymore.
* ezGALSamplerStateVulkan / ezGALResourceViewVulkan / ezGALBufferVulkan / ezGALUnorderedAccessViewVulkan now have getters for vk::DescriptorImageInfo / vk::DescriptorBufferInfo used to build the descriptor set.

## Vulkan Shader
* The EZ shader system stores reflection data for each individual shader stage. However, in Vulkan all stages combined share the same descriptor binding slots and any collisions between stages must be resolved to build a valid shader. To do so we store meta data in the shader to remap descriptor binding slots at runtime which allos us to deduplicate shader stage code that is used by mulitple shaders.
* Meta data (ezSpirvMetaData) is written into the shader binary blob. It has all the data necessary to reconstruct the descriptor set layout and remap descriptor bindings.
* No versioning info was present in the old shader stage binary blob so any old Vulkan shader cache needs to be deleted.
* Remapping happens in ezGALShaderVulkan::InitPlatform. As the vertex shader is the first stage it never requires remapping. This will build a remapping table as the shader reflection used by the high level renderer is per stage and assumes it can map resources to stages. A remapping table was added to map our descriptor binding to the original per-stage resource binding model. This happens during ezGALCommandEncoderImplVulkan::FlushDeferredStateChanges.

## Buffer Updates
* Implemented ezGALCommandEncoderImplVulkan::UpdateBufferPlatform.
* Added ezGALBufferVulkan::DiscardBuffer which creates a new buffer and puts the old one in a ring buffer to be reclaimed later once it is no longer in use.

## Vulkan Device
* Upgraded to Vulkan 1.1
  * We now use ezClipSpaceYMode::Regular and rely in the Vulkan 1.1 feature that a nagative height performs y-inversion of the clip-space to framebuffer-space transform.
  * Descriptor set pools return vk::Result::eErrorOutOfPoolMemory if exhaused. Removing the requirement to count usage yourself.
* Added SetDebugName to unify setting Vulkan object debug names.
* Added GetCurrentFrame / GetSafeFrame to allow external code to figure out if a resource is still in use. Anything that was touched <= GetSafeFrame is safe to reuse.
* m_uiFrameCounter starts at 1 now to make sure GetCurrentFrame  / GetSafeFrame are not identical at the start.
* Per frame data m_reclaimResources / m_pendingDeletions is now swapped out at the end of the frame with m_reclaimResourcesPrevious / m_pendingDeletionsPrevious respectively to allow resources to be added to the list outside of the renderframe which would have previously filed the objects under an old fence, reclaiming them too early.
* Some code cleanup.

## Unit Tests
* Added a constant buffer test that updates two constant buffers multiple times per frame.
* Added a structured buffer test that updates a buffer each frame used for an instanced draw call of the same scene generated by the previous test.

### DIFF
--- a/Code/Engine/RendererVulkan/CommandEncoder/Implementation/CommandEncoderImplVulkan.cpp
+++ b/Code/Engine/RendererVulkan/CommandEncoder/Implementation/CommandEncoderImplVulkan.cpp
@@ -700,12 +700,12 @@ void ezGALCommandEncoderImplVulkan::SetRasterizerStatePlatform(const ezGALRaster
 
 void ezGALCommandEncoderImplVulkan::SetViewportPlatform(const ezRectFloat& rect, float fMinDepth, float fMaxDepth)
 {
-  // We use ezClipSpaceYMode::Regular and rely in the Vulkan 1.1 feature that a nagative height performs y-inversion of the clip-space to framebuffer-space transform.
+  // We use ezClipSpaceYMode::Regular and rely in the Vulkan 1.1 feature that a negative height performs y-inversion of the clip-space to framebuffer-space transform.
   // https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VK_KHR_maintenance1.html
   vk::Viewport viewport = {rect.x, rect.height + rect.y, rect.width, -rect.height, fMinDepth, fMaxDepth};
   if (m_viewport != viewport)
   {
-    // viewport is marked as dynamic in the pipeline layout and thus does not mark m_bPipelineStateDirty.
+    // Viewport is marked as dynamic in the pipeline layout and thus does not mark m_bPipelineStateDirty.
     m_viewport = viewport;
     m_bViewportDirty = true;
   }
--- a/Code/Engine/RendererVulkan/CommandEncoder/Implementation/CommandEncoderImplVulkan.cpp
+++ b/Code/Engine/RendererVulkan/CommandEncoder/Implementation/CommandEncoderImplVulkan.cpp
@@ -900,7 +900,7 @@ void ezGALCommandEncoderImplVulkan::FlushDeferredStateChanges()
     m_bIndexBufferDirty = false;
   }
 
-  if (true/*m_bDescriptorsDirty*/)
+  if (true /*m_bDescriptorsDirty*/)
   {
     //#TODO_VULKAN we always create a new descriptor set as we don't know if a buffer was modified since the last draw call (ezGALBufferVulkan::DiscardBuffer).
     // Need to figure out a fast check if any buffer or buffer of a resource view was discarded.
--- a/Code/Engine/RendererVulkan/CommandEncoder/Implementation/CommandEncoderImplVulkan.cpp
+++ b/Code/Engine/RendererVulkan/CommandEncoder/Implementation/CommandEncoderImplVulkan.cpp
@@ -3,6 +3,7 @@
 #include <RendererVulkan/Cache/ResourceCacheVulkan.h>
 #include <RendererVulkan/CommandEncoder/CommandEncoderImplVulkan.h>
 #include <RendererVulkan/Device/DeviceVulkan.h>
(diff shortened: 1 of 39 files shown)


Code/Engine/RendererVulkan/CommandEncoder/Implementation/CommandEncoderImplVulkan.cpp
```diff
@@ -720,7 +700,9 @@ void ezGALCommandEncoderImplVulkan::SetRasterizerStatePlatform(const ezGALRaster
 
 void ezGALCommandEncoderImplVulkan::SetViewportPlatform(const ezRectFloat& rect, float fMinDepth, float fMaxDepth)
 {
-  vk::Viewport viewport = {rect.x, rect.y, rect.width, rect.height, fMinDepth, fMaxDepth};
+  // We use ezClipSpaceYMode::Regular and rely in the Vulkan 1.1 feature that a nagative height performs y-inversion of the clip-space to framebuffer-space transform.
```
C-Core: typo in negative height

Code/Engine/RendererVulkan/Device/Implementation/DeviceVulkan.cpp
```diff
@@ -373,29 +377,9 @@ ezResult ezGALDeviceVulkan::InitPlatform()
...
-  m_Timestamps.SetCountUninitialized(1024);
-  for (ezUInt32 i = 0; i < m_Timestamps.GetCount(); ++i)
-  {
-    //if (FAILED(m_pDevice->CreateQuery(&timerQueryDesc, &m_Timestamps[i])))
-    //{
-    //  ezLog::Error("Creation of native DirectX query for timestamp has failed!");
-    //  return EZ_FAILURE;
-    //}
-  }
+  // We use ezClipSpaceYMode::Regular and rely in the Vulkan 1.1 feature that a nagative height performs y-inversion of the clip-space to framebuffer-space transform.
```
C-Core: typo in negative

Code/Engine/RendererVulkan/Pools/Implementation/DescriptorSetPoolVulkan.cpp
```diff
@@ -0,0 +1,140 @@
...
+}
+
+vk::DescriptorPool ezDescriptorSetPoolVulkan::GetNewPool()
+{
+  if (s_freePools.IsEmpty())
+  {
+    ezHybridArray<vk::DescriptorPoolSize, 20> poolSizes;
+    for (auto weight : s_descriptorWeights)
+    {
+      if (static_cast<ezUInt32>(weight.Value() * 1000) > 0)
```
C-Core: make the 1000 a constant and maybe use a power of two like 1024?

C-Core (approved):

Code/Engine/RendererVulkan/Device/Implementation/DeviceVulkan.cpp
```diff
@@ -106,7 +107,7 @@ ezGALDeviceVulkan::ezGALDeviceVulkan(const ezGALDeviceCreationDescription& Descr
   : ezGALDevice(Description)
   , m_device(nullptr)
   //, m_pDebug(nullptr)
-  , m_uiFrameCounter(0)
+  , m_uiFrameCounter(1)
```
C-Core: move to header

Code/Engine/RendererVulkan/Pools/Implementation/DescriptorSetPoolVulkan.cpp
Sanakan8472: Done!

PR #595, opened 2022-06-21
Dynamic library support for Linux

Ingrater:
* Add support for building ezEngine as dynamic libraries on linux. All tests run through, the default wasn't changed. On linux we still build static libraries by default. All changes done rely on posix functionality. This could in theory also work on mac. Because I have no way of testing on Mac, it is linux only for now.
* Add support for dynamic plugins on linux.


jankrassnigg:
Nice!!! I'll try to take a closer look some time tomorrow, but I don't doubt it's all in order. Feel free to complete the PR as you see fit.

Ingrater:
@jankrassnigg we need to sync on this. I'm currently not sure, what should happen with the various plugin tests if the FoundationTest binary is fully statically linked.

PR #600, opened 2022-07-01
Add Curve Generator functionality to the ezEditor.

TheophilusE:
- Generate curves from predefined functions that are part of the ezMath library.

### DIFF
--- a/Code/Engine/Foundation/Math/Easing.h
+++ b/Code/Engine/Foundation/Math/Easing.h
@@ -1,0 +1,147 @@
+#pragma once
+
+#include <Foundation/Basics.h>
+#include <Foundation/Math/Declarations.h>
+
+namespace ezMath
+{
+  enum ezEasingFunctions
+  {
+    InLinear,
+    OutLinear,
+    InSine,
+    OutSine,
+    InOutSine,
+    InQuad,
+    OutQuad,
+    InOutQuad,
+    InCubic,
+    OutCubic,
+    InOutCubic,
+    InQuartic,
+    OutQuartic,
+    InOutQuartic,
+    InQuintic,
+    OutQuintic,
+    InOutQuintic,
+    InExpo,
+    OutExpo,
+    InOutExpo,
+    InCirc,
+    OutCirc,
+    InOutCirc,
+    InBack,
+    OutBack,
+    InOutBack,
+    InElastic,
+    OutElastic,
+    InOutElastic,
+    InBounce,
+    OutBounce,
+    InOutBounce
+  };
+
+  typedef double (*ezEasingFunction)(double);
+
+  /// \brief Helper function that returns an easing function pointer.
+  ezEasingFunction EZ_FOUNDATION_DLL GetEasingFunction(ezEasingFunctions easingFunction);
+
+  template <typename Type>
+  Type EaseInLinear(Type t);
+
+  template <typename Type>
+  Type EaseOutLinear(Type t);
+
+  template <typename Type>
+  Type EaseInSine(Type t);
+
+  template <typename Type>
+  Type EaseOutSine(Type t);
+
+  template <typename Type>
+  Type EaseInOutSine(Type t);
+
+  template <typename Type>
+  Type EaseInQuad(Type t);
+
+  template <typename Type>
+  Type EaseOutQuad(Type t);
+
+  template <typename Type>
+  Type EaseInOutQuad(Type t);
+
+  template <typename Type>
+  Type EaseInCubic(Type t);
+
+  template <typename Type>
+  Type EaseOutCubic(Type t);
+
+  template <typename Type>
+  Type EaseInOutCubic(Type t);
+
+  template <typename Type>
+  Type EaseInQuartic(Type t);
+
+  template <typename Type>
+  Type EaseOutQuartic(Type t);
+
+  template <typename Type>
+  Type EaseInOutQuartic(Type t);
+
+  template <typename Type>
+  Type EaseInQuintic(Type t);
+
+  template <typename Type>
+  Type EaseOutQuintic(Type t);
+
+  template <typename Type>
+  Type EaseInOutQuintic(Type t);
+
+  template <typename Type>
+  Type EaseInExpo(Type t);
+
+  template <typename Type>
+  Type EaseOutExpo(Type t);
+
+  template <typename Type>
+  Type EaseInOutExpo(Type t);
+
+  template <typename Type>
+  Type EaseInCirc(Type t);
+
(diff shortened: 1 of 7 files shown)


Code/Engine/Foundation/Math/Implementation/Easing.cpp
```diff
@@ -0,0 +1,53 @@
+
+#include <Foundation/Math/Easing.h>
+
+#include <unordered_map>
+
+namespace ezMath
+{
+  ezEasingFunction GetEasingFunction(ezEasingFunctions easingFunction)
+  {
+    static std::unordered_map<ezEasingFunctions, ezEasingFunction> easingFunctions;
```
jankrassnigg: No need for the lookup table. A switch statement would be even more efficient.

Code/Engine/Foundation/Math/Implementation/Easing_inl.h
```diff
@@ -0,0 +1,268 @@
+#pragma once
+
+#include <Foundation/Math/Math.h>
+
+namespace ezMath
+{
+
+  template <typename Type>
+  EZ_ALWAYS_INLINE Type EaseInLinear(Type t)
```
jankrassnigg, line 9: Since you are taking function pointers to these, inlining might not be possible.
But for your use case it's not critical anyway. Just an observation.

Code/Engine/Foundation/Math/Easing.h
```diff
@@ -0,0 +1,147 @@
...
+    InOutBack,
+    InElastic,
+    OutElastic,
+    InOutElastic,
+    InBounce,
+    OutBounce,
+    InOutBounce
+  };
+
+  typedef double (*ezEasingFunction)(double);
```
jankrassnigg: In New code we prefer "using" rather than typedef.

Code/Engine/Foundation/Math/Easing.h
```diff
@@ -0,0 +1,147 @@
+#pragma once
+
+#include <Foundation/Basics.h>
+#include <Foundation/Math/Declarations.h>
+
+namespace ezMath
+{
+  enum ezEasingFunctions
+  {
```
jankrassnigg, line 17: Could add a comment where the code is from and if available, add a link to a page that shows the different functions in action.

jankrassnigg (changes requested):
Very well done, great job !
Please remove the lookup table, than I'm completely happy 👍

TheophilusE:
> Very well done, great job !
> Please remove the lookup table, than I'm completely happy 👍

Will do promptly, Thanks.

Code/Engine/Foundation/Math/Implementation/Easing.cpp
TheophilusE: Resolved.

Code/Engine/Foundation/Math/Implementation/Easing_inl.h
TheophilusE: Removed the function pointer altogether, It is more useful to have the function return the value of the easing function type.

Code/Engine/Foundation/Math/Easing.h
TheophilusE: Removed the definition.

Code/Engine/Foundation/Math/Easing.h
TheophilusE: Updated somewhat, I'd love to hear your opinion on it though.

Code/Engine/Foundation/Math/Implementation/Easing.cpp
```diff
@@ -0,0 +1,115 @@
...
+namespace ezMath
+{
+  template <typename Type>
+  Type EZ_FOUNDATION_DLL GetEasingValue(ezEasingFunctions easingFunction, Type input)
+  {
+    switch (easingFunction)
+    {
+      case InLinear:
+        return EaseInLinear<Type>(input);
+        break;
```
jankrassnigg: The "break" after the "return" is redundant and some compilers may generate "unreachable code" warnings.

Code/Engine/Foundation/Math/Easing.h
```diff
@@ -38,13 +46,14 @@ namespace ezMath
     InOutElastic,
     InBounce,
     OutBounce,
-    InOutBounce
-  };
+    InOutBounce,
 
-  typedef double (*ezEasingFunction)(double);
+    EasingCount, // All easing function types must be stated before this.
```
jankrassnigg: Our convention is to call this last one "ENUM_COUNT". But not important.

Code/Engine/Foundation/Math/Easing.h
```diff
@@ -38,13 +46,14 @@ namespace ezMath
...
+  };
 
-  /// \brief Helper function that returns an easing function pointer.
-  ezEasingFunction EZ_FOUNDATION_DLL GetEasingFunction(ezEasingFunctions easingFunction);
+  /// \brief Helper function that returns the easing value from an easing function.
+  template <typename Type>
+  Type EZ_FOUNDATION_DLL GetEasingValue(ezEasingFunctions easingFunction, Type input);
```
jankrassnigg: Since it's a template, the EZ_FOUNDATION_DLL won't have any effect, the function will always be instantiated. Best to remove it.

Code/Engine/Foundation/Math/Easing.h
jankrassnigg: Yep, this is useful!

Code/Engine/Foundation/Math/Implementation/Easing_inl.h
jankrassnigg: Yes, very good, I prefer it that way as well.

jankrassnigg (changes requested):
Please remove the redundant "break"s and the DLL export on the template, then we can merge the PR :-)

Code/Engine/Foundation/Math/Implementation/Easing.cpp
TheophilusE: Missed this, will be removed shortly.

Code/Engine/Foundation/Math/Easing.h
TheophilusE: Resolved

Code/Engine/Foundation/Math/Easing.h
TheophilusE: Definitely.

jankrassnigg (approved):
Very nice, thanks a lot :)

PR #604, opened 2022-07-05
Vulkan support on linux

Ingrater:
* Add `EZ_DECL_EXPORT` and `EZ_DECL_IMPORT` as platform independent replacement of `__declspec(dllexport)` and `__declspec(dllimport)`.
 * Linux Window Setup (using xcb).
 * Check for missing dynamic symbols during linktime on linux (as opposed to at program startup). Fix some linker issues.
 * Make ShaderCompilerDXC compile on linux.
 * Make renderer tests work on linux. Fix shutdown issue in vulkan device.
 * Remove SFML download from linux azure pipeline. More parallel build for linux azure pipeline.
 * First working triangle on linux.
 * Correctly handle if the vulkan device does not support RGBA back buffers and use a BGRA back buffer instead.
 * Make asteroids sample compile on linux.
 * Better error message on unsupported back buffer format in vulkan swap chain.


Code/Engine/Core/ResourceManager/Implementation/WorkerTasks.cpp
```diff
@@ -49,6 +49,10 @@ void ezResourceManagerWorkerDataLoad::Execute()
   EZ_ASSERT_DEV(pLoader != nullptr, "No Loader function available for Resource Type '{0}'", pResourceToLoad->GetDynamicRTTI()->GetTypeName());
 
   ezResourceLoadData LoaderData = pLoader->OpenDataStream(pResourceToLoad);
+  if (LoaderData.m_pDataStream == nullptr)
+  {
+    ezLog::Warning("Loader returned no stream!");
+  }
```
jankrassnigg: I think this shouldn't be here.

Code/Engine/Core/System/Implementation/xcb/InputDevice_xcb.h
```diff
@@ -0,0 +1,20 @@
+#pragma once
+
+#include <Core/Input/DeviceTypes/MouseKeyboard.h>
+
+class EZ_CORE_DLL ezStandardInputDevice : public ezInputDeviceMouseKeyboard
```
jankrassnigg, line 5: Is this supposed to stay? I thought you wanted to use glfw instead?

Code/Engine/Core/System/Window.h
```diff
@@ -274,6 +293,11 @@ class EZ_CORE_DLL ezWindow : public ezWindowBase
 
   mutable ezWindowHandle m_WindowHandle = ezWindowHandle();
 
+#if EZ_ENABLED(EZ_SUPPORTS_SFML)
+#elif EZ_ENABLED(EZ_PLATFORM_LINUX)
```
jankrassnigg, line 297: Shouldn't there be something between the #if and #elif ? At least a #pragma warning or whatever, that this isn't implemented ?

Code/Engine/Foundation/Threading/Implementation/TaskWorkerThread.cpp
```diff
@@ -14,7 +14,7 @@ static const char* GenerateThreadName(ezWorkerThreadType::Enum ThreadType, ezUIn
 }
 
 ezTaskWorkerThread::ezTaskWorkerThread(ezWorkerThreadType::Enum ThreadType, ezUInt32 uiThreadNumber)
-  : ezThread(GenerateThreadName(ThreadType, uiThreadNumber), 32 * 1024) /* 32 KB of stack size */
+  : ezThread(GenerateThreadName(ThreadType, uiThreadNumber), 128 * 1024) /* 128 KB of stack size */
```
jankrassnigg: Maybe add a comment why at least 128 KB are needed, for future generations to not try to reduce it again.

Code/Engine/GameEngine/GameApplication/Implementation/GameApplicationInit.cpp
```diff
@@ -234,7 +234,6 @@ const char* ezGameApplication::GetActiveRenderer()
 
 void ezGameApplication::Init_SetupGraphicsDevice()
 {
-#if EZ_ENABLED(EZ_PLATFORM_WINDOWS)
```
jankrassnigg, line 237: But it only runs on Win + Linux, not on ALL platforms, no ?

Code/Engine/RendererVulkan/RendererVulkanDLL.h
```diff
@@ -5,10 +5,14 @@
...
-#    define EZ_RENDERERVULKAN_DLL __declspec(dllexport)
+#  if EZ_ENABLED(EZ_PLATFORM_WINDOWS)
+#    ifdef BUILDSYSTEM_BUILDING_RENDERERVULKAN_LIB
+#      define EZ_RENDERERVULKAN_DLL __declspec(dllexport)
+#    else
+#      define EZ_RENDERERVULKAN_DLL __declspec(dllimport)
+#    endif
 #  else
-#    define EZ_RENDERERVULKAN_DLL __declspec(dllimport)
+#    define EZ_RENDERERVULKAN_DLL __attribute__((visibility("default")))
```
jankrassnigg: Maybe we should add a macro in Foundation that wraps __declspec(dllimport) and __attribute__(...), so that we don't need to do all that in every lib.

jankrassnigg: Like EZ_DECL_EXPORT and EZ_DECL_IMPORT

Code/ThirdParty/dxc/cgmanifest.json
```diff
@@ -5,7 +5,7 @@
         "type": "git",
         "git": {
           "repositoryUrl": "https://github.com/microsoft/DirectXShaderCompiler",
-          "commitHash": "b640fa4ba4afb8eb9f15b4a34ab5187c7d94de97"
```
jankrassnigg: I can't comment on the DXC DLL above.
Is there maybe a better way to ship this? It's quite large and we probably need to update it every once in a while. Maybe we could at least move it into some submodule?

jankrassnigg: I just saw that the .so file is 25 MB, we can't commit that. Can talk offline about that.

Code/UnitTests/RendererTest/Basics/PipelineStates.cpp
```diff
@@ -778,7 +778,7 @@ void ezRendererTestPipelineStates::Timestamps()
 
   if (m_iFrame > 2 && !m_bTimestampsValid)
   {
-    if (m_bTimestampsValid = m_pDevice->GetTimestampResult(m_timestamps[0], m_GPUTime[0]).Succeeded() && m_pDevice->GetTimestampResult(m_timestamps[1], m_GPUTime[1]).Succeeded())
+    if ((m_bTimestampsValid = m_pDevice->GetTimestampResult(m_timestamps[0], m_GPUTime[0]).Succeeded() && m_pDevice->GetTimestampResult(m_timestamps[1], m_GPUTime[1]).Succeeded()))
```
jankrassnigg, line 781: The bracket seems redundant.

Code/UnitTests/RendererTest/CMakeLists.txt
```diff
@@ -12,13 +13,31 @@ target_link_libraries(${PROJECT_NAME}
...
+  target_link_libraries(${PROJECT_NAME}
+    PUBLIC
+    RendererDX11
+  )
+  ez_link_target_dx11(${PROJECT_NAME})
+  
+  add_dependencies(${PROJECT_NAME}
+    ShaderCompilerHLSL
+  )
+endif()
```
jankrassnigg: Do we need a helper function for all this?

jankrassnigg (changes requested):

Code/Engine/Core/System/Implementation/xcb/InputDevice_xcb.h
Ingrater: @Sanakan8472 wanted me to commit this intermediate state so he can cross check his future changes on linux. Once I have glfw fully working, I will remove the xcb implementation again.

Code/Engine/Core/System/Window.h
Ingrater: No, there shouldn't be anything here. Because this section contains the platform implementation specific members of ezWindow. And in the SFML case we doesn't need any members.
This will be cleaner once glfw goes in and I remove SFML completely.

Code/Engine/GameEngine/GameApplication/Implementation/GameApplicationInit.cpp
Ingrater: Currently win + linux are the only supported platforms, correct. But in theory we can support all platforms that we can run vulkan on (android, MacOS). I just wanted to avoid that the next person who adds a new platform has to change this place again. After all this function does not have any platform specific code. It only uses our abstracted interfaces. So I don't see any reason to make it platform specific.

Code/ThirdParty/dxc/cgmanifest.json
Ingrater: The windows dll is 18 mb, and that is also checked in, so I didn't think much about it. @Sanakan8472 want's to do a seperate submodule with all the vulkan dependnecies.

Code/UnitTests/RendererTest/Basics/PipelineStates.cpp
Ingrater: No, it is not. It silences a warning in clang, letting clang now that we indeed want to do an asignment here (and not a ==)

Code/UnitTests/RendererTest/CMakeLists.txt
Ingrater: yeah, might be worth doing going forward

Code/ThirdParty/dxc/cgmanifest.json
jankrassnigg: Yes, but checking that in was already a bad idea. We should fix this now.

Ingrater: I will remove all changes to the ThirdParty/dxc folder. This will however mean that anyone wanting to build on linux right now, will have to obtain the dxc headers + library themself.

jankrassnigg: Fine for the time being.

Code/Samples/Asteroids/CMakeLists.txt
```diff
@@ -1,6 +1,10 @@
 ez_cmake_init()
 
-ez_requires_d3d()
+if(EZ_CMAKE_PLATFORM_WINDOWS)
```
jankrassnigg: ez_requires_renderer

Code/UnitTests/RendererTest/CMakeLists.txt
```diff
@@ -1,7 +1,8 @@
 ez_cmake_init()
 
-
-ez_requires_d3d()
+if ((NOT EZ_CMAKE_PLATFORM_WINDOWS) AND (NOT EZ_CMAKE_PLATFORM_LINUX))
```
jankrassnigg: ez_requires_one_of ?

jankrassnigg (approved):

Code/BuildSystem/CMake/ezUtilsCppFlags.cmake
```diff
@@ -261,20 +267,25 @@ function(ez_set_build_flags_gcc TARGET_NAME)
...
-			if(NOT ("${CMAKE_EXE_LINKER_FLAGS}" MATCHES "fuse-ld="))
-				target_link_options(${TARGET_NAME} PRIVATE -B ${MOLD_DIR})
-			endif()
-		elseif("${TARGET_TYPE}" STREQUAL "EXECUTABLE")
-			if(NOT ("${CMAKE_SHARED_LINKER_FLAGS}" MATCHES "fuse-ld="))
-				target_link_options(${TARGET_NAME} PRIVATE -B ${MOLD_DIR})
-			endif()
-		endif()
-	endif()
+	#if (MOLD_PATH)
```
Sanakan8472: I've never heared of mold, is it broken and should we remove support for it or why was it commented out?

Code/BuildSystem/CMake/ezUtilsVars.cmake
```diff
@@ -59,4 +59,7 @@ mark_as_advanced(FORCE EZ_ENABLE_COMPILER_STATIC_ANALYSIS)
 ### Qt
 # set (EZ_VCPKG_INSTALL_QT OFF CACHE BOOL "Whether to install Qt via vcpkg.")
 
-
+######################################
+### SFML
```
Sanakan8472, line 63: Shouldn't we remove all SFML references instead?

Code/Engine/RendererVulkan/Device/Implementation/DeviceVulkan.cpp
```diff
@@ -197,7 +206,10 @@ vk::Result ezGALDeviceVulkan::SelectDeviceExtensions(vk::DeviceCreateInfo& devic
   };
 
   VK_SUCCEED_OR_RETURN_LOG(AddExtIfSupported(VK_KHR_SWAPCHAIN_EXTENSION_NAME, m_extensions.m_bDeviceSwapChain));
-  AddExtIfSupported(VK_EXT_SHADER_VIEWPORT_INDEX_LAYER_EXTENSION_NAME, m_extensions.m_bShaderViewportIndexLayer);
+  if (AddExtIfSupported(VK_EXT_SHADER_VIEWPORT_INDEX_LAYER_EXTENSION_NAME, m_extensions.m_bShaderViewportIndexLayer) != vk::Result::eSuccess)
+  {
+    ezLog::Warning("The optional extension '" VK_EXT_SHADER_VIEWPORT_INDEX_LAYER_EXTENSION_NAME "' is not present");
```
Sanakan8472: I've added a general warning if an extension can't be found in the AddExtIfSupported lambda above so you can remove this block and just call AddExtIfSupported and ignore the return value for optional extensions.

Code/Engine/RendererVulkan/Device/Implementation/DeviceVulkan_inl.h
```diff
@@ -1,4 +1,4 @@
-#include "..\DeviceVulkan.h"
+#include "../DeviceVulkan.h"
```
Sanakan8472: Isn't this a circular dependency and should be removed entirely?

Sanakan8472 (approved):

Code/BuildSystem/CMake/ezUtilsCppFlags.cmake
Ingrater: mold is a modern fast linker for linux (https://github.com/rui314/mold). It is completely optional in ez. Using it reduces our link times by a factor of 10. I never intended to commit this change, this was just to get the same linker error as on CI.

Code/BuildSystem/CMake/ezUtilsVars.cmake
Ingrater: I will remove all of SFML together with my glfw work.

PR #606, opened 2022-07-10
GLFW support for linux

Ingrater:
This PR includes:
* Add glfw 3.3.7 to our third party directory
* Modify glfw cmake scripts to be compatible with ours
* Implement window management using glfw (resize, close, etc all work)
* Implement Mouse & Keyboard input using glfw
* Add ezScreen implementation based on glfw
* Remove SFML and xcb completely
* Fix issue in plugin system, where we keep plugins which failed to load in the "g_LoadedModules" list.

This is a first iteration of glfw support. In my next PR I plan to
* Add gamepad support pased on glfw
* Check if it is possible to implement all of our mouse capture modes using glfw and do so.
* Possibly: Glfw support for windows

The idea with this PR is that I won't squash the commits. Instead I will rebase the PR once all reviewing is done so that there are only 2 commits left in the PR: 
* A commit containing the original glfw source code
* A commit with the remaining changes

This will allow us going forward to easily see which changes we made to the glfw source in case we need to update it.


Code/Engine/Core/Input/Declarations.h
```diff
@@ -256,6 +256,7 @@ EZ_DECLARE_FLAGS_OPERATORS(ezInputSlotFlags);
 #define ezInputSlot_KeyRightAlt       "keyboard_right_alt"
 #define ezInputSlot_KeyLeftWin        "keyboard_left_win"
 #define ezInputSlot_KeyRightWin       "keyboard_right_win"
+#define ezInputSlot_KeyMenu           "keyboard_menu"
```
jankrassnigg: Can't remember anymore, but maybe key_apps is meant to be the same.

jankrassnigg (approved):

Code/Engine/RendererVulkan/CommandEncoder/Implementation/CommandEncoderImplVulkan.cpp
```diff
@@ -749,7 +749,8 @@ void ezGALCommandEncoderImplVulkan::GenerateMipMapsPlatform(const ezGALResourceV
     }
     else
     {
-      EZ_ASSERT_NOT_IMPLEMENTED;
+      // Do not commit!
```
Sanakan8472: It says "Do not commit!"?

Sanakan8472 (approved):

Code/Engine/Core/Input/Declarations.h
Ingrater: It indeed was the same.

Code/Engine/RendererVulkan/CommandEncoder/Implementation/CommandEncoderImplVulkan.cpp
Ingrater: Reverted.

PR #609, opened 2022-07-18
Gamepad input using GLFW

Ingrater:
* Add GLFW support on windows (optional, default off)
* Implement Gamepad input using GLFW
* Add central access point for getting an controller input device. (ezControllerInput)

### DIFF
--- a/Code/Engine/Core/System/ControllerInput.h
+++ b/Code/Engine/Core/System/ControllerInput.h
@@ -1,0 +1,15 @@
+#pragma once
+
+#include <Core/CoreDLL.h>
+
+class ezInputDeviceController;
+
+class EZ_CORE_DLL ezControllerInput
+{
+public:
+  // \brief Returns the global controller input device. May be nullptr.
+  static ezInputDeviceController* GetDevice();
+
+  // \brief Set the global controller input device.
+  static void SetDevice(ezInputDeviceController* pDevice);
+};
--- a/Code/Engine/Core/System/ControllerInput.h
+++ b/Code/Engine/Core/System/ControllerInput.h
@@ -7,6 +7,9 @@ class ezInputDeviceController;
 class EZ_CORE_DLL ezControllerInput
 {
 public:
+  // \brief Returns if a global controller input device exists.
+  static bool HasDevice();
+
   // \brief Returns the global controller input device. May be nullptr.
   static ezInputDeviceController* GetDevice();
 
--- a/Code/Engine/Core/System/Implementation/glfw/ControllerInput_glfw.inl
+++ b/Code/Engine/Core/System/Implementation/glfw/ControllerInput_glfw.inl
@@ -1,0 +1,227 @@
+#include <Core/Input/DeviceTypes/Controller.h>
+#include <Core/Input/InputManager.h>
+#include <Foundation/Configuration/Startup.h>
+#include <Foundation/Types/UniquePtr.h>
+
+#include <GLFW/glfw3.h>
+
+class ezControllerInputGlfw : public ezInputDeviceController
+{
+public:
+  virtual void InitializeDevice() override;
+
+  virtual void UpdateInputSlotValues() override;
+
+  virtual void ResetInputSlotValues() override;
+
+  virtual void RegisterInputSlots() override;
+
+  virtual bool IsControllerConnected(ezUInt8 uiPhysical) const override;
+
+private:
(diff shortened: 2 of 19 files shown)


Code/Engine/Core/System/ControllerInput.h
```diff
@@ -0,0 +1,15 @@
...
+
+#include <Core/CoreDLL.h>
+
+class ezInputDeviceController;
+
+class EZ_CORE_DLL ezControllerInput
+{
+public:
+  // \brief Returns the global controller input device. May be nullptr.
+  static ezInputDeviceController* GetDevice();
```
jankrassnigg, line 14: Could add a "HasDevice" function for non-null checks.

Code/Engine/Core/System/Implementation/glfw/ControllerInput_glfw.inl
```diff
@@ -0,0 +1,227 @@
+#include <Core/Input/DeviceTypes/Controller.h>
```
jankrassnigg, line 1: Why is there an empty cpp file ?

Code/Engine/Core/System/Implementation/glfw/ControllerInput_glfw.inl
```diff
@@ -0,0 +1,227 @@
...
+}
+
+void ezControllerInputGlfw::SetDeadZone(const char* szButton)
+{
+  ezStringBuilder s;
+
+  for (ezInt32 i = 0; i < MaxControllers; ++i)
+  {
+    s.Format("controller{0}_{1}", i, szButton);
+    ezInputManager::SetInputSlotDeadZone(s.GetData(), 0.23f);
```
jankrassnigg, line 115: what's the purpose of this ?

Code/Samples/Asteroids/GameState.cpp
```diff
@@ -71,12 +69,10 @@ void AsteroidGameState::BeforeWorldUpdate()
 
 void AsteroidGameState::ConfigureInputActions()
 {
-#if EZ_ENABLED(EZ_PLATFORM_WINDOWS)
-  ezInputDeviceXBox360::GetDevice()->EnableVibration(0, true);
-  ezInputDeviceXBox360::GetDevice()->EnableVibration(1, true);
-  ezInputDeviceXBox360::GetDevice()->EnableVibration(2, true);
-  ezInputDeviceXBox360::GetDevice()->EnableVibration(3, true);
-#endif
+  ezControllerInput::GetDevice()->EnableVibration(0, true);
```
jankrassnigg: Wouldn't you need to wrap all of these calls in a check? Unless you provide a fallback NULL implementation for controller devices.

Code/ThirdParty/glfw/finalize.cmake
```diff
@@ -11,4 +11,10 @@ if(TARGET glfw)
 
 
     set_property(GLOBAL APPEND PROPERTY "EXPORT_PROJECTS" glfw)
+    
+    ez_set_project_ide_folder(glfw "Code/ThirdParty/GLFW")
+    ez_set_project_ide_folder(update_mappings "Code/ThirdParty/GLFW")
```
jankrassnigg, line 16: what is this ?

jankrassnigg (approved):

Code/Engine/Core/System/Implementation/glfw/ControllerInput_glfw.inl
Ingrater: Accident, should not be there.

Code/Engine/Core/System/Implementation/glfw/ControllerInput_glfw.inl
Ingrater: I copied the SetDeadZone code straight from the XBox controller plugin. I though this is "the way to go"

Code/Samples/Asteroids/GameState.cpp
Ingrater: You are right. Will wrap.

Code/ThirdParty/glfw/finalize.cmake
Ingrater: GLFW creates two cmake targets. One called "glfw" and the other one called "update_mappings" (which generates some lookup tables). I move both of them into the Code/ThirdParty filter within the visual studio solution.

PR #614, opened 2022-07-21
Only ask for the validation layer if it is actually present

Ingrater:
* Only request the Vulkan validation layer if it exists.
* Do not hide the mouse in renderer tests by default, as Linux platforms have issues showing the mouse during debugging.

### DIFF
--- a/Code/Engine/RendererVulkan/Device/Implementation/DeviceVulkan.cpp
+++ b/Code/Engine/RendererVulkan/Device/Implementation/DeviceVulkan.cpp
@@ -39,28 +39,53 @@
 #  include <GLFW/glfw3.h>
 #endif
 
-VKAPI_ATTR VkBool32 VKAPI_CALL debugCallback(VkDebugUtilsMessageSeverityFlagBitsEXT messageSeverity, VkDebugUtilsMessageTypeFlagsEXT messageType, const VkDebugUtilsMessengerCallbackDataEXT* pCallbackData, void* pUserData)
+EZ_DEFINE_AS_POD_TYPE(VkLayerProperties);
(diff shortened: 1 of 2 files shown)


Code/Engine/RendererVulkan/Device/Implementation/DeviceVulkan.cpp
```diff
@@ -39,28 +39,53 @@
...
-      break;
-    case VK_DEBUG_UTILS_MESSAGE_SEVERITY_ERROR_BIT_EXT:
-      ezLog::Error("VK error: {}", pCallbackData->pMessage);
-      break;
-    default:
-      break;
+    switch (messageSeverity)
+    {
+      case VK_DEBUG_UTILS_MESSAGE_SEVERITY_VERBOSE_BIT_EXT:
+        ezLog::Debug("VK Verbose: {}", pCallbackData->pMessage);
```
jankrassnigg: That it is a "verbose / info / error" message, is already covered by the ezLog function that you are calling. I would only prefix it with "VK:" or "Vulkan:"

jankrassnigg (approved):

Code/Engine/RendererVulkan/Device/Implementation/DeviceVulkan.cpp
Ingrater: I did not write that code. Only it's intendation changed. Should I still change the message as part of this PR?

jankrassnigg: Interesting. I leave that up to you.

PR #615, opened 2022-07-22
Inspector support on linux

Ingrater:
* Make third party ADS compile on Linux
* Make GuiFoundation compile on Linux
* Make Inspector compile on Linux

### DIFF
--- a/Code/BuildSystem/CMake/ezUtilsQt.cmake
+++ b/Code/BuildSystem/CMake/ezUtilsQt.cmake
@@ -71,7 +71,7 @@ function(ez_prepare_find_qt)
         endif()
 
         if (EZ_CMAKE_PLATFORM_LINUX)
-            set (Qt5WinExtras_DIR "Qt5X11Extras_DIR-NOTFOUND" CACHE PATH "" FORCE)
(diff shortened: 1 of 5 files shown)


Code/BuildSystem/CMake/ezUtilsQt.cmake
```diff
@@ -69,6 +69,10 @@ function(ez_prepare_find_qt)
         if (EZ_CMAKE_PLATFORM_WINDOWS)
             set (Qt5WinExtras_DIR "Qt5WinExtras_DIR-NOTFOUND" CACHE PATH "" FORCE)
         endif()
+
+        if (EZ_CMAKE_PLATFORM_LINUX)
+            set (Qt5WinExtras_DIR "Qt5X11Extras_DIR-NOTFOUND" CACHE PATH "" FORCE)
```
Ingrater: Qt5WinExtras_DIR -> Qt5X11Extras_DIR

jankrassnigg:
Btw. we haven't updated ADS in quite a while, it may have a bunch of general improvements. But we also haven't updated Qt in a long time, probably we would need both (not sure whether we can use a new ADS with our current Qt version).
Updating Qt5 to Qt6 is actually unproblematic in my experience. It's mostly the problem that we need to generate new libs and so on.

jankrassnigg (approved):

PR #640, opened 2022-08-06
AudioSystem Plugin

na2axl:
The AudioSystem is a plugin abstracting most of the features provided by audio engines like Wwise, FMOD or Amplitude. But actually, any sound engine can be used as middleware, even really small ones, like SoLoud.

The AudioSystem is splitted in a runtime plugin (ezAudioSystemPlugin), an editor plugin (ezEditorPluginAudioSystem), and a static library (ezSharedPluginAudioSystem) which will be used by the middleware's editor plugins to implement required editor actions (eg: reload controls).

## The Architecture

The implemented AudioSystem architecture is a bit inspired to what CryEngine is offering, which is quite good. The AudioSystem is defined by 03 major actors:
- The Audio System itself
- The Audio Translation Layer
- The Audio Middleware

### The Audio System (`ezAudioSystem`)

This is the singleton class directly communicating to the core of EZ, by implementing the provided `ezSoundInterface`. His main purpose is to receive audio requests and collect controls registered through AudioControlCollection resources. Audio requests sent to the system are asynchronously processed in a separated thread (the Audio Thread). There is also a possibility to send synchronous requests, which blocks the caller thread until processed. An audio request can have an optional callback, which is guaranteed to be executed on the caller thread only once, after the audio request has been processed.

### The Audio Translation Layer (`ezAudioTranslationLayer`)

That singleton class is the bridge between the Audio System and Audio Middleware. Audio requests received by the Audio System are dispatched to this class to be processed. That always happens in the Audio Thread. The Audio Translation Layer also stores the controls registered through the Audio System by loaded Audio Control Collection resources.

### The Audio Middleware (`ezAudioMiddleware`)

This is the interface class that should be implemented by audio middlewares. It declares all the methods needed by the Audio System to communicate with 3rd party audio engines.

## The resources

For now, the only proposed resource is the Audio Control Collection. It's like the asset collection but only group audio controls generated by Audio Middleware. That resource helps to abstract audio events, parameters, effects, etc. with the same name across audio engines.
Demo
 
Each entry has a name, a type, and the path to the control file generated to the middleware. The file will be embedded into the asset once transformed.

## The components

The plugin will expose most of its functionalities (real-time parameters, environment effect, listeners, etc.) as components to the user through the editor and scripts. Currently only 02 components are ready:
- Audio Controls Component (`ezAudioControlsComponent`): Loads the given Audio Control Collection asset once initialized. This will ensure that the Audio System will be able to lookup control IDs using the control names provided in the asset.
image

- Audio Trigger Component (`ezAudioTriggerComponent`): `Play` or `Stop` audio by triggering events. This component gets as properties the `PlayTrigger` name and an optional `StopTrigger` name. Those names should be known by the Audio System, otherwise nothing will happen. It is also possible to provide a `SoundObstructionType` value, which is used by the Audio System to know how to deal with obstruction/occlusion computation with that particular audio event.
image


na2axl:
Added an Audio RTPC Component (`ezAudioRtpcComponent`), allowing to set the value of parameters exposed by the audio middleware.
image

na2axl:
Added an Audio Listener Component (`ezAudioListenerComponent`) for rendering audio. The component can have 02 optional parameters to change who is providing position and orientation data. That can be useful for some use cases. But when not set, the component uses the owner's transformation.
image

na2axl:
Closing as discussed in https://github.com/ezEngine/ezEngine/discussions/610#discussioncomment-3349946

PR #647, opened 2022-08-13
Fix shared library build on linux.

Ingrater:
* Fix shared library build on linux (which was broken during the recent MSFT contributions merge)
* Make the EZ_COMPILE_ENGINE_AS_DLL default on when building on Linux.
* Remove EZ_ALIGN, EZ_ALIGN_X and ezAligned<x> and replace with c++11 alignas.
* Change warning / error settings so that incorrectly placed attributes (EZ_FOUNDATION_DLL and others) cause an compile error.


Code/BuildSystem/CMake/ezUtils.cmake
```diff
@@ -111,19 +111,19 @@ function(ez_set_target_output_dirs TARGET_NAME LIB_OUTPUT_DIR DLL_OUTPUT_DIR)
 
 	set_target_properties(${TARGET_NAME} PROPERTIES
 		RUNTIME_OUTPUT_DIRECTORY_${EZ_BUILDTYPENAME_DEBUG_UPPER} "${OUTPUT_DLL_DEBUG}"
-		LIBRARY_OUTPUT_DIRECTORY_${EZ_BUILDTYPENAME_DEBUG_UPPER} "${OUTPUT_LIB_DEBUG}"
+		LIBRARY_OUTPUT_DIRECTORY_${EZ_BUILDTYPENAME_DEBUG_UPPER} "${OUTPUT_DLL_DEBUG}"
```
jankrassnigg, line 114: On Windows this will put .lib files into the directory where the .exe files are.

Ingrater: Not according to the cmake documentation.

https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#library-output-artifacts

I'll retest on windows.

Ingrater: @jankrassnigg I can not confirm your statement. When I build with these changes on windows not a single .lib file is in the Output/Bin folder. Have you had this issue on your machine?

jankrassnigg: Maybe it was like that in the past. If it works as desired now, then I'm fine with the change.

jankrassnigg (approved):

Code/BuildSystem/CMake/CMakeUtils/ezUtilsPlatformVars.cmake
```diff
@@ -0,0 +1,11 @@
+# Cmake variables which are platform dependent
+
+# #####################################
+# ## General settings
+# #####################################
+if(EZ_CMAKE_PLATFORM_WINDOWS OR EZ_CMAKE_PLATFORM_LINUX)
```
jankrassnigg: Why was this moved into a separate file?

CMakeLists.txt
```diff
@@ -51,6 +51,8 @@ endif()
 
 ez_pull_platform_vars()
 
+include("CMakeUtils/ezUtilsPlatformVars")
```
jankrassnigg: This line should not be necessary. All files in this folder are automatically included.

Code/BuildSystem/CMake/CMakeUtils/ezUtilsPlatformVars.cmake
Ingrater: Because otherwise EZ_CMAKE_PLATFORM_WINDOWS and EZ_CMAKE_PLATFORM_LINUX are not defined yet.

CMakeLists.txt
Ingrater: The entire point of this include is to make sure that the EZ_CMAKE_PLATFORM ... vars are already defined.
In fact I don't want this file to be auto included. Where do I have to place it, so that it is not auto included?

Ingrater: nvm found it.

jankrassnigg: Literally into any other folder, even sub folders would work.

Code/Engine/Foundation/Basics.h
```diff
@@ -10,14 +10,19 @@
 #include <Foundation/UserConfig.h>
 
 // Configure the DLL Import/Export Define
+// EZ_FOUNDATION_DLL_CPP17 should be used if a class or struct has additional c++17 attributes e.g.
+// struct [[nodiscard]] EZ_FOUNDATION_DLL_CPP17 StructName { ... };
 #if EZ_ENABLED(EZ_COMPILE_ENGINE_AS_DLL)
 #  ifdef BUILDSYSTEM_BUILDING_FOUNDATION_LIB
 #    define EZ_FOUNDATION_DLL EZ_DECL_EXPORT
+#    define EZ_FOUNDATION_DLL_CPP17 EZ_DECL_EXPORT_CPP17
```
Ingrater: Looking for a better name for this. Maybe `EZ_FOUNDATION_DLL_CPP_ATTR` ?
And no we can't always use the [[attr]] syntax as this is not valid in all places we need.

jankrassnigg: So when can you use the one Syntax and when the other?  Functions versus classes? Maybe that can help to find a better name.

Ingrater: Looking at it closer it seems the only problems are statements like this:
`friend EZ_FOUNDATION_DLL void ezRun(ezApplication* pApplicationInstance);`
of which we have 6 in foundation.
I have the hunch that in this case the `EZ_FOUNDATION_DLL` does not actually do anything, as its only a friend declaration. Let me try that out

Ingrater: Ok so the main problem was that the `[[gnu::visibility("default")]]` attribute would conflict with `__attribute__((alignas(X)))`.
I pretty much removed all usage of our `EZ_ALIGN` macro and replaced it with c++11 `alignas`.

We can now use `[[gnu::visibility("default")]]` everywhere.

Ingrater: The only reason EZ_ALIGN still exists is because we use EZ_ALIGN_16 in our shader source code.
We could replace EZ_ALIGN_16 with something like EZ_SHADER_STRUCT, and then get rid of EZ_ALIGN completely.

jankrassnigg: Sounds good.

Ingrater: I was able to get completely rid of most EZ_ALIGN and related macros (only EZ_ALIGNMENT_OF and EZ_MINIMUM_ALIGNMENT still exist).

Unfortunately declaring a friend function on windows is a special case and we now have a new macro for that. In foundation for example we now have EZ_FOUNDATION_DLL_FRIEND. In core EZ_CORE_DLL_FRIEND. We only have like 8 cases where we need these macros, so most developers should not need to care about them.

Code/BuildSystem/CMake/CMakeUtils/ezUtilsCppFlags.cmake
```diff
@@ -248,6 +248,16 @@ function(ez_set_build_flags_gcc TARGET_NAME)
...
 	target_compile_options(${TARGET_NAME} PRIVATE -Wno-multichar)
 
+	if(NOT(CMAKE_CURRENT_SOURCE_DIR MATCHES "Code/ThirdParty"))
+		# Warning / Error settings for ez code
+		# attributes = error if a attribute is placed incorrectly (e.g. EZ_FOUNDATION_DLL)
+		# unused-result = error if [[nodiscard]] return value is not handeled (ezResult)
+		target_compile_options(${TARGET_NAME} PRIVATE -Werror=attributes -Werror=unused-result -Wno-ignored-attributes)
+	else()
+		# Ignore all warnings in third party code.
+		target_compile_options(${TARGET_NAME} PRIVATE -Wno-everything)
```
jankrassnigg, line 258: This is going to create compliance issues in certain projects

Ingrater: How so? 
* I copied this block from ez_set_build_flags_clang where it already exists. Why is it not causing compliance issues there?
* Ast this only affects gcc builds and I'm pretty sure that certain projects don't use gcc at all, how does it affect them?

jankrassnigg: Hm, if it's GCC only, then maybe it's no problem. For MSVC builds compliance requires that a whole bunch of warnings are not generally disabled.
Not a big problem, if there are going to be issues at some point, I'll find a solution.

PR #649, opened 2022-08-13
Additional Vulkan Features

Sanakan8472:
- Various validation layer fixes.
  - Check for supported stages during device creation.
- Disable Vulkan sampler creation anisotropyEnable if m_Description.m_MipFilter == ezGALTextureFilterMode::Anisotropic is not set. This fixes unit tests using SwiftShader.
- All newly created textures are now cleared to zero memory to fix unit tests on Intel GPUs.
- Implemented explicit clear render targets command.
- SwapChains are now resized in place via the new **UpdateSwapChain** function. The same function is now used to change the present mode. This is necessary as the old model crashed on NVidia cards.
- ImageCopyVulkan class:
  - Readback and generate mipmaps is now implemented partially for NVidia cards which don't support Blit and thus must use a manual render pass to do the copy.
  - Only supports 1 layer and 1 mip level on NVidia hardware for now.
  - Low-level renderer does not have access to resource loading so a callback was added to retrieve a hard-coded shader via **ezShaderUtils::RequestBuiltinShader**.
- Fixed some bugs in unit tests like missing image comparisions.


Code/Engine/RendererCore/RenderContext/Implementation/RenderContext.cpp
```diff
@@ -853,6 +855,56 @@ ezGALSamplerStateHandle ezRenderContext::GetDefaultSamplerState(ezBitflags<ezDef
...
+      break;
+    case ezShaderUtils::ezBuiltinShaderType::DownscaleImage:
+      hActiveShader = ezResourceManager::LoadResource<ezShaderResource>("Shaders/Pipeline/Downscale.ezShader");
+      break;
+  }
+
+  EZ_ASSERT_DEV(hActiveShader.IsValid(), "Could not load builtin shader!");
+
+  if (!hActiveShader.IsValid())
+    return;
```
Ingrater: Output an error instead of silently returning?

Code/Engine/RendererCore/RenderContext/Implementation/RenderContext.cpp
```diff
@@ -853,6 +855,56 @@ ezGALSamplerStateHandle ezRenderContext::GetDefaultSamplerState(ezBitflags<ezDef
...
+  static ezHashedString sFalse = ezMakeHashedString("FALSE");
+  if (ezGALDevice::GetDefaultDevice()->GetCapabilities().m_bVertexShaderRenderTargetArrayIndex)
+    permutationVariables.Insert(sVSRTAI, sTrue);
+  else
+    permutationVariables.Insert(sVSRTAI, sFalse);
+
+  ezShaderPermutationResourceHandle hActiveShaderPermutation = ezShaderManager::PreloadSinglePermutation(hActiveShader, permutationVariables, false);
+
+  if (!hActiveShaderPermutation.IsValid())
+    return;
```
Ingrater: Output an error instead of silently returning?

Code/Engine/RendererCore/RenderContext/Implementation/RenderContext.cpp
```diff
@@ -853,6 +855,56 @@ ezGALSamplerStateHandle ezRenderContext::GetDefaultSamplerState(ezBitflags<ezDef
...
+
+  ezShaderPermutationResource* pShaderPermutation = ezResourceManager::BeginAcquireResource(
+    hActiveShaderPermutation, ezResourceAcquireMode::BlockTillLoaded);
+
+  if (!pShaderPermutation->IsShaderValid())
+  {
+    ezResourceManager::EndAcquireResource(pShaderPermutation);
+    return;
```
Ingrater: Output error?

Ingrater (approved):

PR #652, opened 2022-08-15
Updated Advanced Docking System to the latest version

### LINKED ISSUE #635: Update Advanced Docking System
EZ uses ADS: https://github.com/githubuser0xFFFF/Qt-Advanced-Docking-System

The update process is like this:

In Code\ThirdParty\ads
1. Delete all ADS files (cpp, h, qrc, css, png).
2. The CMake files are from EZ and should stay.
3. Copy the new ADS files over.
4. Run CMake, reload Visual Studio and see where stuff breaks.
5. Fix stuff.
6. Test the editor and ezInspector. Also check for #532.

Not sure whether the new ADS will still work with our older Qt version. If not, we need to see what to do (due to Linux support there is a reason why we are not on Qt 6 yet).

jankrassnigg:
Fixes #635

### DIFF
--- a/Code/BuildSystem/AzurePipelines/Linux-x64.yml
+++ b/Code/BuildSystem/AzurePipelines/Linux-x64.yml
@@ -23,7 +23,7 @@ jobs:
       targetType: inline
       script: |
         sudo apt-get update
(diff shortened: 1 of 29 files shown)


Ingrater:
I think on linux the QT Component "Gui-Private" is missing from CMakeLists.txt. That's why you get the missing header error.

PR #653, opened 2022-08-15
Experimental Editor support on Linux

Ingrater:
* Make EditorEngineProcess work on linux
* Fix various invocations of .exe executables to work on linux
* Lots of compile fixes for linux
* Add EZ_EXPERIMENTAL_EDITOR_ON_LINUX cmake variable to enable building the editor on linux
* Fix shared and exclusive sharing of files in ezOSFile_linux


Code/Editor/EditorEngineProcess/EngineProcGameApp.cpp
```diff
@@ -127,6 +128,15 @@ void ezEngineProcessGameApplication::LogWriter(const ezLoggingEventData& e)
   if (msg.m_Entry.m_Type == ezLogMsgType::Flush)
     return;
 
+  if (msg.m_Entry.m_sTag == "IPC")
+    return;
+
+  // Prevent infinite recursion by disabeling logging until we are done sending the message
+  ezLogInterface* oldLogSystem = ezLog::GetThreadLocalLogSystem();
```
jankrassnigg: Iirc there is an ez log mute macro for this.

Code/Editor/EditorEngineProcessFramework/IPC/ProcessCommunicationChannel.cpp
```diff
@@ -101,6 +101,12 @@ ezResult ezProcessCommunicationChannel::WaitForMessage(const ezRTTI* pMessageTyp
   while (m_pWaitForMessageType != nullptr)
   {
     m_pChannel->WaitForMessages();
+    if (!m_pChannel->IsConnected())
```
jankrassnigg, line 104: Why was this additional check necessary?

Code/Editor/EditorFramework/CMakeLists.txt
```diff
@@ -7,7 +7,7 @@ get_filename_component(PROJECT_NAME ${CMAKE_CURRENT_SOURCE_DIR} NAME_WE)
 
 ez_create_target(LIBRARY ${PROJECT_NAME})
 
-ez_link_target_qt(TARGET ${PROJECT_NAME} COMPONENTS Core Gui Widgets WinExtras Svg)
+ez_link_target_qt(TARGET ${PROJECT_NAME} COMPONENTS Core Gui Widgets Network WinExtras Svg)
```
jankrassnigg: What do we use the network stuff for?

Code/Engine/Foundation/IO/Implementation/Posix/OSFile_posix.h
```diff
@@ -106,7 +163,10 @@ ezResult ezOSFile::InternalWrite(const void* pBuffer, ezUInt64 uiBytes)
   while (uiBytes > uiBatchBytes)
   {
     if (fwrite(pBuffer, 1, uiBatchBytes, m_FileData.m_pFileHandle) != uiBatchBytes)
+    {
+      ezLog::Error("fwrite 1gb failed for '{}'", m_sFileName);
```
jankrassnigg: 1gb??

jankrassnigg (approved):

Code/Editor/EditorEngineProcessFramework/IPC/ProcessCommunicationChannel.cpp
Ingrater: I had the following situation
* The editor would wait for a message on the QT main thread (without timeout)
* The EditorEngineProcess would crash before sending the message
* The editor now hangs in this loop forever. As this happens on the QT main thread, the editor would also be unresponsive, e.g. all of the UI is unusable.

Code/Engine/Foundation/IO/Implementation/Posix/OSFile_posix.h
Ingrater: What exactly is the question here? This is a batch write of 1GB size. I wanted to differentiate between this error message and the one a bit further down, so that you can see from the error message if a batch write failed or the final write.

Code/Editor/EditorFramework/CMakeLists.txt
Ingrater: I actually don't know where this came from. But it's not needed, so I removed it.

jankrassnigg (approved):

PR #654, opened 2022-08-15
Fix project build system

na2axl:
The last update on the EZ build system has broken the CppSource template. The editor was not able to generate the solution.  This little change fixes the issue.


jankrassnigg:
Very good, thanks a lot for the fix!

PR #662, opened 2022-08-19
Add BC1,BC6,BC7 CPU based compression

Ingrater:
* Stop gap solution for missing GPU accelerated compression of these formats on linux.


Code/ThirdParty/sal/sal.h
```diff
@@ -0,0 +1,2961 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+/***
+*sal.h - markers for documenting the semantics of APIs
```
jankrassnigg, line 5: Are you trying to trigger my PTSD ?

jankrassnigg (approved):

Code/ThirdParty/sal/sal.h
Ingrater: Not intentionally. But now that you mention it, this is a nice added bonus ;-)

PR #664, opened 2022-08-23
Changed some basic string functionality to take ezStringView instead of const char*

jankrassnigg:


Ingrater (commented):
If we move more stuff to ezStringView we should make sure that the constructors of ezStringView are `constexpr` where possible so that the compiler can still evaluate statements involving literals at compile time.

Also we are missing a ezStringView constructor for a string literal:
```
template<size_t N>
ezStringView(const char (&str)[N]) { ... }
```
Where N is the length of the string inferred at compile time.

jankrassnigg:
> If we move more stuff to ezStringView we should make sure that the constructors of ezStringView are `constexpr` where possible so that the compiler can still evaluate statements involving literals at compile time.
> 
> Also we are missing a ezStringView constructor for a string literal:
> 
> ```
> template<size_t N>
> ezStringView(const char (&str)[N]) { ... }
> ```
> 
> Where N is the length of the string inferred at compile time.

Ah didn't see your comment before. Yes, you are right, I was thinking about that as well, but no time to implement right now.

PR #667, opened 2022-08-26
Provide optimal ways to construct a ezStringView from a literal

Ingrater:
I looked at codegen of msvc,gcc and clang regarding constructing a ezStringView from a string literal. e.g. `ezStringView("Hello World"`
* gcc and clang will generate optimal code with optimizations enabeld
* msvc never generates optimal code, even with optimizations enabled
* gcc and clang generate bad code when optimizations are disabled

The goal is to have a option to get good codegen even in debug builds. For this we now can:

Create constexpr versions of ezStringView. E.g.
`constexpr auto someConstant = ezStringView("bla");`

Or you can use the newly added `_sv` literal suffix:
```
using namespace ezLiterals;

SomeFunctionThatTakesStringView("Hello World"_sv);
```
Going forward we might need to make the `const char*` constructor for ezStringView `explicit`, so that we avoid "bad code" by default.

### DIFF
--- a/Code/Engine/Foundation/Strings/StringView.h
+++ b/Code/Engine/Foundation/Strings/StringView.h
@@ -25,13 +25,13 @@ public:
   ezStringView();
 
   /// \brief Creates a string view starting at the given position, ending at the next '\0' terminator.
-  ezStringView(const char* pStart); // [tested]
+  constexpr ezStringView(const char* pStart); // [tested]
 
   /// \brief Creates a string view for the range from pStart to pEnd.
   ezStringView(const char* pStart, const char* pEnd); // [tested]
 
   /// \brief Creates a string view for the range from pStart to pStart + uiLength.
-  ezStringView(const char* pStart, ezUInt32 uiLength);
+  constexpr ezStringView(const char* pStart, ezUInt32 uiLength);
 
   /// \brief Advances the start to the next character, unless the end of the range was reached.
   void operator++(); // [tested]
@@ -113,5 +113,15 @@ private:
   const char* m_pStart = nullptr;
   const char* m_pEnd = nullptr;
 };
+
+namespace ezLiterals
+{
+  /// \brief String literal suffix to create a ezStringView.
+  ///
+  /// Example:
+  /// using namespace ezLiterals;
+  /// "Hello World"_sv
+  constexpr ezStringView operator "" _sv(const char* pString, size_t len);
+}
 
 #include <Foundation/Strings/Implementation/StringView_inl.h>
--- a/Code/Engine/Foundation/Strings/StringView.h
+++ b/Code/Engine/Foundation/Strings/StringView.h
@@ -121,7 +121,7 @@ namespace ezLiterals
   /// Example:
   /// using namespace ezLiterals;
   /// "Hello World"_sv
-  constexpr ezStringView operator "" _sv(const char* pString, size_t len);
-}
+  constexpr ezStringView operator"" _sv(const char* pString, size_t len);
+} // namespace ezLiterals
 
 #include <Foundation/Strings/Implementation/StringView_inl.h>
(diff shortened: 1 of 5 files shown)


jankrassnigg:
Is the namespsce really necessary?

jankrassnigg:
One thing I was wondering, when we create a string view, we usually know whether it is zero terminated. So we could store that flag in the view and when we call GetData on it, we could omit the copy when it's already terminated. Unfortunately we can't read the value at the end pointer, because that could already be an access violation.

Ingrater:
> Is the namespsce really necessary?

If we don't have a namespace, we pollute the global namespace with a literal suffix called "_sv" which can easily collide with other c++ libraries / user code. So yes, I would say it is neccessary.

If we don't want the namespace we would have to name the literal `_ezsv` or something like that, which is more typing work. All user defined literal suffixes have to start with '_'

The two options essentially are:
1) Type `using namespace ezLiterals;` once at the start of the file and then use the short `_sv` everywhere.
2) Always type the longer `_ezsv`

Ingrater:
> One thing I was wondering, when we create a string view, we usually know whether it is zero terminated. So we could store that flag in the view and when we call GetData on it, we could omit the copy when it's already terminated. Unfortunately we can't read the value at the end pointer, because that could already be an access violation.

That would be another possible optimization. Let's not do that in this PR as well. I'm currently more worried with the 1000ths of times where we know the string length, throw it away, just to then count the number of elements in the next function again.

jankrassnigg (approved):

Code/Engine/Foundation/Strings/StringView.h
```diff
@@ -114,4 +114,14 @@ struct EZ_FOUNDATION_DLL ezStringView : public ezStringBase<ezStringView>
   const char* m_pEnd = nullptr;
 };
 
+namespace ezLiterals
```
jankrassnigg: I'd prefer not to have a namespace for this, because it is inconvenient to write "using ..." everywhere to enable the feature. And we can't put "using ezLiterals" into our headers either, because then you have the name clash problem again. I think having a global "_ezsv" (or just "_ezs" ? ) would be better.

Code/Engine/Foundation/Strings/StringView.h
```diff
@@ -25,13 +25,13 @@ struct EZ_FOUNDATION_DLL ezStringView : public ezStringBase<ezStringView>
   ezStringView();
 
   /// \brief Creates a string view starting at the given position, ending at the next '\0' terminator.
-  ezStringView(const char* pStart); // [tested]
+  constexpr ezStringView(const char* pStart); // [tested]
 
   /// \brief Creates a string view for the range from pStart to pEnd.
   ezStringView(const char* pStart, const char* pEnd); // [tested]
```
jankrassnigg, line 31: I guess making this one constexpr isn't necessary?

Code/Engine/Foundation/Strings/StringView.h
Ingrater: We can put `using namespace ezLiterals` into our headers. `using namespace` statements can not only appear in the global scope, they can also appear within functions / methods. So for header files the using statement would be done inside the function / method that needs it.

As we are not using namespaces anywhere else in ez, I'm going to move to "_ezsv" instead. I don't want to use "_ezs" because I want to keep that for string literals (e.g. producing a `ezString`). To stay in line with STL which uses "sv" as suffix for string views I also want to stick with "_ezsv".

Code/Engine/Foundation/Strings/StringView.h
Ingrater: It is neccessary. Otherwise `constexpr auto someConstant = "Hello World"_ezsv;` doesn't work.

PR #669, opened 2022-08-30
Removed redundand const char* -> ezStringView conversions in ezFoundation

Ingrater:
Also move a few Foundation APIs to ezStringView

### DIFF
--- a/Code/Engine/Core/Assets/AssetFileHeader.h
+++ b/Code/Engine/Core/Assets/AssetFileHeader.h
@@ -36,7 +36,7 @@ public:
   const ezHashedString& GetGenerator() { return m_sGenerator; }
 
   /// \brief Allows to set the generator string
-  void SetGenerator(const char* szGenerator) { m_sGenerator.Assign(szGenerator); }
+  void SetGenerator(ezStringView szGenerator) { m_sGenerator.Assign(szGenerator); }
 
 private:
   ezUInt64 m_uiHash;
--- a/Code/Engine/Core/Console/Implementation/LuaInterpreter.cpp
+++ b/Code/Engine/Core/Console/Implementation/LuaInterpreter.cpp
@@ -160,7 +160,7 @@ void ezCommandInterpreterLua::Interpret(ezCommandInterpreterState& inout_State)
   }
 
   sTemp = "> ";
-  sTemp.Append(sRealCommand.GetData());
+  sTemp.Append(sRealCommand);
   inout_State.AddOutputLine(sTemp, ezConsoleString::Type::Executed);
 
   ezCVar* pCVAR = ezCVar::FindCVarByName(sRealVarName.GetData());
--- a/Code/Engine/Core/World/Implementation/WorldData.h
+++ b/Code/Engine/Core/World/Implementation/WorldData.h
@@ -154,7 +154,7 @@ namespace ezInternal
 
     struct InitBatch
     {
-      InitBatch(ezAllocatorBase* pAllocator, const char* szName, bool bMustFinishWithinOneFrame);
+      InitBatch(ezAllocatorBase* pAllocator, ezStringView szName, bool bMustFinishWithinOneFrame);
 
       ezHashedString m_sName;
       bool m_bMustFinishWithinOneFrame = true;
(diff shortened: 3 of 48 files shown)


Code/Engine/Foundation/Basics/Platform/Win/HResultUtils.cpp
```diff
@@ -15,8 +15,8 @@ EZ_FOUNDATION_DLL ezString ezHRESULTtoString(ezMinWindows::HRESULT result)
 
   // Com error tends to put /r/n at the end. Remove it.
   ezStringBuilder message(ezStringUtf8(messageW).GetData());
-  message.ReplaceAll("\n", "");
-  message.ReplaceAll("\r", "");
+  message.ReplaceAll("\n"_ezsv, ""_ezsv);
```
jankrassnigg: I don't really get why the constructor has to be explicit. What do I gain from this? I now have to mark EVERY string, which I find quite annoying, and I don't see that it has any benefit.

I agree with using ezStringView in parameters over const char*, but frankly I disagree with using the string literal thing everywhere.

Code/Engine/Core/World/Implementation/GameObject_inl.h
```diff
@@ -99,7 +99,7 @@ EZ_ALWAYS_INLINE void ezGameObject::SetName(const ezHashedString& sName)
 EZ_ALWAYS_INLINE void ezGameObject::SetGlobalKey(const char* szGlobalKey)
 {
   ezHashedString sGlobalKey;
-  sGlobalKey.Assign(szGlobalKey);
+  sGlobalKey.Assign(ezStringView(szGlobalKey));
```
jankrassnigg: Currently you are adding lots of unnecessary casts. Once you touch Core, szGlobalKey would also change to an ezStringView and then you just have noise left. I think you should only compile Foundation, fix things there, then disable the explicit constructor again and then make sure that the rest still compiles. And revert any such redundant casts.

jankrassnigg (changes requested):
In my opinion the only place where one should use _ezsv is when you need to explicitly tell the compiler to use a string view variant, rather than a const char* variant. So when you need a cast to disambiguate. All other places are just visual clutter.

Code/Engine/Foundation/Basics/Platform/Win/HResultUtils.cpp
Ingrater: I just want to make this clear:
Because we moved ReplaceAll (and other stirng methods) form const char* to ezStringView. The generated code at the call site is now absolutely horrible.
`message.ReplaceAll("\n", "")` is equivalent to `message.ReplaceAll(ezStringView("\n"), ezStringView(""))` which does a string length computation for each of the parameters. If you look at the disassembly it looks horrible.

By using _ezsv all this string length computation goes away, and the it esentially boils down to pointer + length which are both known at compile time. As the length is now compile time known, it allows the compiler to inline / eliminate much more stuff (such as length == 0 checks).

Code/Engine/Core/World/Implementation/GameObject_inl.h
Ingrater: These casts are not unnecessary. They are necessary and they are exactly what the compiler would do implicitly without the explicit constructor. Just in this case you can clearly see what is going on (because the ezStringView constructor is called explicitly). I clearly prefer this version because you can see from it what is going on (length counting) and you can even grep for these occurances to find places for improvement.

Ingrater:
> In my opinion the only place where one should use _ezsv is when you need to explicitly tell the compiler to use a string view variant, rather than a const char* variant. So when you need a cast to disambiguate. All other places are just visual clutter.

If we want performant code we need to tell the compiler **everywhere** to use ezStringView. If we rely on the implicit `const char*` -> ezStringView conversion very bad code is generated.

Code/Engine/Foundation/Basics/Platform/Win/HResultUtils.cpp
jankrassnigg: IIRC you mentioned checking the assembly with compiler explorer. Do you have a link to try this out?
Do you know whether there are differences in code generation between VS19 and VS20 ?
Would adding an always-inline to the constructor make a difference? Can we wrap it with pragmas to enforce optimization?

If I understood this correctly, the bad code generation mostly happens in Debug builds with MSVC, right?

If the compiler generates awful code, I would consider this a compiler bug. There is certainly some stuff that I'd be willing to do to guarantee best performance, but littering the entire code base with something that is pretty just telling the compiler to please not be so bad at its job, is going a bit far for my taste. So I wonder whether there are options that we have missed.

jankrassnigg (approved):

PR #670, opened 2022-08-30
Sliders for int values with limited ranges

jankrassnigg:
Integer properties whose range is below 256 get an additional slider next to the spinbox to adjust the value.

### DIFF
--- a/Code/Tools/Libs/GuiFoundation/PropertyGrid/Implementation/PropertyWidget.cpp
+++ b/Code/Tools/Libs/GuiFoundation/PropertyGrid/Implementation/PropertyWidget.cpp
@@ -559,7 +559,7 @@ void ezQtPropertyEditorIntSpinboxWidget::OnInit()
         m_pWidget[0]->setMinimum(pClamp->GetMinValue());
         m_pWidget[0]->setMaximum(pClamp->GetMaxValue());
 
-        if ((iMaxValue - iMinValue) < 256)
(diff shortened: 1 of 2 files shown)


Code/Tools/Libs/GuiFoundation/PropertyGrid/Implementation/PropertyWidget.cpp
```diff
@@ -552,9 +552,29 @@ void ezQtPropertyEditorIntSpinboxWidget::OnInit()
...
       case 1:
       {
+        const ezInt32 iMinValue = pClamp->GetMinValue().ConvertTo<ezInt32>();
+        const ezInt32 iMaxValue = pClamp->GetMaxValue().ConvertTo<ezInt32>();
+
         ezQtScopedBlockSignals bs(m_pWidget[0]);
         m_pWidget[0]->setMinimum(pClamp->GetMinValue());
         m_pWidget[0]->setMaximum(pClamp->GetMaxValue());
+
+        if ((iMaxValue - iMinValue) < 256)
```
Sanakan8472: I'd change this to
```
if (pClamp->GetMinValue().IsValid() && pClamp->GetMaxValue().IsValid())
```
`ConvertTo<ezInt32>()` will return 0 if the value can't be converted.

Sanakan8472 (changes requested):

Code/Tools/Libs/GuiFoundation/PropertyGrid/Implementation/PropertyWidget.cpp
jankrassnigg: Fixed, thanks!

PR #671, opened 2022-08-30
Elaborate logic for ezStringView construction

Ingrater:
Automatically pick the optimal way to construct a ezStringView. String literals are now detected and a string view is constructed optimally from them.

This is the alternative to manually adding the suffix _ezsv to every string literal.

### DIFF
--- a/Code/Engine/Foundation/Basics/Compiler/Clang/Clang.h
+++ b/Code/Engine/Foundation/Basics/Compiler/Clang/Clang.h
@@ -6,9 +6,12 @@
 #  undef EZ_COMPILER_CLANG
 #  define EZ_COMPILER_CLANG EZ_ON
 
-/// \todo re-investigate: attribute(always inline) does not work for some reason
-#  define EZ_ALWAYS_INLINE inline
-#  define EZ_FORCE_INLINE inline
+#  define EZ_ALWAYS_INLINE __attribute__((always_inline)) inline
+#  if EZ_ENABLED(EZ_COMPILE_FOR_DEBUG)
+#    define EZ_FORCE_INLINE inline
+#  else
+#    define EZ_FORCE_INLINE __attribute__((always_inline)) inline
+#  endif
 
 #  define EZ_ALIGNMENT_OF(type) EZ_COMPILE_TIME_MAX(__alignof(type), EZ_ALIGNMENT_MINIMUM)
 
--- a/Code/Engine/Foundation/Strings/Implementation/StringView_inl.h
+++ b/Code/Engine/Foundation/Strings/Implementation/StringView_inl.h
@@ -2,10 +2,24 @@
 
 inline ezStringView::ezStringView() = default;
 
-constexpr inline ezStringView::ezStringView(const char* pStart)
+inline ezStringView::ezStringView(char* pStart)
   : m_pStart(pStart)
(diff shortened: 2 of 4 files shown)


jankrassnigg:
If this solves the problem, I like this a lot better. Even though I don't understand how this works 😂
Could you elaborate how this tricks the compiler to generate better code?

jankrassnigg (approved):

Ingrater:
> If this solves the problem, I like this a lot better. Even though I don't understand how this works 😂 Could you elaborate how this tricks the compiler to generate better code?

The bad code gen is caused because the compiler doesn't know the length of the string literal at compile time when it takes the `ezStringView(const char*)` constructor. So it counts the bytes on each construction of a ezStringView. We have to fix this, so we introduce the `template <size_t N> ezStringView(const char(&str)[N])` constructor. String literals convert to `const char[N]` and to `const char*`. But unfortuantely adding this constructor is not enough. As there is also a `const char*` constructor, this one will win in overload resolution, because non templated methods are preferred over templated ones. This means we have to make the `const char*` constructor templated.
```
template<typename T>
ezStringView(T pStart)
```
Now we don't want this constructor to accept any type, we just want it to accept `const char*`
So we need to use std::enable_if. This takes a condition as first argument and a type as second argument. Only if the condition is true does the `::type` member exist. Otherwise the `::type` member doesnt exist and this would lead to a compile error. Due to SFINAE during overload resolution, the compile error just leads to this overload being ignored. Typically std::enable_if is used on the return value of the method you want to enable. As a constructor does not have a return value, we have to add a second argument with default value:
```template <typename T>
ezStringView(T pStart, typename std::enable_if<std::is_same<T, const char*>::value, int>* = 0)
```
If `T == const char*` we end up with
```
ezStringView(const char* pStart, int* = 0)
```
The additional argument is unused and optimized away.
If `T != const char*` a compile error would happen thus the entire constructor is removed from overload resolution.

The next problem we run into are types which are implicitly convertible to `const char*`. Typically these have an `operator const char* () const;` Due to our template `const char*` constructor this does no longer work out of the box and we have to handle it manually. So we have to check i fT is convertible to `const char*`. In this case we don't want to take pStart as value, as it could be a full blown container we don't want to deep copy.
```template <typename T>
ezStringView(const T&& pStart, typename std::enable_if<std::is_convertible<T, const char*>::value, int>* = 0)
```

Unfortunately this constructor now adds ambigiutiy. `const char*` could now go both into our first templated constructor and in the second one, because `const char*` is convertible to `const char *`. So we need to make the condition a bit more complicated:
```template <typename T>
ezStringView(const T&& pStart, typename std::enable_if<std::is_same<T, const char*>::value == false && std::is_convertible<T, const char*>::value, int>* = 0)
```

Now we will notice that constructing a ezStringView from a `char*` no longer works. So we just add a constructor for that:
`ezStringView(char* pStart)`

And finally someone tries to construct a ezStringView from a char[2048] array, which also doesn't work at the moment. So we add another constructor `template<size_t N> ezStringView(char(&str)[N]`

After having done all this, all previous use cases for constructing a ezStringView work again but now a string literal will go down the `ezStringView(const char(&str)[N])` constructor, thus the string length is known at compile time and the compiler can generate good code.

jankrassnigg:
Mind blown! 🤣

PR #676, opened 2022-09-03
Added an in-place curve property

jankrassnigg:
This allows to embed curves directly into an object and removes the need to reference a curve asset.
For many use cases this is more convenient than to always create a new curve asset.
The curve is still stored in the same representation, so usually it should be converted to a more suitable runtime representation.


TheophilusE:
Very good 👍

PR #679, opened 2022-09-05
Vulkan Polishing

Sanakan8472:
## Staging textures
* Staging textures are now normal ezGALTextureVulkan with a bLinearCPU flag added to the ctor. Buffers have a similiar flag to force them to be in CPU readable memory.
* Staging textures can now be layout tracked like any other texture.

## Swap chain format fallback
* If a surface format is not available the format is just replaced with a different one for the textures. The format is no longer overridden and the engine code needs to check after creation what the actual format is.

## Texture readback
* There are now three readback modes defined in ezGALTextureVulkan::StagingMode: Buffer, Texture and TextureAndBuffer. Buffer is the preferred option. TextureAndBuffer is only needed if a format conversion needs to be done.
* ezImageCopyVulkan has been greatly simplified since the proxy texture is now handled explicitly via ezGALTextureVulkan::StagingMode.
* Readback ezImage format is no longer hardcoded but derived from the texture format via the new function ezTextureUtils::GalFormatToImageFormat.
* ezRenderContext::LoadBuiltinShader now support platforms that cannot write to array textures.
* Added ezGALCommandEncoderImplVulkan::CopyImageToBuffer. This uses the new ezGALTextureVulkan::ComputeSubResourceOffsets function to compute buffer offsets.
* For depth textures a non-depth texture format is now selected for the read-back during device creation.

## Temporary Workarounds
* ezGALShaderVulkan store a name field for each BindingMapping. This is used during ezGALCommandEncoderImplVulkan::FlushDeferredStateChanges to decide whether a fallback texture needs to be depth or color. The problem here is that a color texture can't be depth sampled and a depth texture can't be bilinear sampled on some hardware, making each texture type fail validation in some cases. We can't extract this from spirv-reflect as the data is unreliable. Whatever the validation layer does to detect this would need to be replicated in spirv-reflect or we need to fix this in EZ and annotate slots we assume should be auto-filled if not set.

## Unit tests
* All unit tests no pass in lavapipe. The only big difference is that particle tests always sample the wrong mip level, resulting in blurrier results. To compensate ezGameEngineTestParticles::GetImageCompareThreshold was added.
* Custom reference image for lavapipe (Images_Reference_LLVMPIPE) due to differetn line rendering.

## Misc changes
* Fixed a debug assert in the FileReader


Code/Engine/RendererVulkan/Device/Implementation/SwapChainVulkan.cpp
```diff
@@ -34,6 +34,23 @@ namespace
...
   }
+
+  ezGALResourceFormat::Enum GetResourceFormat(vk::Format& format)
+  {
+    switch (format)
+    {
+      case vk::Format::eR8G8B8A8Srgb:
+        return ezGALResourceFormat::RGBAUByteNormalizedsRGB;
+      case vk::Format::eR8G8B8A8Unorm:
+        ezGALResourceFormat::RGBAUByteNormalized;
```
Ingrater: Missing `return`?

Code/Engine/RendererVulkan/Device/Implementation/SwapChainVulkan.cpp
```diff
@@ -34,6 +34,23 @@ namespace
...
+      case vk::Format::eB8G8R8A8Srgb:
+        return ezGALResourceFormat::BGRAUByteNormalizedsRGB;
+      case vk::Format::eB8G8R8A8Unorm:
+        ezGALResourceFormat::BGRAUByteNormalized;
```
Ingrater: Missing `return`?

Code/Engine/RendererVulkan/Resources/Implementation/BufferVulkan.cpp
```diff
@@ -190,11 +198,11 @@ void ezGALBufferVulkan::SetDebugNamePlatform(const char* szName) const
   m_pDeviceVulkan->SetDebugName(szName, m_currentBuffer.m_buffer, m_currentBuffer.m_alloc);
 }
 
-vk::DeviceSize ezGALBufferVulkan::GetAlignment(const ezGALDeviceVulkan* pDevice, vk::BufferUsageFlags usage) const
+vk::DeviceSize ezGALBufferVulkan::GetAlignment(const ezGALDeviceVulkan* pDevice, vk::BufferUsageFlags usage)
 {
   const vk::PhysicalDeviceProperties& properties = pDevice->GetPhysicalDeviceProperties();
 
-  vk::DeviceSize alignment = 4;
+  vk::DeviceSize alignment = ezMath::Max(4ull, properties.limits.nonCoherentAtomSize);
```
Ingrater: `ezMath::Max<vk::DeviceSize>(4, properties.limits.nonCoherentAtomSize);` to make it cross-platform vk::DeviceSize is `unsigned long int` on linux but `unsigned long long int` on windows. So so `ull` only works on windows.

Ingrater (approved):
Runs fine on my nvidia card on linux. No more validation errors.

Code/Engine/RendererVulkan/Device/Implementation/SwapChainVulkan.cpp
Sanakan8472: Ups! Thanks for noticing!

Code/Engine/RendererVulkan/Resources/Implementation/BufferVulkan.cpp
Sanakan8472: Oh didn't know that. Fixed.

PR #683, opened 2022-09-12
Save/Load curve presets and better easing curves generation

jankrassnigg:
The curve editor now allows to save curves as presets.
All curve presets that are located under "Editor/Presets/Curves" (in any data directory) are automatically available from the context menu. From other places they can be opened manually.
Currently there are no presets shipped, but that may change in the future.

Also the "easing" curves that can be generated from the context menu are now more optimized, containing only the control points that are necessary to decently describe the curve.


TheophilusE:
You seem to find more improvements and uses for this, that's great!

PR #684, opened 2022-09-12
Make re-runs of cmake faster

Ingrater:
* Once there is a existing cmake cache, reduce the runtime of cmake by 36% (down to 5.3 seconds from 8.3 seconds on my machine)
* This is possible thanks to the discovery of the cmake `--profiling-output=perf.json --profiling-format=google-trace` command line options, which give you a json file that can be opened in `chrome://tracing`

### DIFF
--- a/Code/BuildSystem/CMake/CMakeUtils/ezUtilsUnityFiles.cmake
+++ b/Code/BuildSystem/CMake/CMakeUtils/ezUtilsUnityFiles.cmake
@@ -76,7 +76,6 @@ function(ez_generate_folder_unity_file TARGET_NAME PROJECT_DIRECTORY SUB_FOLDER_
 
 	if(EXISTS ${FOLDER_UNITY_FILE})
 		# check if the generated unity file is different from the existing one
-		#file(SHA1 ${FOLDER_UNITY_FILE} UNITY_EXISTING_HASH)
 		file(READ ${FOLDER_UNITY_FILE} OLD_UNITY_FILE_CONTENTS)
 		string(SHA1 UNITY_EXISTING_HASH ${OLD_UNITY_FILE_CONTENTS})
 		string(SHA1 UNITY_NEW_HASH ${UNITY_FILE_CONTENTS})
@@ -88,7 +87,6 @@ function(ez_generate_folder_unity_file TARGET_NAME PROJECT_DIRECTORY SUB_FOLDER_
 
 	if(${UNITY_FILE_NEEDS_UPDATE})
 		file(WRITE ${FOLDER_UNITY_FILE} ${UNITY_FILE_CONTENTS})
-		message("Writing unity file ${FOLDER_UNITY_FILE}")
 	endif()
 
 	# add the file to the target and sort it into the proper sub-tree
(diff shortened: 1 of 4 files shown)


Code/BuildSystem/CMake/CMakeUtils/ezUtilsUnityFiles.cmake
```diff
@@ -63,26 +65,30 @@ function(ez_generate_folder_unity_file TARGET_NAME PROJECT_DIRECTORY SUB_FOLDER_
 		set(CUR_FILE_FULL "${PROJECT_DIRECTORY}/${SUB_FOLDER_PATH}/${CUR_FILE}")
 
 		file(TO_NATIVE_PATH ${CUR_FILE_FULL} CUR_FILE_NATIVE)
-		file(APPEND ${TMP_UNITY_FILE} "#include \"${CUR_FILE_NATIVE}\"\n")
+		string(APPEND UNITY_FILE_CONTENTS "#include \"${CUR_FILE_NATIVE}\"\n")
 
 		set_source_files_properties(${CUR_FILE_FULL} PROPERTIES HEADER_FILE_ONLY true)
 	endforeach()
+
```
jankrassnigg, line 72: Shouldn't this write the file, rather than append?

Code/BuildSystem/CMake/CMakeUtils/ezUtilsUnityFiles.cmake
```diff
@@ -63,26 +65,30 @@ function(ez_generate_folder_unity_file TARGET_NAME PROJECT_DIRECTORY SUB_FOLDER_
...
+		string(SHA1 UNITY_EXISTING_HASH ${OLD_UNITY_FILE_CONTENTS})
+		string(SHA1 UNITY_NEW_HASH ${UNITY_FILE_CONTENTS})
 
 		if(${UNITY_EXISTING_HASH} STREQUAL ${UNITY_NEW_HASH})
 			set(UNITY_FILE_NEEDS_UPDATE false)
 		endif()
 	endif()
 
 	if(${UNITY_FILE_NEEDS_UPDATE})
-		file(READ ${TMP_UNITY_FILE} UNITY_FILE_CONTENTS)
```
jankrassnigg, line 84: Is this message really useful? I would expect it to just be spam.

jankrassnigg (approved):

Code/BuildSystem/CMake/CMakeUtils/ezUtilsUnityFiles.cmake
Ingrater: Good catch, this was just for debugging.

Code/BuildSystem/CMake/CMakeUtils/ezUtilsUnityFiles.cmake
Ingrater: No. The idea behind this optimization is that this cmake scripts constructs a string that contains the entire new content of the unity cpp file. Then if the sha1 check shows that a modification happend, the file is written to disk. This reduces and amount of file operations cmake does, and speeds up the unity cpp generation.

jankrassnigg: Yes, I get that part. But at least on my phone I didn't see that you ever write the file, ie replace it, rather you seem to only append, which may work if the file doesn't exist, but not if it does. So that was my question.

PR #696, opened 2022-09-25
Fix two inalid memory access issues in vulkan renderer.

Ingrater:
* Issue 1: When computing the hash of constant buffer data, not all data might be written to by the user of the constant buffer. As a result uninitialized bytes are added to the hash.
* Issue 2: Use of the memory held by a HybridArray even though the hybrid array is out of scope.


Code/Engine/RendererCore/Shader/Implementation/ConstantBufferStorage.cpp
```diff
@@ -9,6 +9,7 @@ ezConstantBufferStorageBase::ezConstantBufferStorageBase(ezUInt32 uiSizeInBytes)
   , m_uiLastHash(0)
 {
   m_Data = ezMakeArrayPtr(static_cast<ezUInt8*>(ezFoundation::GetAlignedAllocator()->Allocate(uiSizeInBytes, 16)), uiSizeInBytes);
+  memset(m_Data.GetPtr(), 0, m_Data.GetCount());
```
jankrassnigg: Ezmemoryutils::zeromemory

jankrassnigg (approved):

PR #703, opened 2022-09-29
Use sharedptr for blackboards

C-Core:
Blackboards are used to share data between multiple subsystems, so their ownership needs to be shared as well to prevent dangling pointer access.

### DIFF
--- a/Code/Engine/GameEngine/Gameplay/BlackboardComponent.h
+++ b/Code/Engine/GameEngine/Gameplay/BlackboardComponent.h
@@ -69,11 +69,11 @@ public:
   ezBlackboardComponent& operator=(ezBlackboardComponent&& other);
 
   /// \brief Try to find a ezBlackboardComponent on pSearchObject or its parents and returns its blackboard
-  static ezBlackboard* FindBlackboard(ezGameObject* pSearchObject);
+  static ezSharedPtr<ezBlackboard> FindBlackboard(ezGameObject* pSearchObject);
 
   /// \brief Returns the blackboard owned by this component
-  ezBlackboard& GetBoard();
-  const ezBlackboard& GetBoard() const;
+  const ezSharedPtr<ezBlackboard>& GetBoard();
+  ezSharedPtr<const ezBlackboard> GetBoard() const;
 
   void SetShowDebugInfo(bool bShow); // [ property ]
   bool GetShowDebugInfo() const;     // [ property ]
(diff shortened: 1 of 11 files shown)


Code/Engine/GameEngine/Gameplay/BlackboardComponent.h
```diff
@@ -69,11 +69,11 @@ class EZ_GAMEENGINE_DLL ezBlackboardComponent : public ezComponent
   ezBlackboardComponent& operator=(ezBlackboardComponent&& other);
 
   /// \brief Try to find a ezBlackboardComponent on pSearchObject or its parents and returns its blackboard
-  static ezBlackboard* FindBlackboard(ezGameObject* pSearchObject);
+  static ezSharedPtr<ezBlackboard> FindBlackboard(ezGameObject* pSearchObject);
 
   /// \brief Returns the blackboard owned by this component
-  ezBlackboard& GetBoard();
-  const ezBlackboard& GetBoard() const;
+  const ezSharedPtr<ezBlackboard>& GetBoard();
```
Ingrater, line 75: I thougt the idea of ezRefCounted is that you can Pass around raw pointers and only use ezSharedPtr when storing a reference. Are we not leveraging this advantage?

Ingrater (approved):

jankrassnigg (approved):

Code/Engine/GameEngine/Gameplay/BlackboardComponent.h
jankrassnigg: I'm not sure whether we allow this implicitly, but we definitely never do this on purpose, because it hides from the user that something is shared. Shared objects should always be passed as sharedptrs when their ownership is meant to be shared with others, which is the case here. I would only pass the object around as a raw pointer, when the intent is to only read from it and when the lifetime is already guaranteed by someone else. Here we explicitly give out the pointer for others to extend the lifetime and therefore it is better to communicate that in an obvious way.

C-Core: No we can't do that since we also need the allocator which was used to allocate the refcounted object. 
Furthermore I don't see it as an advantage to pass around raw pointers since the user then doesn't know what the ownership model is. If you pass around a sharedptr, uniqueptr, handle etc. the ownership model is immediately visible to the user.

PR #708, opened 2022-10-01
Stop using QT features that have been deprecated in 5.15

Ingrater:
* Should make it easier to port to Qt 6.


jankrassnigg:
Are you sure all of this still compiles with the version we use? Isn't that 5.13?

Ingrater:
> Are you sure all of this still compiles with the version we use? Isn't that 5.13?

Didn't know that ^^. Linux is using 5.15, and the official recommendation for porting to Qt6 is, that you should remove deprecation warnings on 5.15 first. I think I'll keep this PR around so it can be re-used when we switch to Qt6.

jankrassnigg:
You can wrap the code with ifdef qt version 5.15.
There are extra qt macros for that. Then it works now and in the future.

jankrassnigg (approved):

PR #714, opened 2022-10-06
Fixes for warnings on Linux

jankrassnigg:


Code/Engine/Foundation/Reflection/Implementation/StaticRTTI.h
```diff
@@ -276,7 +282,7 @@ EZ_ALWAYS_INLINE const ezRTTI* ezGetStaticRTTI()
 
 
 // [internal] Helper macro to get the return type of a getter function.
-#define EZ_GETTER_TYPE(Class, GetterFunc) decltype(((Class*)nullptr)->GetterFunc())
+#define EZ_GETTER_TYPE(Class, GetterFunc) decltype(((Class*)nullptr + 123)->GetterFunc())
```
Ingrater: It would be better to use std::declval here or build our own equivalent of it.

Ingrater (changes requested):

Ingrater (approved):

PR #716, opened 2022-10-08
Upgrade to qt6

Ingrater:
This upates ezEngine to use Qt6

* Using Qt 6.2.4 on Linux (as this is the version that comes with the package manager in Ubuntu 22 / Linux Mint 21).
* Using Qt 6.3.2 on MacOS (fetched through brew)
* Using Qt 6.4.0 on Windows (new custom build binary package with only the stuff we need)

**Major gotcha:**
* `QWinTaskbarButton` and `QWinTaskbarProgress` have been removed without replacement. This means the editor will loose some of its current Windows integration.

_This PR will not address any of the deprecating warnings caused by switching to Qt6. The goal is to get everything compiling and working, deprecation warnings can be fixed later._


ezCMakeConfig.cmake
```diff
@@ -13,8 +13,8 @@ set_property(GLOBAL PROPERTY EZ_CMAKE_RELPATH_CODE "Code")
 
 set_property(GLOBAL PROPERTY EZ_CONFIG_PATH_7ZA "${CMAKE_SOURCE_DIR}/Data/Tools/Precompiled/7za.exe")
 
-set_property(GLOBAL PROPERTY EZ_CONFIG_QT_WINX64_VERSION "Qt-5.13.0-vs141-x64")
-set_property(GLOBAL PROPERTY EZ_CONFIG_QT_WINX64_URL "https://github.com/ezEngine/thirdparty/releases/download/Qt-5.13.0-vs141-x64/Qt-5.13.0-vs141-x64.zip")
+set_property(GLOBAL PROPERTY EZ_CONFIG_QT_WINX64_VERSION "Qt6-6.4.0-vs143-x64")
+set_property(GLOBAL PROPERTY EZ_CONFIG_QT_WINX64_URL "http://stuff.benjamin-thaut.de/ez/Qt6-6.4.0-vs143-x64.7z")
```
Ingrater: This is a temporary URL that has to be replaced with a github hosted package instead.

jankrassnigg:
Why no more task bar progress? That's the best feature!

jankrassnigg (approved):

Ingrater:
> Why no more task bar progress? That's the best feature!

The QT developers wanted to make a cross platform solution that works on all platforms, so they removed the windows specific one. Unfortunately the cross platform solution is not done yet ...

PR #717, opened 2022-10-08
Fix Shipping builds

na2axl:
Check if profiling is enabled before using profiling scope functions in the DX11 renderer.


jankrassnigg:
Looks good, thank you!

jankrassnigg (approved):

PR #720, opened 2022-10-10
Force clean build support

Sanakan8472:
- Change gitClean.txt to force a clean build on CI.

### DIFF
--- a/Code/BuildSystem/AzurePipelines/Windows-x64.yml
+++ b/Code/BuildSystem/AzurePipelines/Windows-x64.yml
@@ -40,6 +40,20 @@ jobs:
     submodules: true
     lfs: true
     clean: false
+  - task: PowerShell@2
+    displayName: Check force clean
+    inputs:
+      targetType: inline
+      script: |
+        $pathA = Join-Path $(System.DefaultWorkingDirectory) "gitClean.txt"
+        $pathB = Join-Path $(System.DefaultWorkingDirectory) "gitCleanCopy.txt"
+        $clean = -not (Test-Path $pathB) -or (compare-object (get-content $pathA) (get-content $pathB))
+        if ($clean)
+        {
+            Write-Host "Cleaning repo!"
+            git clean -dfx
+            Copy-Item -Path $pathA -Destination $pathB
+        }
   - task: AzureKeyVault@1
     displayName: 'Azure Key Vault: ezKeys'
     inputs:
--- a/Code/EditorPlugins/Assets/EditorPluginAssets/MaterialAsset/ShaderTypeRegistry.cpp
+++ b/Code/EditorPlugins/Assets/EditorPluginAssets/MaterialAsset/ShaderTypeRegistry.cpp
@@ -294,7 +294,7 @@ const ezRTTI* ezShaderTypeRegistry::GetShaderType(const char* szShaderPath)
     {
       if (!ezQtEditorApp::GetSingleton()->MakeDataDirectoryRelativePathAbsolute(sAbsPath))
       {
-        ezLog::Error("Can't make path absolute: '{0}'", szShaderPath);
+        ezLog::Warning("Can't make path absolute: '{0}'", szShaderPath);
         return nullptr;
       }
       sAbsPath.MakeCleanPath();
--- a/Code/UnitTests/EditorTest/Misc/Misc.cpp
+++ b/Code/UnitTests/EditorTest/Misc/Misc.cpp
@@ -31,6 +31,12 @@ ezResult ezEditorTestMisc::InitializeTest()
   if (SUPER::OpenProject("Data/UnitTests/EditorTest").Failed())
     return EZ_FAILURE;
 
+  if (ezStatus res = ezAssetCurator::GetSingleton()->TransformAllAssets(ezTransformFlags::None); res.Failed())
+  {
+    ezLog::Error("Asset transform failed: {}", res.m_sMessage);
+    return EZ_FAILURE;
+  }
+
   return EZ_SUCCESS;
 }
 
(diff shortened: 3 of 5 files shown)


Code/BuildSystem/AzurePipelines/Windows-x64.yml
```diff
@@ -90,9 +104,9 @@ jobs:
     inputs:
       cmakeArgs: -DCMAKE_PREFIX_PATH=$(System.DefaultWorkingDirectory)\ThirdPartyWinx64\Qtvs141x64; -DEZ_QT_DIR="$(System.DefaultWorkingDirectory)\ThirdPartyWinx64\Qtvs141x64" -DEZ_ENABLE_FOLDER_UNITY_FILES=0 -DEZ_BUILD_OPENXR=1 -DEZ_BUILD_PHYSX=1 -DEZ_BUILD_FMOD=1 -DEZ_BUILD_GAMES=1 -DEZ_COMPILE_ENGINE_AS_DLL=1 -DEZ_ENABLE_QT_SUPPORT=1 -DCMAKE_SYSTEM_VERSION=10.0.18362.0 -DCMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION=10.0.18362.0 -A "x64" -G "Visual Studio 16 2019" ../
   - task: MSBuild@1
-    displayName: Build solution build/ezVs2019.sln
+    displayName: Build solution build/s_build.sln
```
jankrassnigg: Where does this name "s_build" come from? I don't see that we are setting the solution name anywhere.

Code/BuildSystem/AzurePipelines/Windows-x64.yml
```diff
@@ -90,9 +104,9 @@ jobs:
     inputs:
       cmakeArgs: -DCMAKE_PREFIX_PATH=$(System.DefaultWorkingDirectory)\ThirdPartyWinx64\Qtvs141x64; -DEZ_QT_DIR="$(System.DefaultWorkingDirectory)\ThirdPartyWinx64\Qtvs141x64" -DEZ_ENABLE_FOLDER_UNITY_FILES=0 -DEZ_BUILD_OPENXR=1 -DEZ_BUILD_PHYSX=1 -DEZ_BUILD_FMOD=1 -DEZ_BUILD_GAMES=1 -DEZ_COMPILE_ENGINE_AS_DLL=1 -DEZ_ENABLE_QT_SUPPORT=1 -DCMAKE_SYSTEM_VERSION=10.0.18362.0 -DCMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION=10.0.18362.0 -A "x64" -G "Visual Studio 16 2019" ../
```
jankrassnigg: -DEZ_BUILD_PHYSX=1 -DEZ_BUILD_FMOD=1 -DEZ_BUILD_GAMES=1

They all should be removed.
PhysX will be removed at some point and the default switches to Jolt. One of them will always be on.
Fmod is now on by default anyway.
The GAMES flag doesn't exist anymore.

Code/BuildSystem/AzurePipelines/Windows-x64.yml
Sanakan8472: This name is being auto-generated this way. Normally the root folder of the checkout is called ezEngine so the sln is called ezEngine_build.sln. However, on CI the checkout folder is callsed 's'. So s_build.sln it is.
This is cached so only found this out once I cleaned the build.

Code/BuildSystem/AzurePipelines/Windows-x64.yml
Sanakan8472: Do I need to add a flag for jolt?

Code/BuildSystem/AzurePipelines/Windows-x64.yml
```diff
@@ -40,6 +40,20 @@ jobs:
...
+    inputs:
+      targetType: inline
+      script: |
+        $pathA = Join-Path $(System.DefaultWorkingDirectory) "gitClean.txt"
+        $pathB = Join-Path $(System.DefaultWorkingDirectory) "gitCleanCopy.txt"
+        $clean = -not (Test-Path $pathB) -or (compare-object (get-content $pathA) (get-content $pathB))
+        if ($clean)
+        {
+            Write-Host "Cleaning repo!"
+            git clean -dfx
```
Ingrater, line 54: Doesn't `git clean` remove the gitCleanCopy.txt?

Ingrater (approved):

Code/BuildSystem/AzurePipelines/Windows-x64.yml
Sanakan8472: Yes of course, that's why the copy is done after the clean call now. We only need to detect that the file has changed. Once we know that we can purge everything and copy the checked-in file to indicate a cleaned state.

Code/BuildSystem/AzurePipelines/Windows-x64.yml
jankrassnigg: Use -DEZ_SOLUTION_NAME:string="solution" that will hardcode the solution name and you don't need to wonder what the folder name is. It's what I used in other github actions.

Code/BuildSystem/AzurePipelines/Windows-x64.yml
jankrassnigg: Nope, just remove the stuff and once I switch the default I can run a clean on CI to make it switch there as well :)

Code/BuildSystem/AzurePipelines/Windows-x64.yml
jankrassnigg: You PR actually removed the cache on CI and made my PR fail to build as well, so I had to make this change there as well. You'll probably get a conflict because of this.

PR #727, opened 2022-10-11
Node editor connections as "real" document objects

C-Core:
Node editor connections are now document objects just like the nodes itself. This allows them to have properties as well.

- File structure has been changed slightly: Connection objects are now self contained and not directly referenced by nodes anymore. There is backwards compatibility code though.
- Moved tons of duplicated code in the assets that used nodes into the node manager
- General cleanups


Code/Tools/Libs/ToolsFoundation/NodeObject/DocumentNodeManager.h
```diff
@@ -24,31 +23,38 @@ struct EZ_TOOLSFOUNDATION_DLL ezDocumentNodeManagerEvent
...
     , m_pObject(pObject)
-    , m_pConnection(pConnection)
+  //, m_Connection(connection)
   {
   }
 
   Type m_EventType;
   const ezDocumentObject* m_pObject;
-  const ezConnection* m_pConnection;
+  // const ezConnection* m_pConnection;
```
jankrassnigg: cleanup commented out code

Code/Engine/Core/Utils/Blackboard.h
```diff
@@ -163,3 +151,42 @@ class EZ_CORE_DLL ezBlackboard : public ezRefCounted
...
+  };
+
+  static bool Compare(ezComparisonOperator::Enum cmp, double f1, double f2);
+};
+
+EZ_DECLARE_REFLECTABLE_TYPE(EZ_CORE_DLL, ezComparisonOperator);
+
+//////////////////////////////////////////////////////////////////////////
+
+struct EZ_CORE_DLL ezBlackboardCondition
```
jankrassnigg: Is this supposed to be part of this PR ?

Code/Engine/Core/Utils/Implementation/Blackboard.cpp
```diff
@@ -179,5 +179,67 @@ ezResult ezBlackboard::Deserialize(ezStreamReader& stream)
...
+// clang-format off
+EZ_BEGIN_STATIC_REFLECTED_ENUM(ezComparisonOperator, 1)
+  EZ_ENUM_CONSTANTS(ezComparisonOperator::Equal, ezComparisonOperator::NotEqual)
+  EZ_ENUM_CONSTANTS(ezComparisonOperator::Less, ezComparisonOperator::LessEqual)
+  EZ_ENUM_CONSTANTS(ezComparisonOperator::Greater, ezComparisonOperator::GreaterEqual)
+EZ_END_STATIC_REFLECTED_ENUM;
+// clang-format on
+
+// static
+bool ezComparisonOperator::Compare(ezComparisonOperator::Enum cmp, double f1, double f2)
```
jankrassnigg: Don't we already have this somewhere else? And if not, should this be in the Blackboard, or maybe somewhere more general?

Code/Engine/Foundation/Types/Implementation/UniquePtr_inl.h
```diff
@@ -214,20 +214,44 @@ EZ_ALWAYS_INLINE bool operator==(const ezUniquePtr<T>& lhs, const T* rhs)
   return lhs.Borrow() == rhs;
 }
 
+template <typename T>
+EZ_ALWAYS_INLINE bool operator==(const ezUniquePtr<T>& lhs, T* rhs)
```
jankrassnigg, line 218: Do these operators make sense? Wouldn't they always return false anyway, unless you actually compare a unique ptr with itself ??

Code/Engine/RendererCore/AnimationSystem/AnimGraph/AnimGraphNode.h
```diff
@@ -59,26 +59,6 @@ class EZ_RENDERERCORE_DLL ezAnimGraphNode : public ezReflectedClass
 //////////////////////////////////////////////////////////////////////////
 //////////////////////////////////////////////////////////////////////////
 
-struct ezComparisonOperator
```
jankrassnigg: Ok I see you moved this from here to the blackboard, but maybe there is an even better place.

Code/Tools/Libs/GuiFoundation/NodeEditor/Implementation/Pin.cpp
```diff
@@ -38,7 +39,9 @@ void ezQtPin::RemoveConnection(ezQtConnection* pConnection)
   if (m_Connections.IsEmpty())
     ConnectedStateChanged(false);
 
-  UpdateConnections();
+  // TODO: is this really necessary?
```
jankrassnigg: cleanup

Code/Tools/Libs/GuiFoundation/NodeEditor/NodeScene.moc.h
```diff
@@ -35,6 +35,28 @@ class EZ_GUIFOUNDATION_DLL ezQtNodeScene : public QGraphicsScene
   static ezRttiMappedObjectFactory<ezQtConnection>& GetConnectionFactory();
   static ezVec2 GetLastMouseInteractionPos() { return s_LastMouseInteraction; }
 
+  struct VisualStyleFlags
```
jankrassnigg: Currently these are flags for on/off, but if in the future we were to add other modes for the lines or the arrows, this wouldn't be nice to extend. Might be better to have an enum for LineStyle and LineEndStyle or so.

Code/Tools/Libs/GuiFoundation/NodeEditor/Implementation/NodeScene.cpp
```diff
@@ -843,13 +803,25 @@ void ezQtNodeScene::ConnectPinsAction(const ezPin* pSourcePin, const ezPin* pTar
...
   {
-    ezConnectNodePinsCommand cmd;
-    cmd.m_ObjectSource = pSourcePin->GetParent()->GetGuid();
-    cmd.m_ObjectTarget = pTargetPin->GetParent()->GetGuid();
-    cmd.m_sSourcePin = pSourcePin->GetName();
-    cmd.m_sTargetPin = pTargetPin->GetName();
+    ezAddObjectCommand cmd;
+    cmd.m_pType = m_pManager->GetConnectionType();
+    cmd.m_NewObjectGuid.CreateNewUuid();
+    // cmd.m_sParentProperty
```
jankrassnigg: cleanup

jankrassnigg (approved):

Code/Engine/Foundation/Types/Implementation/UniquePtr_inl.h
C-Core: You often have a unique ptr that holds an object in one data structure and a raw pointer in another data structure. Then you need to be able to compare those two without always having to write .Borrow() on the unique ptr. The new variants are needed for some case of "const T", can't remember which exactly ;)

Code/Engine/Core/Utils/Blackboard.h
C-Core: I reverted these changes as they should not be part of this PR.

Code/Tools/Libs/ToolsFoundation/NodeObject/Implementation/DocumentNodeManager.cpp
```diff
@@ -268,126 +298,259 @@ void ezDocumentNodeManager::MoveNode(const ezDocumentObject* pObject, const ezVe
...
+      }
+    }
+  }
+}
 
-        for (const auto& con : data.m_Connections)
+void ezDocumentNodeManager::GetMetaDataHash(const ezDocumentObject* pObject, ezUInt64& inout_uiHash) const
+{
+  if (IsNode(pObject))
+  {
```
Sanakan8472, line 430: Why is this not hashing the node positon?

Code/Tools/Libs/ToolsFoundation/NodeObject/DocumentNodeManager.h
```diff
@@ -123,84 +126,56 @@ class EZ_TOOLSFOUNDATION_DLL ezDocumentNodeManager : public ezDocumentObjectMana
   };
 
   bool IsNode(const ezDocumentObject* pObject) const;
-  ezStatus CanConnect(const ezPin* pSource, const ezPin* pTarget, CanConnectResult& result) const;
+  bool IsConnection(const ezDocumentObject* pObject) const;
+
+  ezArrayPtr<const ezConnection* const> GetConnections(const ezPin& pin) const;
+  bool HasConnections(const ezPin& pin) const;
+  bool IsConnected(const ezPin& source, const ezPin& target) const;
+  ezStatus CanConnect(const ezPin& source, const ezPin& target, CanConnectResult& result) const;
```
Sanakan8472: Nit: I would suggest making the CanConnect and Connect functions roughly have the same signature. In this case, you ommit the pObject so it could be that later an unsupported type is thrown in. Similiar to CanAdd on the object manager, you could put in `const ezRTTI* pObjectType` to identify the type of connection you want to add. I know in this case it is not really necessary as every document type currently has one fixed connection type so the logic is rather uninteresting.

Sanakan8472 (approved):

Code/Tools/Libs/ToolsFoundation/NodeObject/Implementation/DocumentNodeManager.cpp
C-Core: Because the previous implementation didn't hash it either. I just copied the logic from the previous implementations adapted it to the new data structure and moved it to the manager. Do you think it should hash the node position as well?

Code/Tools/Libs/ToolsFoundation/NodeObject/DocumentNodeManager.h
C-Core: Done

Code/Tools/Libs/ToolsFoundation/NodeObject/Implementation/DocumentNodeManager.cpp
C-Core: Hash is only used for transform and for that node position is irrelevant. Added a comment in the code.

PR #729, opened 2022-10-16
Fix editor crash on exit

Ingrater:


Code/Tools/Libs/GuiFoundation/ContainerWindow/ContainerWindow.cpp
```diff
@@ -91,14 +91,6 @@ ezQtContainerWindow::~ezQtContainerWindow()
   ezQtDocumentWindow::s_Events.RemoveEventHandler(ezMakeDelegate(&ezQtContainerWindow::DocumentWindowEventHandler, this));
   ezToolsProject::s_Events.RemoveEventHandler(ezMakeDelegate(&ezQtContainerWindow::ProjectEventHandler, this));
   ezQtUiServices::s_Events.RemoveEventHandler(ezMakeDelegate(&ezQtContainerWindow::UIServicesEventHandler, this));
-
-  // The dock manager does not take ownership of dock widgets.
-  auto dockWidgets = m_DockManager->dockWidgetsMap();
```
Ingrater, line 96: @Sanakan8472 I suspect that this code is no longer neccessary. Do you remember what the issue back then was when you wrote this code?

jankrassnigg (approved):

Code/Tools/Libs/GuiFoundation/ContainerWindow/ContainerWindow.cpp
Sanakan8472: @Ingrater as the comment states, the dock manager does not take ownership of the dock widgets so they would just leak. You can verify this by just putting a breakpoint in one of the dtors. Make sure to test both shown and hidden docks widgets.

Code/Tools/Libs/GuiFoundation/ContainerWindow/ContainerWindow.cpp
```diff
@@ -153,10 +145,18 @@ void ezQtContainerWindow::closeEvent(QCloseEvent* e)
   {
     pWindow->DisableWindowLayoutSaving();
   }
+
+  // We need to destroy the dock manager here, doing it in the constructor leads to an access violation.
```
Sanakan8472, line 149: NIt: destructor instead of constructor

Code/Tools/Libs/GuiFoundation/ContainerWindow/ContainerWindow.cpp
Sanakan8472: Seems to be working fine now. maybe the behaviour did change when we updated ADS the last time?

Ingrater: Yes, probably

PR #741, opened 2022-10-17
Fixed allocations during compare and hashing

C-Core:
* When a key that requires allocation (e.g. an dynamic array) was used with ezMap or ezSet a lookup with a compatible type resulted in a temp key allocated because the comparer only allowed the exact key type by default.
* Added operator< for arrays.
* Strings as key in ezHashSet or ezHashTable could also result in a temp string allocated during lookup.
* Ported % optimization from ezHashTable to ezHashSet.

### DIFF
--- a/Code/Engine/Foundation/Containers/Implementation/ArrayBase_inl.h
+++ b/Code/Engine/Foundation/Containers/Implementation/ArrayBase_inl.h
@@ -72,6 +72,12 @@ EZ_ALWAYS_INLINE bool ezArrayBase<T, Derived>::operator!=(const ezArrayPtr<const
 }
 
 template <typename T, typename Derived>
+EZ_ALWAYS_INLINE bool ezArrayBase<T, Derived>::operator<(const ezArrayPtr<const T>& rhs) const
+{
+  return GetArrayPtr() < rhs;
+}
+
+template <typename T, typename Derived>
 EZ_ALWAYS_INLINE const T& ezArrayBase<T, Derived>::operator[](const ezUInt32 uiIndex) const
 {
   EZ_ASSERT_DEV(uiIndex < m_uiCount, "Out of bounds access. Array has {0} elements, trying to access element at index {1}.", m_uiCount, uiIndex);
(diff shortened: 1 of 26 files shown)


Code/Engine/Foundation/Containers/Implementation/HashSet_inl.h
```diff
@@ -279,7 +279,7 @@ bool ezHashSetBase<K, H>::Insert(CompatibleKeyType&& key)
 {
   Reserve(m_uiCount + 1);
 
-  ezUInt32 uiIndex = H::Hash(key) % m_uiCapacity;
+  ezUInt32 uiIndex = H::Hash(key) & (m_uiCapacity - 1);
```
jakrams, line 282: add an assert that m_uiCapacity has a value with which this works ?

jakrams (approved):

Code/Engine/Foundation/Containers/Implementation/HashSet_inl.h
C-Core: There is an assert in the SetCapacity method below, which is the only method modifying m_uiCapacity.

Code/Engine/Foundation/Containers/Implementation/ArrayBase_inl.h
```diff
@@ -71,6 +71,12 @@ EZ_ALWAYS_INLINE bool ezArrayBase<T, Derived>::operator!=(const ezArrayPtr<const
   return !(*this == rhs);
 }
 
+template <typename T, typename Derived>
+EZ_ALWAYS_INLINE bool ezArrayBase<T, Derived>::operator<(const ezArrayPtr<const T>& rhs) const
+{
+  return GetArrayPtr() < rhs;
```
Ingrater, line 77: This odering is not stable over a serialze/ deserialze operation. E.g. if I where to serialize a `ezDynamicArray<ezDynamicArray<int>>` which was sorten before serialization, it will no longer be sorted after deserialization. Do we care about this anywhere?

C-Core: No we don't care about ordering.

PR #748, opened 2022-10-19
Material improvements

### LINKED ISSUE #739: [FEATURE] Materials should support ARM texture types (baked ao / roughness / metallic)
Would be useful if materials had a mode that uses a packed ARM texture, rather than separate textures.

Once the feature is available in principle, automatic import for such textures would also be nice.

### LINKED ISSUE #745: [BUG] Some meshes don't import smoothly
This mesh assets during import:
https://polyhaven.com/a/lantern_chandelier_01

jankrassnigg:
* The material roughness value is now multiplied with the roughness value from roughness textures, to allow tweaking the final value.
* Enabling normal maps and roughness maps are now two separate bools. This required to update all material assets, since this is unfortunately shader state that can't be automatically patched. This is a breaking change and if you have custom materials, you will need to adjust your materials as well. The fix is to enable "UseRoughnessTexture" and "UseNormalTexture" in the materials as necessary.
* Added support for packed Occlusion/Roughness/Metalness (ORM) textures. If enabled, these three values are taken from a single texture rather than three different ones.
* Added automatic import of ORM textures, when available.
* Several minor tweaks and fixes.
* Fixes #745
* Fixes #739


Code/Editor/EditorFramework/Assets/Implementation/AssetDocumentGenerator.cpp
```diff
@@ -301,9 +301,7 @@ void ezAssetDocumentGenerator::SortAndSelectBestImportOption(ezDynamicArray<ezAs
 
   for (auto& singleImport : allImports)
   {
-
-    singleImport.m_ImportOptions.Sort(
-      [](const ezAssetDocumentGenerator::Info& lhs, const ezAssetDocumentGenerator::Info& rhs) -> bool { return lhs.m_sName < rhs.m_sName; });
+    // singleImport.m_ImportOptions.Sort([](const ezAssetDocumentGenerator::Info& lhs, const ezAssetDocumentGenerator::Info& rhs) -> bool { return ezTranslate(lhs.m_sName) < ezTranslate(rhs.m_sName); });
```
C-Core: Not needed anymore?

C-Core (approved):

PR #751, opened 2022-10-21
Character improvements

jrouwe:
* Update to Jolt version f59dbeb9917103a0cca7c4935b7e9a65537f8632
* Added supporting volume for character, contacts that fall outside will return the state NotSupported and will no longer prevent the character from falling
* Improved fraction correction in GetFirstContactForSweep so that it uses the actual face of the contact instead of estimating it with an infinite plane. Before the character would jump up a lot during stair walk with a frequency of 300Hz.
* Allow walk stairs to slide during horizontal movement. This fixes a bug where a wall next to a stairs would prevent the player from doing stair walking
* Ensure that we prefer returning a supporting contact even if the normal of a non-supporting contact is pointing up more


jankrassnigg:
Very cool, thanks a lot !

jankrassnigg (approved):

PR #753, opened 2022-10-23
Using the new CharacterVirtual::ExtendedUpdate function

jrouwe:
- Sync to Jolt version 8f82525a41f0fcd0a9faabe45295b2909eb58614
- Removed individual calls to Update, WalkStairs and StickToFloor and replaced them with a single call to ExtendedUpdate
- Changed the way gravity is applied to the character


jrouwe:
Please test this change thoroughly first! I may have made code changes that you don't agree with...

jankrassnigg:
Looks good, I'll merge it and test it thoroughly.

PR #761, opened 2022-10-24
State Machine

C-Core:
The state machine runtime consists of the following classes:

* ezStateMachineState: the base class for state machine states. Derive from this class to implement custom states.
* ezStateMachineTransition: the base class for state machine transitions. The most used implementation will be ezStateMachineTransition_BlackboardConditions, which triggers a transition when certain conditions on blackboard entries are met. But custom transition classes can be implemented as well.
* ezStateMachineDescription: describes the structure of a state machine, e.g. which states it has and how the are connected with transitions.
* ezStateMachineInstance: the actual state machine at runtime. Automatically triggers transitions into other states when their condition is met.

The typical use case would be to have one ezStateMachineDescription and multiple ezStateMachineInstance that reference the description. In this case states and transitions are shared between instances and thus must only operate on the passed instance data. Use EZ_INSTANCE_DATA_TYPE_ATTRIBUTE to declare an instance data type. This allows us to very quickly create instances at runtime without the need to instantiate the whole structure for each state machine instance. 
It also possible to use an ezStateMachineInstance without an description. In this case the states have to be owned by an overarching system and can be unique to each instance. This can be useful for very highlevel system with very few instances.

In the Editor it is possible to create state machine assets: 
image

Those can be used on ezStateMachineComponent or as nested state machines in the ezStateMachineState_NestedStateMachine.


Code/EditorPlugins/Assets/EditorPluginAssets/StateMachineAsset/StateMachineGraph.cpp
```diff
@@ -0,0 +1,227 @@
+#include <EditorPluginAssets/EditorPluginAssetsPCH.h>
+
+#include <EditorPluginAssets/StateMachineAsset/StateMachineGraph.h>
+#include <EditorPluginAssets/StateMachineAsset/StateMachineGraphQt.moc.h>
+#include <GameEngine/StateMachine/StateMachine.h>
+#include <GuiFoundation/UIServices/DynamicStringEnum.h>
+
+// clang-format off
+EZ_BEGIN_SUBSYSTEM_DECLARATION(EditorFramework, StateMachine)
```
jankrassnigg: this isn't in the EditorFramework, so the group is wrong, and the dependency on the AssetPlugin is probably also wrong

Code/EditorPlugins/Assets/EditorPluginAssets/StateMachineAsset/StateMachineGraph.h
```diff
@@ -0,0 +1,89 @@
...
+class ezStateMachineNode : public ezStateMachineNodeBase
+{
+  EZ_ADD_DYNAMIC_REFLECTION(ezStateMachineNode, ezStateMachineNodeBase);
+
+public:
+  ezString m_sName;
+  ezStateMachineState* m_pType = nullptr;
+};
+
+class ezStateMachineNodeAny : public ezStateMachineNodeBase
```
jankrassnigg, line 48: some comments in this file would be nice, especially about this Any node thing

jankrassnigg (commented):
will continue reviewing later

Code/EditorPlugins/Assets/EditorPluginAssets/StateMachineAsset/StateMachineGraph.cpp
C-Core: Copy & paste from visual script asset ;)

Code/EditorPlugins/Assets/EditorPluginAssets/StateMachineAsset/StateMachineGraphQt.moc.h
```diff
@@ -0,0 +1,57 @@
...
+};
+
+class ezQtStateMachineConnection : public ezQtConnection
+{
+public:
+  ezQtStateMachineConnection(QGraphicsItem* parent = nullptr);
+
+  virtual QPen DeterminePen() const override;
+
+  bool m_bExecutionHighlight = false;
```
jankrassnigg: Isn't the bool redundant? If the highlight time is in the past, nothing to do? Otherwise would you need to reset the bool at some point?

Code/EditorPlugins/Assets/EditorPluginAssets/VisualScriptAsset/VisualScriptTypeRegistry.cpp
```diff
@@ -72,20 +72,20 @@ EZ_BEGIN_SUBSYSTEM_DECLARATION(EditorFramework, VisualScript)
     EZ_DEFAULT_NEW(ezVisualScriptTypeRegistry);
 
     ezVisualScriptTypeRegistry::GetSingleton()->UpdateNodeTypes();
-    const ezRTTI* pBaseType = ezVisualScriptTypeRegistry::GetSingleton()->GetNodeBaseType();
+    const ezRTTI* pBaseType = ezVisualScriptTypeRegistry::GetSingleton()->GetNodeBaseType();
 
     ezQtNodeScene::GetPinFactory().RegisterCreator(ezGetStaticRTTI<ezVisualScriptPin>(), [](const ezRTTI* pRtti)->ezQtPin* { return new ezQtVisualScriptPin(); });
-    ezQtNodeScene::GetConnectionFactory().RegisterCreator(ezGetStaticRTTI<ezVisualScriptConnection>(), [](const ezRTTI* pRtti)->ezQtConnection* { return new ezQtVisualScriptConnection(); });
+    /*ezQtNodeScene::GetConnectionFactory().RegisterCreator(ezGetStaticRTTI<ezVisualScriptConnection>(), [](const ezRTTI* pRtti)->ezQtConnection* { return new ezQtVisualScriptConnection(); });*/
```
jankrassnigg: cleanup

Code/EditorPlugins/Assets/EditorPluginAssets/VisualScriptAsset/VisualScriptTypeRegistry.cpp
```diff
@@ -72,20 +72,20 @@ EZ_BEGIN_SUBSYSTEM_DECLARATION(EditorFramework, VisualScript)
...
   ON_CORESYSTEMS_SHUTDOWN
-  {
-    const ezRTTI* pBaseType = ezVisualScriptTypeRegistry::GetSingleton()->GetNodeBaseType();
+  {
+    const ezRTTI* pBaseType = ezVisualScriptTypeRegistry::GetSingleton()->GetNodeBaseType();
     ezQtNodeScene::GetNodeFactory().UnregisterCreator(pBaseType);
 
     ezQtNodeScene::GetPinFactory().UnregisterCreator(ezGetStaticRTTI<ezVisualScriptPin>());
-    ezQtNodeScene::GetConnectionFactory().UnregisterCreator(ezGetStaticRTTI<ezVisualScriptConnection>());
+    //ezQtNodeScene::GetConnectionFactory().UnregisterCreator(ezGetStaticRTTI<ezVisualScriptConnection>());
```
jankrassnigg: cleanup

Code/EditorPlugins/ProcGen/EditorPluginProcGen/ProcGenGraphAsset/ProcGenGraphAsset.cpp
```diff
@@ -169,9 +161,9 @@ ezStatus ezProcGenGraphAssetDocument::WriteAsset(ezStreamWriter& stream, const e
       {
         EZ_SUCCEED_OR_RETURN(WriteByteCode(pPlacementNode));
 
-        ezProcGenNodeBase* pPGNode;
-        EZ_VERIFY(context.m_DocObjToProcGenNodeTable.TryGetValue(pPlacementNode, pPGNode), "Implementation error");
-        auto pPlacementOutput = ezStaticCast<ezProcGen_PlacementOutput*>(pPGNode);
+        auto pPGNode = context.m_DocObjToProcGenNodeTable.GetValue(pPlacementNode);
+        EZ_VERIFY(pPGNode != nullptr, "Implementation error");
```
jankrassnigg: are these nullptr checks really necessary? the error message is pointless and with a nullptr it will crash in the next line anyway

Code/Engine/Foundation/Memory/Implementation/AllocatorBase_inl.h
```diff
@@ -32,6 +32,12 @@ namespace ezInternal
       other.m_pAllocator = nullptr;
     }
 
+    EZ_ALWAYS_INLINE NewInstance(std::nullptr_t)
+    {
+      m_pInstance = nullptr;
```
jankrassnigg: Arent't those the default values anyway ?

Code/Engine/GameEngine/StateMachine/StateMachineComponent.h
```diff
@@ -0,0 +1,107 @@
...
+  ezHashedString m_sNewStateName;
+
+private:
+  const char* GetNewStateName() const { return m_sNewStateName; }
+  void SetNewStateName(const char* szName) { m_sNewStateName.Assign(szName); }
+};
+
+//////////////////////////////////////////////////////////////////////////
+
+class ezStateMachineState_SendMsg : public ezStateMachineState
```
jankrassnigg, line 28: I guess this file could use a couple of comments. For example what message this class sends.

Code/Engine/GameEngine/StateMachine/StateMachineComponent.h
```diff
@@ -0,0 +1,107 @@
...
+  ezStateMachineInstance* GetStateMachineInstance() { return m_pStateMachineInstance.Borrow(); }
+  const ezStateMachineInstance* GetStateMachineInstance() const { return m_pStateMachineInstance.Borrow(); }
+
+  void SetResource(const ezStateMachineResourceHandle& hResource);
+  const ezStateMachineResourceHandle& GetResource() const { return m_hResource; }
+
+  void SetResourceFile(const char* szFile); // [ property ]
+  const char* GetResourceFile() const;      // [ property ]
+
+  void SetInitialState(const char* szName);                       // [ property ]
```
jankrassnigg, line 108: comment please what this is needed for

Code/Engine/GameEngine/StateMachine/Implementation/StateMachineResource.cpp
```diff
@@ -0,0 +1,88 @@
...
+}
+
+ezStateMachineResource::~ezStateMachineResource() = default;
+
+ezUniquePtr<ezStateMachineInstance> ezStateMachineResource::CreateInstance(ezReflectedClass& owner)
+{
+  if (m_pDescription != nullptr)
+  {
+    auto pInstance = EZ_DEFAULT_NEW(ezStateMachineInstance, owner, m_pDescription);
+    return pInstance;
```
jankrassnigg: return without temp variable ?

Code/Engine/GameEngine/StateMachine/Implementation/StateMachineResource.cpp
```diff
@@ -0,0 +1,88 @@
...
+  // skip the absolute file path data that the standard file reader writes into the stream
+  {
+    ezStringBuilder sAbsFilePath;
+    (*Stream) >> sAbsFilePath;
+  }
+
+  ezAssetFileHeader AssetHash;
+  AssetHash.Read(*Stream).IgnoreResult();
+
+  ezUniquePtr<ezStateMachineDescription> pDescription = EZ_DEFAULT_NEW(ezStateMachineDescription);
```
jankrassnigg, line 65: I don't think you need to allocate the descriptor, can just exist on the stack.

Code/Engine/Foundation/IO/TypeVersionContext.h
```diff
@@ -0,0 +1,64 @@
...
+  ezHashSet<const ezRTTI*> m_KnownTypes;
+};
+
+/// \brief Use this class to restore type versions written to a stream using a ezTypeVersionWriteContext.
+class EZ_FOUNDATION_DLL ezTypeVersionReadContext : public ezSerializationContext<ezTypeVersionReadContext>
+{
+  EZ_DECLARE_SERIALIZATION_CONTEXT(ezTypeVersionReadContext);
+
+public:
+  /// \brief Setup the type info used internally.
```
jankrassnigg: That comment doesn't seem to make sense.

Code/Engine/Foundation/IO/TypeVersionContext.h
```diff
@@ -0,0 +1,64 @@
+#pragma once
+
+#include <Foundation/Containers/HashSet.h>
```
jankrassnigg, line 3: Include for HashTable is missing

Code/Engine/Foundation/IO/Implementation/TypeVersionContext.cpp
```diff
@@ -0,0 +1,105 @@
...
+    stream >> sTypeName;
+    stream >> uiTypeVersion;
+
+    if (const ezRTTI* pType = ezRTTI::FindTypeByName(sTypeName))
+    {
+      m_TypeVersions.Insert(pType, uiTypeVersion);
+    }
+    else
+    {
+      ezLog::Warning("Ignoring unknown type '{}'", sTypeName);
```
jankrassnigg, line 92: Maybe add a LOG_BLOCK in this function.

Code/Engine/GameEngine/StateMachine/Implementation/StateMachineBuiltins.h
```diff
@@ -0,0 +1,58 @@
+#pragma once
+
+#include <Core/Utils/Blackboard.h>
+#include <GameEngine/StateMachine/StateMachineResource.h>
+
+class EZ_GAMEENGINE_DLL ezStateMachineState_NestedStateMachine : public ezStateMachineState
```
jankrassnigg, line 7: please comment what this class is used for and how

Code/Engine/GameEngine/StateMachine/Implementation/StateMachineBuiltins.h
```diff
@@ -0,0 +1,58 @@
...
+{
+  EZ_ADD_DYNAMIC_REFLECTION(ezStateMachineState_NestedStateMachine, ezStateMachineState);
+
+public:
+  ezStateMachineState_NestedStateMachine(const char* szName = nullptr);
```
jankrassnigg: Going forward we should try to use ezStringView rather than const char*

Code/Engine/GameEngine/StateMachine/Implementation/StateMachineComponent.cpp
```diff
@@ -0,0 +1,312 @@
...
+#include <GameEngine/Gameplay/BlackboardComponent.h>
+#include <GameEngine/StateMachine/StateMachineComponent.h>
+
+// clang-format off
+EZ_IMPLEMENT_MESSAGE_TYPE(ezMsgStateMachineStateChanged);
+EZ_BEGIN_DYNAMIC_REFLECTED_TYPE(ezMsgStateMachineStateChanged, 1, ezRTTIDefaultAllocator<ezMsgStateMachineStateChanged>)
+{
+  EZ_BEGIN_PROPERTIES
+  {
+    EZ_ACCESSOR_PROPERTY("NewStateName", GetNewStateName, SetNewStateName),
```
jankrassnigg, line 15: Could it also be useful to also send along the previous state? In Qt state changes often pass along both.
Might sometimes be more interesting to know that state you exited and not which one you entered.

Code/Engine/GameEngine/StateMachine/Implementation/StateMachineComponent.cpp
```diff
@@ -0,0 +1,312 @@
...
+ezStateMachineState_SendMsg::ezStateMachineState_SendMsg(const char* szName /*= nullptr*/)
+  : ezStateMachineState(szName)
+{
+}
+
+ezStateMachineState_SendMsg::~ezStateMachineState_SendMsg() = default;
+
+void ezStateMachineState_SendMsg::OnEnter(ezStateMachineInstance& instance, void* pStateInstanceData) const
+{
+  if (auto pOwner = ezDynamicCast<ezStateMachineComponent*>(&instance.GetOwner()))
```
jankrassnigg: So this state actually only works with components at the moment. This should be documented in the header.

Code/Engine/GameEngine/StateMachine/Implementation/StateMachineComponent.cpp
```diff
@@ -0,0 +1,312 @@
...
+    }
+    m_ComponentsToReload.Clear();
+  }
+
+  // update
+  {
+    for (auto it = this->m_ComponentStorage.GetIterator(context.m_uiFirstComponentIndex, context.m_uiComponentCount); it.IsValid(); ++it)
+    {
+      ComponentType* pComponent = it;
+      if (pComponent->IsActiveAndInitialized())
```
jankrassnigg, line 161: ActiveAndSimulating ?

jankrassnigg: I know the whole function is only then executed, so I guess this would be redundant.

Code/Engine/GameEngine/StateMachine/Implementation/StateMachineComponent.cpp
```diff
@@ -0,0 +1,312 @@
...
+    new ezCategoryAttribute("Gameplay"),
+  }
+  EZ_END_ATTRIBUTES;
+}
+
+EZ_END_DYNAMIC_REFLECTED_TYPE
+// clang-format on
+
+ezStateMachineComponent::ezStateMachineComponent() = default;
+ezStateMachineComponent::ezStateMachineComponent(ezStateMachineComponent&& other) = default;
```
jankrassnigg, line 220: why do we need the move constructor ?

Code/Engine/GameEngine/StateMachine/Implementation/StateMachineComponent.cpp
```diff
@@ -0,0 +1,312 @@
...
+
+  s >> m_hResource;
+  s >> m_sInitialState;
+}
+
+void ezStateMachineComponent::OnActivated()
+{
+  SUPER::OnActivated();
+
+  InstantiateStateMachine();
```
jankrassnigg, line 248: why not in OnSimulationStarted ?

Code/Engine/GameEngine/StateMachine/Implementation/StateMachineComponent.cpp
```diff
@@ -0,0 +1,312 @@
...
+}
+
+void ezStateMachineComponent::SetResource(const ezStateMachineResourceHandle& hResource)
+{
+  if (m_hResource == hResource)
+    return;
+
+  m_hResource = hResource;
+
+  if (IsActiveAndInitialized())
```
jankrassnigg, line 265: IsActiveAndSimulating ?

Code/Engine/GameEngine/StateMachine/Implementation/StateMachineComponent.cpp
```diff
@@ -0,0 +1,312 @@
...
+
+const char* ezStateMachineComponent::GetResourceFile() const
+{
+  if (!m_hResource.IsValid())
+    return "";
+
+  return m_hResource.GetResourceID();
+}
+
+void ezStateMachineComponent::SetInitialState(const char* szName)
```
jankrassnigg, line 292: ezStringView

Code/Engine/GameEngine/StateMachine/Implementation/StateMachineComponent.cpp
```diff
@@ -0,0 +1,312 @@
...
+
+  m_sInitialState = sInitialState;
+
+  if (IsActiveAndInitialized())
+  {
+    InstantiateStateMachine();
+  }
+}
+
+bool ezStateMachineComponent::SetState(const char* szName)
```
jankrassnigg: ezStringView

jankrassnigg: Or ezHashedString. Maybe that's even better performance-wise.

Also, I think this should return ezResult rather than bool

Code/Engine/GameEngine/StateMachine/Implementation/StateMachineComponent.cpp
```diff
@@ -0,0 +1,312 @@
...
+  }
+
+  return false;
+}
+
+void ezStateMachineComponent::SendStateChangedMsg(ezMsgStateMachineStateChanged& msg, ezTime delay)
+{
+  if (delay > ezTime::Zero())
+  {
+    m_StateChangedSender.PostEventMessage(msg, this, GetOwner(), delay, ezObjectMsgQueueType::NextFrame);
```
jankrassnigg, line 325: IIRC ezObjectMsgQueueType::NextFrame is the default value so could be omitted.

Code/Engine/GameEngine/StateMachine/Implementation/StateMachineComponent.cpp
```diff
@@ -0,0 +1,312 @@
...
+  }
+  else
+  {
+    m_StateChangedSender.SendEventMessage(msg, this, GetOwner());
+  }
+}
+
+void ezStateMachineComponent::InstantiateStateMachine()
```
jankrassnigg, line 333: Maybe this function should return ezResult.

Code/Engine/GameEngine/StateMachine/StateMachine.h
```diff
@@ -0,0 +1,182 @@
...
+#include <Foundation/Containers/SmallArray.h>
+#include <Foundation/Reflection/Reflection.h>
+#include <Foundation/Strings/HashedString.h>
+#include <Foundation/Types/SharedPtr.h>
+
+class ezBlackboard;
+class ezStateMachineInstance;
+
+/// \brief Base class for a state in a state machine.
+/// Note that states are shared between multiple instances and thus
```
jankrassnigg, line 16: needs a blank (commented) line, to separate brief and detailed comment

Code/Engine/GameEngine/StateMachine/StateMachine.h
```diff
@@ -0,0 +1,182 @@
...
+/// shouldn't modify any data on their own but always operate on the passed instance.
+class EZ_GAMEENGINE_DLL ezStateMachineState : public ezReflectedClass
+{
+  EZ_ADD_DYNAMIC_REFLECTION(ezStateMachineState, ezReflectedClass);
+
+public:
+  ezStateMachineState(const char* szName = nullptr);
```
jankrassnigg: Pass in ezHashedString or ezStringView

Code/Engine/GameEngine/StateMachine/StateMachine.h
```diff
@@ -0,0 +1,182 @@
...
+
+  void SetName(const char* szName);
+  const char* GetName() const { return m_sName; }
+  const ezHashedString& GetNameHashed() const { return m_sName; }
+
+protected:
+  friend class ezStateMachineDescription;
+  friend class ezStateMachineInstance;
+
+  virtual void OnEnter(ezStateMachineInstance& instance, void* pStateInstanceData) const = 0;
```
jankrassnigg: Maybe pass in pointer to previous state? Might be useful.

Code/Engine/GameEngine/StateMachine/StateMachine.h
```diff
@@ -0,0 +1,182 @@
...
+
+  virtual ezResult Serialize(ezStreamWriter& stream) const;
+  virtual ezResult Deserialize(ezStreamReader& stream);
+
+private:
+  ezHashedString m_sName;
+};
+
+/// \brief Base class for a transition in a state machine.
+/// The target state of a transition is automatically set once its condition has been met.
```
jankrassnigg: blank comment line

Code/Engine/GameEngine/StateMachine/StateMachine.h
```diff
@@ -0,0 +1,182 @@
...
+  friend class ezStateMachineDescription;
+  friend class ezStateMachineInstance;
+
+  virtual bool IsConditionMet(ezStateMachineInstance& instance, void* pTransitionInstanceData) const = 0;
+
+  virtual ezResult Serialize(ezStreamWriter& stream) const;
+  virtual ezResult Deserialize(ezStreamReader& stream);
+};
+
+/// \brief Attribute to define instance data types
```
jankrassnigg: Sure, but what's an instance data type?
This name is very generic, if this is tied to state machines only, then the name should reflect that.

Code/Engine/GameEngine/StateMachine/StateMachine.h
```diff
@@ -0,0 +1,182 @@
...
+/// what states it has and how to transition between them.
+class EZ_GAMEENGINE_DLL ezStateMachineDescription : public ezRefCounted
+{
+  EZ_DISALLOW_COPY_AND_ASSIGN(ezStateMachineDescription);
+
+public:
+  ezStateMachineDescription();
+  ~ezStateMachineDescription();
+
+  ezUInt32 AddState(ezUniquePtr<ezStateMachineState>&& pState);
```
jankrassnigg, line 107: comment please: what is the meaning of the return value

Code/Engine/GameEngine/StateMachine/StateMachine.h
```diff
@@ -0,0 +1,182 @@
...
+
+  /// \brief A uiFromStateIndex of ezInvalidIndex generates a transition that can be done from any possible state.
```
jankrassnigg: The comment is very brief. So with an invalid start state index you can create a transition from any node to the target node. What if I add nodes/states later on? Will they also get that transition or does it only affect already added ones?

Code/Engine/GameEngine/StateMachine/StateMachine.h
```diff
@@ -0,0 +1,182 @@
...
+  ezResult Serialize(ezStreamWriter& stream) const;
+  ezResult Deserialize(ezStreamReader& stream);
+
+private:
+  friend class ezStateMachineInstance;
+
+  struct TransitionContext
+  {
+    ezUniquePtr<ezStateMachineTransition> m_pTransition;
+    ezUInt32 m_uiToStateIndex;
```
jankrassnigg: init

Code/Engine/GameEngine/StateMachine/StateMachine.h
```diff
@@ -0,0 +1,182 @@
...
+
+  ezDynamicArray<StateContext> m_States;
+  ezHashTable<ezHashedString, ezUInt32> m_StateNameToIndexTable;
+
+  ezDynamicArray<const ezInstanceDataTypeAttribute*> m_InstanceDataAttributes;
+  ezUInt32 m_uiTotalInstanceDataSize = 0;
+};
+
+/// \brief The state machine instance represents the actual state machine.
+/// Typically it is created from a description but for small use cases it can also be used without a description.
```
jankrassnigg, line 143: From the interface alone I don't see how to use this without a description.
Also, is there a real use case for such a scenario?

Code/Engine/GameEngine/StateMachine/Implementation/StateMachine.cpp
```diff
@@ -0,0 +1,462 @@
...
+// clang-format off
+EZ_BEGIN_DYNAMIC_REFLECTED_TYPE(ezStateMachineState, 1, ezRTTINoAllocator)
+EZ_END_DYNAMIC_REFLECTED_TYPE;
+// clang-format on
+
+ezStateMachineState::ezStateMachineState(const char* szName /*= nullptr*/)
+{
+  if (ezStringUtils::IsNullOrEmpty(szName) == false)
+  {
+    m_sName.Assign(szName);
```
jankrassnigg: I would assume that ezHashedString already earlies out on an empty string ?? If not, we should just fix it there.

Code/Engine/GameEngine/StateMachine/Implementation/StateMachine.cpp
```diff
@@ -0,0 +1,462 @@
...
+
+  auto& sStateName = pState->GetNameHashed();
+  if (sStateName.IsEmpty() == false)
+  {
+    EZ_VERIFY(m_StateNameToIndexTable.Contains(sStateName) == false, "A state with name '{}' already exists.", sStateName);
+    m_StateNameToIndexTable.Insert(sStateName, uiIndex);
+  }
+
+  auto& stateContext = m_States.ExpandAndGetRef();
+  if (auto pInstanceDataAttribute = pState->GetDynamicRTTI()->GetAttributeByType<ezInstanceDataTypeAttribute>())
```
jankrassnigg: in theory you could have multiple such attributes, would that make sense ?

jankrassnigg:
In a nested state, you can select your own state machine as the resource, and on transform it will just crash.

Code/Engine/GameEngine/StateMachine/Implementation/StateMachineBuiltins.cpp
```diff
@@ -0,0 +1,184 @@
+#include <GameEngine/GameEnginePCH.h>
+
+#include <Foundation/IO/TypeVersionContext.h>
+#include <GameEngine/StateMachine/Implementation/StateMachineBuiltins.h>
+
```
jankrassnigg, line 5: Currently you only have the send message state. I could imagine a "log" state might be useful at least for debugging. Also in a lot of cases I would guess a state that itself does nothing is already sufficient. Both "do nothing" and "log something" could be handled by the same state, though.

Code/Engine/GameEngine/StateMachine/Implementation/StateMachineBuiltins.cpp
```diff
@@ -0,0 +1,184 @@
...
+  if (bSetInitialState)
+  {
+    pStateMachineInstance->SetStateOrFallback(m_sInitialState).IgnoreResult();
+  }
+}
+
+void ezStateMachineState_NestedStateMachine::OnExit(ezStateMachineInstance& instance, void* pStateInstanceData) const
+{
+  if (m_bResetOnExit)
```
jankrassnigg: Not sure how reset on enter and exit make a difference? I will always exit before I enter again, so whether I reset the start state on enter or on exit, where is the difference?

Code/Engine/GameEngine/StateMachine/Implementation/StateMachine.cpp
```diff
@@ -0,0 +1,462 @@
...
+  stateContext.m_pState = std::move(pState);
+
+  return uiIndex;
+}
+
+ezResult ezStateMachineDescription::AddTransition(ezUInt32 uiFromStateIndex, ezUInt32 uiToStateIndex, ezUniquePtr<ezStateMachineTransition>&& pTransistion)
+{
+  if (uiFromStateIndex == uiToStateIndex)
+  {
+    ezLog::Error("Can't add a transition to itself");
```
jankrassnigg: Are these really errors or asserts? Also below.

Code/Engine/GameEngine/StateMachine/Implementation/StateMachine.cpp
```diff
@@ -0,0 +1,462 @@
...
+
+    for (ezUInt32 i = 0; i < uiNumStates; ++i)
+    {
+      stream >> sTypeName;
+      if (const ezRTTI* pType = ezRTTI::FindTypeByName(sTypeName))
+      {
+        ezUniquePtr<ezStateMachineState> pState = pType->GetAllocator()->Allocate<ezStateMachineState>();
+        EZ_SUCCEED_OR_RETURN(pState->Deserialize(stream));
+
+        EZ_VERIFY(AddState(std::move(pState)) == i, "Implementation error");
```
jankrassnigg, line 211: Instead of EZ_VERIFY you can just use ".AssertSuccess()" on ezResult.

Code/Engine/GameEngine/StateMachine/Implementation/StateMachine.cpp
```diff
@@ -0,0 +1,462 @@
...
+{
+  if (pState != nullptr && m_pDescription != nullptr)
+  {
+    ezUInt32 uiStateIndex = ezInvalidIndex;
+    if (m_pDescription->m_StateNameToIndexTable.TryGetValue(pState->GetNameHashed(), uiStateIndex))
+    {
+      SetStateInternal(uiStateIndex);
+      return;
+    }
+  }
```
jankrassnigg, line 310: If we have a description, but can't find the given state, wouldn't that be an error ?

Code/Engine/GameEngine/StateMachine/Implementation/StateMachine.cpp
```diff
@@ -0,0 +1,462 @@
...
+    m_InstanceData.Clear();
+  }
+}
+
+void ezStateMachineInstance::SetState(ezStateMachineState* pState)
+{
+  if (pState != nullptr && m_pDescription != nullptr)
+  {
+    ezUInt32 uiStateIndex = ezInvalidIndex;
+    if (m_pDescription->m_StateNameToIndexTable.TryGetValue(pState->GetNameHashed(), uiStateIndex))
```
jankrassnigg: Couldn't you just call SetState(hashedName) here ?

Code/Engine/GameEngine/StateMachine/Implementation/StateMachine.cpp
```diff
@@ -0,0 +1,462 @@
...
+  }
+}
+
+void ezStateMachineInstance::ExitCurrentState()
+{
+  if (m_pCurrentState != nullptr)
+  {
+    void* pInstanceData = GetCurrentStateInstanceData();
+    m_pCurrentState->OnExit(*this, pInstanceData);
+  }
```
jankrassnigg, line 416: set m_pCurrentState to null here ?

Code/Engine/GameEngine/StateMachine/Implementation/StateMachine.cpp
```diff
@@ -0,0 +1,462 @@
...
+  {
+    for (auto& transitionContext : *m_pCurrentTransitions)
+    {
+      void* pInstanceData = GetInstanceData(transitionContext.m_uiInstanceDataOffset);
+      if (transitionContext.m_pTransition->IsConditionMet(*this, pInstanceData))
+      {
+        return transitionContext.m_uiToStateIndex;
+      }
+    }
+  }
```
jankrassnigg: I would swap the two. The transitions on the current state always should have precedence over the any-transitions.

jankrassnigg (changes requested):

Code/Engine/Foundation/Memory/Implementation/AllocatorBase_inl.h
C-Core: No, NewInstance has no default constructor

Code/Engine/GameEngine/StateMachine/Implementation/StateMachineResource.cpp
C-Core: The description is hold by a shared ptr since it is shared by multiple instances thus is needs to be allocated on the heap.

Code/Engine/GameEngine/StateMachine/Implementation/StateMachine.cpp
C-Core: AddState doesn't return an ezResult and I need to compare to i

Code/Engine/GameEngine/StateMachine/Implementation/StateMachineBuiltins.cpp
C-Core: Maybe, but not as part of this PR

Code/Engine/GameEngine/StateMachine/Implementation/StateMachine.cpp
C-Core: You could, but doesn't make any sense

Code/Engine/GameEngine/StateMachine/StateMachine.h
C-Core: You could use a state machine instance without a description for very high level use cases, like a gamestate with states like main menu, ingame, pause menu etc. Then you could hold the states in your gamestate and use the state machine instance just to correctly keep track of the current state and call the corresponding state methods.

Code/Engine/Foundation/IO/TypeVersionContext.h
C-Core: HashTable is implicitely included by memorystream.h

C-Core:
> In a nested state, you can select your own state machine as the resource, and on transform it will just crash.

Yes I know, but we don't have a concept to prevent that. Same goes for e.g. material assets, you can assign it self as parent.

Code/Engine/GameEngine/StateMachine/Implementation/StateMachineComponent.cpp
C-Core: Because you might want to see the effect of the initial state enter in the editor.

Code/Engine/GameEngine/StateMachine/Implementation/StateMachineComponent.cpp
C-Core: Same as above

Code/Engine/GameEngine/StateMachine/Implementation/StateMachineComponent.cpp
C-Core: ezHashedString and ezResult don't work because this function is reflected for scripting

Code/Engine/GameEngine/StateMachine/Implementation/StateMachineComponent.cpp
C-Core: because it has a unique_ptr as member

Code/Engine/GameEngine/StateMachine/Implementation/StateMachineBuiltins.cpp
C-Core: Made it one bool flag and renamed it. Also added comment to clarify what it does.

Code/Engine/Foundation/Memory/Implementation/AllocatorBase_inl.h
jankrassnigg: Well you could just initialize the variables in the class directly:

    T* m_pInstance = nullptr;
    ezAllocatorBase* m_pAllocator = nullptr;

jankrassnigg (approved):

PR #777, opened 2022-10-26
Send main docs before sub-docs

Sanakan8472:
#fixes 754. Main documents must be sent to the engine process before their sub-documents or the engine will crash.

### DIFF
--- a/Code/Editor/EditorFramework/IPC/EngineProcessConnection.cpp
+++ b/Code/Editor/EditorFramework/IPC/EngineProcessConnection.cpp
@@ -400,7 +400,7 @@ ezResult ezEditorEngineProcessConnection::RestartProcess()
   }
   docs.Sort([](const ezAssetDocument* a, const ezAssetDocument* b) {
     if (a->IsMainDocument() != b->IsMainDocument())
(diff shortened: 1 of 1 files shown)


Code/Editor/EditorFramework/IPC/EngineProcessConnection.cpp
```diff
@@ -392,10 +390,23 @@ ezResult ezEditorEngineProcessConnection::RestartProcess()
...
+
+  // Resend all open documents. Make sure to send main documents before child documents.
   for (auto it = m_DocumentByGuid.GetIterator(); it.IsValid(); ++it)
   {
-    SendDocumentOpenMessage(it.Value(), true);
+    docs.PushBack(it.Value());
+  }
+  docs.Sort([](const ezAssetDocument* a, const ezAssetDocument* b) {
+    if (a->IsMainDocument() != b->IsMainDocument())
+      return a->IsMainDocument() > b->IsMainDocument();
```
jankrassnigg: You don't need this bool-as-int-and-then-compare behavior, because you already know that a != b. Can just return whether a is the main document.

jankrassnigg (approved):
Approved with suggestions.

PR #781, opened 2022-10-28
Make it possible to build ezEngine on windows using clang + ninja

Ingrater:


Code/ThirdParty/Imgui/imconfig.h
```diff
@@ -18,7 +18,7 @@
 // Start ezEngine edit
 // Configure the DLL Import/Export Define
 #ifdef BUILDSYSTEM_COMPILE_ENGINE_AS_DLL
-#ifdef WIN32
+#ifdef _WIN32
```
jankrassnigg, line 21: These are third party libs, are you sure this is incorrect?

jankrassnigg: Maybe we should define the proper stuff in cmake, rather than change this code.

Code/ThirdParty/Lua/luaconf.h
```diff
@@ -12,7 +12,7 @@
 #include <stddef.h>
 
 // BEGIN EZ SPECIFIC
-#if defined(BUILDSYSTEM_COMPILE_ENGINE_AS_DLL) && defined(WIN32)
+#if defined(BUILDSYSTEM_COMPILE_ENGINE_AS_DLL) && defined(_WIN32)
```
jankrassnigg, line 15: Same here

Code/Engine/Foundation/ThirdParty/utf8/checked.h
```diff
@@ -28,6 +28,8 @@ DEALINGS IN THE SOFTWARE.
 #ifndef UTF8_FOR_CPP_CHECKED_H_2675DCD0_9480_4c0c_B92A_CC14C027B731
 #define UTF8_FOR_CPP_CHECKED_H_2675DCD0_9480_4c0c_B92A_CC14C027B731
 
+// NOLINTBEGIN(*)
```
jankrassnigg: Is this supposed to be in this pr?

Code/Engine/Foundation/Containers/IdTable.h
```diff
@@ -50,7 +50,7 @@ class ezIdTableBase
 
     explicit ConstIterator(const ezIdTableBase<IdType, ValueType>& idTable);
 
-    const ezIdTableBase<IdType, ValueType>& m_idTable;
+    const ezIdTableBase<IdType, ValueType>& m_IdTable;
     IndexType m_uiCurrentIndex; // current element index that this iterator points to.
```
jankrassnigg, line 54: Vital build fix? 😁

Code/ThirdParty/Imgui/imconfig.h
Ingrater: Yes this is correct, the issue only existiert due to a ezEngine modification of this lib (same as the others). So I fix the ezEngine edit.

Code/ThirdParty/Lua/luaconf.h
Ingrater: Same as above

Code/Engine/Foundation/Containers/IdTable.h
Ingrater: As this is the preparation for the next pr, does it really matter?

jankrassnigg: Of course not.

jankrassnigg (approved):

Code/ThirdParty/Imgui/imconfig.h
Ingrater: Some further details. The `WIN32` define is not standardized. It is only defined if you build from within visual studio. The `_WIN32` define on the other hand is build into the compiler (both clang and msvc). See https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170
So it's cleaner to use `_WIN32` instead of defining `WIN32` ourselfs.

PR #787, opened 2022-11-02
Color Scheme

C-Core:
* Added a new ezColorScheme class based on https://yeun.github.io/open-color/ which also replaces the color palette we had in ezColor.
* Added helper function to obtain nice looking colors for our UI scheme.
* Used the new UI colors in the Editor for node graphs, selection highlight, log and various other places.
* Also replaced a few visualizer colors, but there are still many to go. Can be done at a later point.

image


Code/Engine/Foundation/Math/ColorScheme.h
```diff
@@ -0,0 +1,92 @@
...
+    Orange,
+    Gray,
+
+    Count
+  };
+
+  /// \brief Get the scheme color with the given brightness (0..9) and with optional saturation and alpha.
+  EZ_FORCE_INLINE static ezColor GetColor(Enum schemeColor, ezUInt8 uiBrightness, float fSaturation = 1.0f, float fAlpha = 1.0f)
+  {
+    EZ_ASSERT_DEV(uiBrightness <= 9, "Brighness is too large");
```
jankrassnigg: Typo in assert msg

Code/Engine/Foundation/Math/ColorScheme.h
```diff
@@ -0,0 +1,92 @@
...
+
+private:
+  EZ_ALWAYS_INLINE constexpr static void GetInterpolation(float fIndex, ezUInt32& out_uiIndexA, ezUInt32& out_uiIndexB, float& out_fFrac)
+  {
+    fIndex = ezMath::Saturate(fIndex);
+
+    constexpr ezUInt32 uiCountWithoutGray = Count - 1;
+    constexpr ezUInt32 uiMaxIndex = uiCountWithoutGray - 1;
+    out_uiIndexA = ezUInt32(fIndex * uiMaxIndex);
+    out_uiIndexB = out_uiIndexA + 1 % uiCountWithoutGray;
```
jankrassnigg: I think you need braces here, the modulo has precedence

jankrassnigg (approved):
How many documentation Screenshots do I now have to recreate??

Code/Engine/Foundation/Math/ColorScheme.h
C-Core: Good catch

C-Core:
> How many documentation Screenshots do I now have to recreate??

Most of the graph assets are not documented and I tried to keep the colors similar to the colors before, so I don't think you need to recreate any screenshots.

PR #790, opened 2022-11-03
Removed TypeSize from ezReflectedTypeDescriptor

jankrassnigg:
This value tends to change randomly (debug vs release builds?) and doesn't seem to have any actual use. Since it is serialized to the documents, it introduces pointless document modifications.


Code/Tools/Libs/ToolsFoundation/Reflection/ReflectedType.h
```diff
@@ -104,7 +104,6 @@ struct EZ_TOOLSFOUNDATION_DLL ezReflectedTypeDescriptor : public ezAttributeHold
   ezBitflags<ezTypeFlags> m_Flags;
   ezDynamicArray<ezReflectedPropertyDescriptor> m_Properties;
   ezDynamicArray<ezReflectedFunctionDescriptor> m_Functions;
-  ezUInt32 m_uiTypeSize;
   ezUInt32 m_uiTypeVersion;
```
C-Core: Can you initialize type version here so you don't need to fill this out if you don't care

Sanakan8472 (approved):

PR #791, opened 2022-11-03
Nicer pattern material and more color changes

C-Core:
* Nicer grid texture and pattern material that is used by graybox geometry by default.
* Grid color changes
* Property grid color changes
* Curve Editor color tweaks
* New fallback thumbnail images
* Fixed a few places where svg images were rasterized too soon which resulted in the wrong size and thus loss of quality.

image


Code/Tools/Libs/GuiFoundation/QtResources/resources.qrc
```diff
@@ -1,7 +1,7 @@
 <RCC>
   <qresource prefix="GuiFoundation">
-	<file>EZ-logo.svg</file>
-	<file>Icons/Add16.png</file>
+    <file>EZ-logo.svg</file>
```
jankrassnigg, line 3: Where are the SVGs from? Do we need to mention some license?

jankrassnigg (approved):

Code/Tools/Libs/GuiFoundation/QtResources/resources.qrc
C-Core: I made them myself

PR #792, opened 2022-11-09
State Machine improvements and unit tests

C-Core:
* Added a new timeout transition type that allows to automatically transition after the state machine has been in the source state for a certain amount of time.
* Added a compound transition type that allows to build complex conditions out of simple building blocks.
* Added an operator to the blackboard condition type to switch between AND and OR.
* Added a compound state type that similar to the compound transition type allows building complex states out of smaller building blocks.
* Added code unit tests that test the internals of the state machine and various state and transition types.
* Added a game engine unit tests that tests the state machine component and its interaction with other components (blackboard, script)
* Improved event message handling when a event message is sent during component initialization. It now ensures that potential event message handler components are also initialized at that point. Otherwise the message will be lost and the event message sender will cache a wrong result.

image


jankrassnigg:
We have optipng.exe checked in (precompiled tools). Please run the PNGs through that (just drop the file onto the exe). It will significantly reduce their file sizes.

C-Core:
> We have optipng.exe checked in (precompiled tools). Please run the PNGs through that (just drop the file onto the exe). It will significantly reduce their file sizes.

Good to know, will do once the linux build is fixed.

jankrassnigg (approved):

PR #794, opened 2022-11-14
Use a software rasterizer for occlusion culling

### LINKED ISSUE #621: [FEATURE] Use a software rasterizer to improve culling
See this repo:
https://github.com/rawrunprotected/rasterizer/tree/master/SoftwareRasterizer

It implements a very fast software rasterizer. The idea would be to use this for occlusion culling.

Basically, at the start of a frame, rasterize all occluders. For this one could use the greybox geometry.
Then during the culling phase that currently only uses frustum culling, use the bounding box or sphere of an object to check whether it is occluded and early out.

There are a couple of steps to achieve this:

The rasterizer is currently written as a demo application. It would need to be cleaned up and separated out into a stand-alone library. Probably not difficult though. If one were to attempt this, first contact the author and make sure to submit all changes back to that repository so that others benefit from that work as well.

AFAIK currently the library only rasterizes triangles. There is potential for more performance by rasterizing quads, boxes or spheres directly. Should be discussed with the author. Not important for a first integration though.

The library uses MSVC specific SSE instructions and thus only works on Windows ATM. Again, could be discussed with the author, whether there are ways to also bring this to other platforms. Again, not important, as we can easily disable the optimization where it's not available.

jankrassnigg:
Fixes #621 

Integrates [this high performance software rasterizer](https://github.com/rawrunprotected/rasterizer/tree/master/SoftwareRasterizer) into the render pipeline to enable occlusion culling.

There is an occluder component that can be used to place (invisible) box occluders in a scene. Additionally, the greybox geometry also adds occluders.

When a view is rendered, occluders up to a maximum distance are gathered and rendered to a low resolution image.
For scene rendering, this buffer can be queried, whether an area is fully occluded. If so, rendering of an object can be skipped. This usually results in a speedup of 20% to 50%.

Currently this rasterizer requires AVX1 and thus won't be available on all CPUs. If the CPU doesn't support AVX1, the feature is simply deactivated. Also it currently uses MSVC intrinsics and therefore is only available on Windows.

The feature is currently experimental. It may produce rendering artifacts (flickering of shadows, objects getting hidden that should already be visible). Some more tweaking may be necessary to figure out how to utilize it best.

As comparison, here is our Corridor scene without occlusion culling (walls set to transparent):

grafik

And with occlusion culling:

grafik

And this is the occlusion buffer generated from the greybox geometry:

grafik


Code/Engine/GameEngine/Gameplay/Implementation/GreyBoxComponent.cpp
```diff
@@ -385,8 +406,39 @@ void ezGreyBoxComponent::OnMsgExtractGeometry(ezMsgExtractGeometry& msg) const
...
+  if (!IsActiveAndInitialized() || !m_bUseAsOccluder)
+    return;
+
+  ezEnum<ezGreyBoxShape> shape = m_Shape;
+  if (shape == ezGreyBoxShape::StairsX && m_Curvature == ezAngle())
+    shape = ezGreyBoxShape::RampX;
+  if (shape == ezGreyBoxShape::StairsY && m_Curvature == ezAngle())
+    shape = ezGreyBoxShape::RampY;
+
+  if (m_pOccluderObject == nullptr)
```
C-Core, line 409: Thread safety?

Code/Engine/RendererCore/Components/Implementation/RenderComponent.cpp
```diff
@@ -43,10 +43,11 @@ void ezRenderComponent::OnUpdateLocalBounds(ezMsgUpdateLocalBounds& msg)
 
   bool bAlwaysVisible = false;
 
-  if (GetLocalBounds(bounds, bAlwaysVisible).Succeeded())
+  if (GetLocalBounds(bounds, bAlwaysVisible, msg).Succeeded())
   {
-    ezSpatialData::Category category =
-      GetOwner()->IsDynamic() ? ezDefaultSpatialDataCategories::RenderDynamic : ezDefaultSpatialDataCategories::RenderStatic;
+    // TODO: add occlusion category
+
```
C-Core: Todo is done

Code/Engine/GameEngine/Gameplay/Implementation/GreyBoxComponent.cpp
jankrassnigg: I frankly don't know. I was wondering about it but in my tests it never made problems. ¯\_(ツ)_/¯

Code/Engine/GameEngine/Gameplay/Implementation/GreyBoxComponent.cpp
```diff
@@ -2,6 +2,7 @@
 
 #include <Core/Graphics/Geometry.h>
 #include <Core/Interfaces/PhysicsWorldModule.h>
+#include <Core/Messages/TransformChangedMessage.h>
```
C-Core: include not needed anymore

Code/Engine/GameEngine/Gameplay/Implementation/GreyBoxComponent.cpp
```diff
@@ -318,7 +335,7 @@ void ezGreyBoxComponent::OnBuildStaticMesh(ezMsgBuildStaticMesh& msg) const
     return;
 
   ezGeometry geom;
-  BuildGeometry(geom);
+  BuildGeometry(geom, m_Shape, ezMat4::IdentityMatrix(), false);
```
C-Core: additional transform parameter can be removed again

Code/Engine/GameEngine/Gameplay/Implementation/GreyBoxComponent.cpp
C-Core: Allocate and build occlusion object in the same place as the render mesh to prevent race conditions. If two threads enter the branch at the same time the message will contain a dangling pointer.

Code/Engine/GameEngine/Gameplay/GreyBoxComponent.h
```diff
@@ -120,12 +124,15 @@ class EZ_GAMEENGINE_DLL ezGreyBoxComponent : public ezRenderComponent
...
   void InvalidateMesh();
-  void BuildGeometry(ezGeometry& geom) const;
+  void BuildGeometry(ezGeometry& geom, ezEnum<ezGreyBoxShape> shape, const ezMat4& transform, bool bOnlyRoughDetails) const;
 
   template <typename ResourceType>
   ezTypedResourceHandle<ResourceType> GenerateMesh() const;
 
   ezMeshResourceHandle m_hMesh;
+
+  mutable ezSharedPtr<ezRasterizerObject> m_pOccluderObject;
```
C-Core: Afaik there is nothing shared about this object, use unique ptr instead?

Code/Engine/RendererCore/Components/Implementation/OccluderComponent.cpp
```diff
@@ -0,0 +1,106 @@
...
+    msg.AddBounds(ezBoundingBox(-m_vExtents * 0.5f, m_vExtents * 0.5f), ezDefaultSpatialDataCategories::OcclusionStatic);
+  else
+    msg.AddBounds(ezBoundingBox(-m_vExtents * 0.5f, m_vExtents * 0.5f), ezDefaultSpatialDataCategories::OcclusionDynamic);
+}
+
+void ezOccluderComponent::OnMsgExtractOccluderData(ezMsgExtractOccluderData& msg) const
+{
+  if (IsActiveAndInitialized())
+  {
+    if (m_pOccluderObject == nullptr)
```
C-Core, line 68: Same race condition as in the gray box component

Code/Engine/RendererCore/Components/OccluderComponent.h
```diff
@@ -0,0 +1,54 @@
...
+  {
+    return m_vExtents;
+  }
+
+  void SetExtents(const ezVec3& extents);
+
+private:
+  ezVec3 m_vExtents = ezVec3(5.0f);
+
+  mutable ezSharedPtr<ezRasterizerObject> m_pOccluderObject;
```
C-Core: uniquePtr?

Code/Engine/RendererCore/Pipeline/Implementation/RenderPipeline.cpp
```diff
@@ -1242,4 +1323,150 @@ void ezRenderPipeline::CreateDgmlGraph(ezDGMLGraph& graph)
...
+  }
+
+  pRasterizer->EndScene();
+
+  return pRasterizer;
+}
+
+void ezRenderPipeline::PreviewOcclusionBuffer(const ezRasterizerView& rasterizer, const ezView& view)
+{
+  if (!cvar_SpatialCullingOcclusionVisView || !rasterizer.HasRasterizedAnyOccluders())
```
C-Core, line 1374: This function needs some serious cleanup at some point. I understand the problem and the solution is not so simple with our current infrastructure so we can leave it as is for now.

Code/Engine/RendererCore/Rasterizer/RasterizerView.h
```diff
@@ -0,0 +1,103 @@
...
+  bool m_bAnyOccludersRasterized = false;
+  const ezCamera* m_pCamera = nullptr;
+  ezUInt32 m_uiResolutionX = 0;
+  ezUInt32 m_uiResolutionY = 0;
+  float m_fAspectRation = 1.0f;
+  ezUniquePtr<Rasterizer> m_pRasterizer;
+
+  struct Instance
+  {
+    ezTransform m_Transform;
```
C-Core, line 80: would be better to use simd transform

Code/Engine/RendererCore/Pipeline/RenderData.h
```diff
@@ -129,4 +131,27 @@ struct EZ_RENDERERCORE_DLL ezMsgExtractRenderData : public ezMessage
...
+    d.m_Transform = transform;
+  }
+
+private:
+  friend class ezRenderPipeline;
+
+  struct Data
+  {
+    const ezRasterizerObject* m_pObject = nullptr;
+    ezTransform m_Transform;
```
C-Core, line 151: ezSimdTransform

C-Core (changes requested):

Code/Engine/GameEngine/Gameplay/GreyBoxComponent.h
jankrassnigg: It's pretty shared now :)

Code/Engine/RendererCore/Rasterizer/RasterizerView.h
jankrassnigg: I did originally, but in the end I have to multiply it with an ezMat4 and there is no conversion to that from ezSimdTransform, so passing an ezTransform is easier. And it's really not used for anything else.

Code/Engine/RendererCore/Pipeline/RenderData.h
jankrassnigg: same as above

Code/Engine/RendererCore/Pipeline/Implementation/RenderPipeline.cpp
jankrassnigg: I'm painfully aware of that.

Code/Engine/RendererCore/Components/Implementation/OccluderComponent.cpp
jankrassnigg: Since objects are shared now I have to use a mutex during mesh creation, which prevents a race condition. Please have another look. I can change it dispite that, but I don't really see a need for it now.

C-Core (approved):

PR #797, opened 2022-11-16
Several minor improvements

jankrassnigg:


Code/Engine/Foundation/Memory/FrameAllocator.h
```diff
@@ -37,3 +37,8 @@ class EZ_FOUNDATION_DLL ezFrameAllocator
 
   static ezDoubleBufferedStackAllocator* s_pAllocator;
 };
+
+struct ezFrameAllocatorWrapper
```
C-Core: This wrapper doesn't make any sense. The underlaying allocator is changed every frame but when using an allocator wrapper this function is only called once on container construction

PR #800, opened 2022-11-18
Vulkan window / swapchain fixes

Sanakan8472:
* Engine now asserts if a window is destroyed but still being used by a swapchain. While there is a reference count on the **ezWindowBase** class, it is not a reference counted object as we don't control the lifetime of the underlying native window in many cases. Thus, this is only for debugging for incorrect usage.
* Editor will now wait for the engine view to be destroyed (via **ezViewDestroyedResponseMsgToEditor**) before destroying a native window handle. Another case of use after free.
* **ezEngineProcessViewContext** will now update the exiting swapchain in place instead of recreating it on every resize. This fixes AMD giving up and just pinning the swapchain size to a random value and never updating it again when resizing in the editor.
* Added **ezView::ForceUpdate** as it is required for the editor render pipeline to be rebuilt on every viewport change but that is not something the **ezView** should do on its own.
* Swapchains can now be asked for the current resolution for convenience instead of asking their textures. This is required as the swapchain may be constraint to a different resolution than the window (e.g. no odd width).
* **ezWindowOutputTargetGAL::m_OnSwapChainChanged** will not fire the actual size change instead of the window's requested size change.
* Added **ezGALDevice::WaitIdle** to flush any usage of GPU resources that are in use / pending deleting. Required for the immediate destruction of a surfaces when a window gets deleted.
* Fixed a bug where **ezPipelineBarrierVulkan** would contain stale texture barriers for textures that were already deleted.
* Fixed a bug where the **ezPipelineBarrierVulkan** of the **ezInitContextVulkan** was not updated when a texture was deleted, potentially firing barriers for dead textures.
* ShaderExplorer now runs on linux.
* Ensured that ads is compiled with **Q_OS_LINUX** define on linux.
* Fixed wrong ezGALSamplerStateVulkan handling of anisotropic filtering. Fixing unit test image differences on AMD and greatly reducing the threshold for particle tests.
* Yet another change of ezImageCopyVulkan. If you make changes to the copy shaders and reload at runtime the old shader handles no longer resolve. To fix this, I simply test the shaders first and if any is nullptr, I delete the cash and rebuild it.


Code/Engine/RendererVulkan/Device/Implementation/SwapChainVulkan.cpp
```diff
@@ -347,7 +350,8 @@ ezResult ezGALSwapChainVulkan::DeInitPlatform(ezGALDevice* pDevice)
   DestroySwapChainInternal(pVulkanDevice);
   if (m_vulkanSurface)
   {
-    pVulkanDevice->DeleteLater(m_vulkanSurface);
+    ezVulkanAllocation context = reinterpret_cast<ezVulkanAllocation>(m_WindowDesc.m_pWindow);
```
Sanakan8472: I really shouldn't abuse the **ezVulkanAllocation** handle to piggy back my window ptr, will add a context in another push.

Ingrater (approved):

jankrassnigg (approved):

PR #801, opened 2022-11-19
Asset Background Processing Improvements

Sanakan8472:
* Pausing asset processing no longer blocks until all current active processes have finished. Instead, the button turns into a force stop button. If pressed the pending processes are terminated.
* **ezAssetCurator::TransformAsset** / **CreateThumbnail** now return a new type **ezTransformStatus**, which is like an **ezStatus** but the enum inside is extended to Success, Failure and NeedsImport.
* Transforming an asset in the background is now detectable via the **ezTransformFlags::BackgroundProcessing**. Assets that are transformed in teh background should never save themselves to disk and if ths is needed just return **ezTransformStatus::NeedsImport**. These will be retransformed on the main editor process to ensure proper tracking with no side-effects to the currently open documents.
* **AssetProcessor** no longer uses tasks but a single dedicated thread instead.
* Fixed bug where file added events were not fired.
* Fixed bug where thumbnails were genereated for assets that didn't have all their dependencies transformed yet.
* Fixed natvis file after recent refactorings.


Code/Engine/Foundation/ezEngine.natvis
```diff
@@ -180,14 +180,14 @@
...
-				<ValuePointer>m_ptr</ValuePointer>
+				<ValuePointer>m_pPtr</ValuePointer>
 			</ArrayItems>
 		</Expand>
 	</Type>
 
 	<!-- Basic Types -->
 	<Type Name="ezVariant">
-		<DisplayString Condition="m_Type == Type::TypedObject &amp;&amp; m_bIsShared">({m_Data.shared->m_pType->m_szTypeName,sb}*) {reinterpret_cast&lt;void*&gt;(m_Data.shared->m_Ptr)}</DisplayString>
+		<DisplayString Condition="m_Type == Type::TypedObject &amp;&amp; m_bIsShared">({m_Data.shared->m_pType->m_szTypeName,sb}*) {reinterpret_cast&lt;void*&gt;(m_Data.shared->m_pPtr)}</DisplayString>
```
C-Core: For the Variant shared data this is still called m_Ptr. @Ingrater is that something clang tidy missed? @Sanakan8472 Either revert the natvis change or fix it in ezVariant, either works for me.

Sanakan8472: Natvis reverted as other changes are already in dev.

Code/Editor/EditorFramework/Panels/AssetBrowserPanel/CuratorControl.cpp
```diff
@@ -157,7 +177,8 @@ void ezQtCuratorControl::AssetProcessorEvents(const ezAssetProcessorEvent& e)
   {
     case ezAssetProcessorEvent::Type::ProcessTaskStateChanged:
     {
-      UpdateBackgroundProcessState();
+      QMetaObject::invokeMethod(this, "UpdateBackgroundProcessState", Qt::QueuedConnection);
+      //UpdateBackgroundProcessState();
```
jankrassnigg: nit: cleanup

Code/Editor/EditorFramework/Assets/Declarations.h
```diff
@@ -82,3 +87,81 @@ struct EZ_EDITORFRAMEWORK_DLL ezAssetDocumentTypeDescriptor : public ezDocumentT
...
+  explicit ezTransformStatus(const ezFormatString& fmt)
+    : m_Result(ezTransformResult::Failure)
+  {
+    ezStringBuilder sMsg;
+    m_sMessage = fmt.GetText(sMsg);
+  }
+
+  EZ_ALWAYS_INLINE bool Succeeded() const { return m_Result == ezTransformResult::Success; }
+  EZ_ALWAYS_INLINE bool Failed() const { return m_Result == ezTransformResult::Failure; }
+  //void LogFailure(ezLogInterface* pLog = nullptr);
```
jankrassnigg: nit: cleanup

Code/Editor/EditorFramework/Assets/AssetProcessorMessages.h
```diff
@@ -0,0 +1,29 @@
...
+  ezDynamicArray<ezString> m_DepRefHull;
+};
+
+class EZ_EDITORFRAMEWORK_DLL ezProcessAssetResponseMsg : public ezProcessMessage
+{
+  EZ_ADD_DYNAMIC_REFLECTION(ezProcessAssetResponseMsg, ezProcessMessage);
+
+public:
+  ezTransformStatus m_Status;
+  mutable ezDynamicArray<ezLogEntry> m_LogEntries;
```
jankrassnigg, line 28: Why does this need to be mutable? If possible maybe not pass the message around as const?

Code/Editor/EditorFramework/Assets/Implementation/AssetCurator.cpp
```diff
@@ -149,8 +149,7 @@ void ezAssetCurator::StartInitialize(const ezApplicationFileSystemConfig& cfg)
 
   m_pWatcher = EZ_DEFAULT_NEW(ezAssetWatcher, m_FileSystemConfig);
 
-  ezSharedPtr<ezDelegateTask<void>> pInitTask = EZ_DEFAULT_NEW(ezDelegateTask<void>, "", [this]()
-    {
+  ezSharedPtr<ezDelegateTask<void>> pInitTask = EZ_DEFAULT_NEW(ezDelegateTask<void>, "", [this]() {
```
jankrassnigg: nit: Give the task a proper name?

Code/Editor/EditorFramework/Assets/Implementation/AssetBrowserWidget.cpp
```diff
@@ -696,8 +696,8 @@ void ezQtAssetBrowserWidget::OnTransform()
     ezUuid guid = m_pModel->data(index, ezQtAssetBrowserModel::UserRoles::AssetGuid).value<ezUuid>();
     QString sPath = m_pModel->data(index, ezQtAssetBrowserModel::UserRoles::RelativePath).toString();
     range.BeginNextStep(sPath.toUtf8());
-    auto res = ezAssetCurator::GetSingleton()->TransformAsset(guid, ezTransformFlags::TriggeredManually);
-    if (res.m_Result.Failed())
+    ezTransformStatus res = ezAssetCurator::GetSingleton()->TransformAsset(guid, ezTransformFlags::BackgroundProcessing);
```
jankrassnigg: Is this change intentional ?

Code/Editor/EditorFramework/Assets/Implementation/AssetBrowserView.cpp
```diff
@@ -236,6 +235,9 @@ void ezQtIconViewDelegate::paint(QPainter* painter, const QStyleOptionViewItem&
       case ezAssetInfo::TransformState::TransformError:
         ezQtUiServices::GetSingleton()->GetCachedIconResource(":/EditorFramework/Icons/AssetFailedTransform16.png").paint(painter, thumbnailRect);
         break;
+      case ezAssetInfo::TransformState::NeedsImport:
+        ezQtUiServices::GetSingleton()->GetCachedIconResource(":/EditorFramework/Icons/AssetFailedTransform16.png").paint(painter, thumbnailRect); // TODO
```
jankrassnigg: todo ?

jankrassnigg (approved):

Code/Editor/EditorFramework/Assets/AssetProcessorMessages.h
Sanakan8472: So when we receive the result message we don't want to copy all the log messages so we just swap our message array with the one from the message. It's not that simple to make the arriving message non-const as it weaves through all the IPC infrastructure. While in this case there is only one consumer, in the engine process case for example the same message might be read by multtiple parts of the program so in the initial design I thought it to be best if the messages are immutable.

Code/Editor/EditorFramework/Assets/Implementation/AssetBrowserWidget.cpp
Sanakan8472: First I tought I'd have to change it in order to make this consistent with the backgroundprocessing but it will only make things slower as we will do the same operation twice on the same thread for no good reason. Good catch.

Code/Editor/EditorFramework/Panels/AssetBrowserPanel/CuratorControl.cpp
Sanakan8472: Fixed

Code/Editor/EditorFramework/Assets/Implementation/AssetCurator.cpp
Sanakan8472: Done

Code/Editor/EditorFramework/Assets/Implementation/AssetBrowserView.cpp
Sanakan8472: Added a new image for this case (just needs transform in yellow).

PR #821, opened 2022-12-04
Added component to delay trigger events

jankrassnigg:


Code/Engine/Core/Messages/TriggerMessage.h
```diff
@@ -29,6 +29,8 @@ struct EZ_CORE_DLL ezMsgComponentInternalTrigger : public ezMessage
   /// Identifies what the message should trigger.
   ezHashedString m_sMessage;
 
+  ezInt32 m_iPayload = 0;
```
C-Core, line 32: maybe use ezVariant for the payload?

Code/Engine/Core/World/EventMessageHandlerComponent.h
```diff
@@ -6,14 +6,47 @@
 
 struct ezEventMessage;
 
+/// \brief Base class for ezEventMessageHandlerComponent. This can be used if global message handling is not needed and some options shall not show up in the UI.
+class EZ_CORE_DLL ezEventMessageHandlerBaseComponent : public ezComponent
```
C-Core: In our usual naming scheme we put Base at the end of the name, so this should be ezEventMessageHandlerComponent**Base**

Code/Engine/GameEngine/Gameplay/TriggerDelayModifierComponent.h
```diff
@@ -0,0 +1,58 @@
...
+  // ezTriggerDelayModifierComponent
+
+public:
+  ezTriggerDelayModifierComponent();
+  ~ezTriggerDelayModifierComponent();
+
+protected:
+  virtual void Initialize() override;
+
+  void OnMsgTriggerTriggered(ezMsgTriggerTriggered& msg) const;
```
C-Core: This message handler doesn't need to be const

Code/Engine/GameEngine/Gameplay/TriggerDelayModifierComponent.h
```diff
@@ -0,0 +1,58 @@
...
+  void OnMsgComponentInternalTrigger(ezMsgComponentInternalTrigger& msg);
+
+  mutable ezAtomicBool m_bIsActivated = false;
```
C-Core: no need for atomics since queued messages are always processed single threaded

Code/Engine/Core/World/EventMessageHandlerComponent.h
jankrassnigg: After the other refactor, we probably don't need this change anymore, anyway.

Code/Engine/GameEngine/Gameplay/Implementation/TriggerDelayModifierComponent.cpp
```diff
@@ -0,0 +1,158 @@
...
+      // the first object entered the trigger
+
+      if (!m_bIsActivated)
+      {
+        // the trigger is not active yet -> send an activation message with a new activation token
+
+        ++m_iValidActivationToken;
+
+        // store the original trigger message for later
+        m_sMessage = msg.m_sMessage;
```
C-Core, line 77: msg.m_hTriggeringObject is lost

jankrassnigg: Yes, because there is no 1:1 relationship anymore who entered/left a trigger.

C-Core: ah yes makes sense

Code/Engine/Core/World/Implementation/GameObject.cpp
```diff
@@ -895,51 +895,48 @@ void ezGameObject::PostMessageRecursive(const ezMessage& msg, ezTime delay, ezOb
...
 }
 
-void ezGameObject::SendEventMessage(ezEventMessage& msg, const ezComponent* pSenderComponent)
+void ezGameObject::SendEventMessage(ezMessage& msg, const ezComponent* pSenderComponent)
 {
-  ezHybridArray<ezComponent*, 4> eventMsgHandlers;
-  GetWorld()->FindEventMsgHandlers(msg, this, eventMsgHandlers);
-
-  if (eventMsgHandlers.IsEmpty() == false && pSenderComponent != nullptr)
+  if (auto pEventMsg = ezDynamicCast<ezEventMessage*>(&msg))
```
jankrassnigg, line 900: I thought we'd just have overloads for ezMessage and ezEventMessage, then you wouldn't need to do dynamic casts.

C-Core: I didn't want to add another variant of this method because we already have so many. The dynamic cast is super cheap in comparison to the whole search for a message handler, so we should be fine in this case.

PR #837, opened 2022-12-27
Global blackboards are now created on demand and stay alive forever

### LINKED ISSUE #830: Global blackboards should be created on-demand
Global blackboards that are created by a game state don't work as desired in play-the-game mode.

To make this work:
* Have factory method CreateBlackboard("name")
* GetOrCreateGlobalBlackboard("name")

In FindBlackboard("name") always call GetOrCreateGlobalBlackboard("name") if no other BB is found.
Then if the game state creates one, it already exists and references are set up correctly.

jankrassnigg:
Fixes #830.

### DIFF
--- a/Code/Engine/Core/Utils/Blackboard.h
+++ b/Code/Engine/Core/Utils/Blackboard.h
@@ -68,14 +68,32 @@ private:
   ezBlackboard();
 
 public:
+  ~ezBlackboard();
+
   /// \brief Factory method to create a new blackboard.
+  /// \brief Factory method to get access to a globally registered blackboard.
   ///
   /// Since blackboards use shared ownership we need to make sure that blackboards are created in ezCore.dll.
   /// Some compilers (MSVC) create local v-tables which can become stale if a blackboard was registered as global but the dll
   /// which created the blackboard is already unloaded.
   static ezSharedPtr<ezBlackboard> Create(ezAllocatorBase* pAllocator = ezFoundation::GetDefaultAllocator());
-  ~ezBlackboard();
 
+  /// If a blackboard with that name was already created globally before, its reference is returned.
+  /// Otherwise it will be created and permanently registered under that name.
+  /// Global blackboards cannot be removed. Although you can change their name via "SetName()",
+  /// the name under which they are registered globally will not change.
+  ///
+  /// If at some point you want to "remove" a global blackboard, instead call UnregisterAllEntries() to
+  /// clear all its values.
+  static ezSharedPtr<ezBlackboard> GetOrCreateGlobal(const ezHashedString& sBlackboardName, ezAllocatorBase* pAllocator = ezFoundation::GetDefaultAllocator());
+
+  /// \brief Finds a global blackboard with the given name.
+  static ezSharedPtr<ezBlackboard> FindGlobal(const ezTempHashedString& sBlackboardName);
+
+  /// \brief Changes the name of the blackboard.
+  ///
+  /// \note For global blackboards this has no effect under which name they are found. A global blackboard continues to
+  /// be found by the name under which it was originally registered.
   void SetName(const char* szName);
   const char* GetName() const { return m_sName; }
   const ezHashedString& GetNameHashed() const { return m_sName; }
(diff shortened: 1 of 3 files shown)


Code/Engine/Core/Utils/Blackboard.h
```diff
@@ -64,18 +64,28 @@ EZ_DECLARE_REFLECTABLE_TYPE(EZ_CORE_DLL, ezBlackboardEntryFlags);
...
-public:
-  /// \brief Factory method to create a new blackboard.
+  /// \brief Factory method to get access to a globally registered blackboard.
   ///
-  /// Since blackboards use shared ownership we need to make sure that blackboards are created in ezCore.dll.
-  /// Some compilers (MSVC) create local v-tables which can become stale if a blackboard was registered as global but the dll
-  /// which created the blackboard is already unloaded.
-  static ezSharedPtr<ezBlackboard> Create(ezAllocatorBase* pAllocator = ezFoundation::GetDefaultAllocator());
-  ~ezBlackboard();
+  /// If a blackboard with that name was already created globally before, its reference is returned.
```
C-Core, line 84: We still need a factory method for regular blackboards, otherwise you will run into the same problem as mentioned in the deleted comment if the plugin that created the blackboard is unloaded before the last reference.

Code/Engine/Core/Utils/Blackboard.h
```diff
@@ -64,18 +64,28 @@ EZ_DECLARE_REFLECTABLE_TYPE(EZ_CORE_DLL, ezBlackboardEntryFlags);
...
+  /// the name under which they are registered globally will not change.
+  ///
+  /// If at some point you want to "remove" a global blackboard, instead call UnregisterAllEntries() to
+  /// clear all its values.
+  static ezSharedPtr<ezBlackboard> GetOrCreateGlobal(const ezHashedString& sBlackboardName, ezAllocatorBase* pAllocator = ezFoundation::GetDefaultAllocator());
+
+  /// \brief Finds a global blackboard with the given name.
+  static ezSharedPtr<ezBlackboard> FindGlobal(const ezTempHashedString& sBlackboardName);
 
+  /// \brief Changes the name of the blackboard.
```
C-Core, line 96: I would remove the SetName method to avoid confusion. The name needs to be defined when the factory method is called.

C-Core (changes requested):

Code/Engine/Core/Utils/Blackboard.h
jankrassnigg: I had to keep it, because it wouldn't work with components during editing.

Code/Engine/Core/Utils/Blackboard.h
jankrassnigg: Okay

jankrassnigg: I don't quite get the problem though. The allocator is in foundation, the constructor is in core, how can it generate code elsewhere?

C-Core: The keyword is "local vtable": https://groups.google.com/g/microsoft.public.vc.language/c/atSh_2VSc2w/m/EgJ3r_7OzVUJ?pli=1

jankrassnigg: Just when you think you've seen it all...anyway, option 3 in his explanation would also be an option.

jankrassnigg: Also, I guess this affects ALL shared objects. It's just more likely for blackboards.

C-Core: Yes it would also be an option but we don't do that anywhere else in the codebase and factory methods are pretty common in ez. Also it would unify the creation of global and regular blackboards. 
It affects all shared objects where the code lives in one dll but the creation happens in a plugin that can be unloaded. If the creation and code is in the same dll this problem doesn't occur. Although you can still run into problems if that dll is unloaded and another still has a reference, but atm I can't think of another problematic case. Haven't looked at the code though.

jankrassnigg: Yes I actually agree that the creation method is the best solution. I'm wondering whether we should change this for all shared objects, just to be safe.

jankrassnigg: Done. How did you actually discover this issue? There are several other shared classes (like all GAL objects) and it's hard to say whether any of them could also run into this problem.

Code/Engine/Core/Utils/Blackboard.h
```diff
@@ -68,14 +68,32 @@ class EZ_CORE_DLL ezBlackboard : public ezRefCounted
   ezBlackboard();
 
 public:
+  ~ezBlackboard();
+
   /// \brief Factory method to create a new blackboard.
+  /// \brief Factory method to get access to a globally registered blackboard.
```
C-Core: Comment is in wrong place. Merge issue?

C-Core (approved):

PR #840, opened 2023-01-08
Asset Table Writer

### LINKED ISSUE #804: Editor should only reload changed resources
Every time an asset gets updated, we mark all resources for reload, which means the engine will check a lot of files, whether they have been changed.
Since we know which resource was affected, we should only reload that one.

Sanakan8472:
# Asset Table Writer
* Asset Table Writer has been moved out of **ezAssetCurator** into a dedicated class **ezAssetTableWriter**.
* ezAssetTableWriter keeps caches for all tables in the form of **ezAssetTable** instances. Dirty tables are flushes to disk on demand or after a certain amount of time.
* The caching only works as long as these assumptions remain true:
  * ezAssetDocumentManager::AddEntriesToAssetTable does never change over time.
  * ezAssetDocumentManager::GetAssetTableEntry never changes over the lifetime of an asset.
* ezAssetTableWriter also takes care of reloading modified resources. Instead of reloading all resources a new dedicated message is sent to only reload a single resource: **ezReloadResourceMsgToEngine**. Fixes #804
# Misc Changes
* Decals and other components that have arrays of assets will not have an auto-generated name in the scene graph referencing the first entry in the asset array.
* Added a Background flag to the editor to make clear that this instance of the EditorProcessor process is controlled by the editor and not standalone.
* Added non-template **ezResourceManager::ReloadResource** method.
* Added **ezResourceManager::BeginAcquireResourcePointer** and **EndAcquireResourcePointer**. This allows for non-template-based resource acquisition but only for the base resource pointer.


Code/Editor/EditorFramework/Assets/AssetCurator.h
```diff
@@ -258,6 +259,11 @@ class EZ_EDITORFRAMEWORK_DLL ezAssetCurator
   /// \brief Returns the table of all known assets in a locked structure
   const ezLockedSubAssetTable GetKnownSubAssets() const;
 
+  typedef ezLockedObject<ezMutex, const ezHashTable<ezUuid, ezAssetInfo*>> ezLockedAssetTable;
```
jankrassnigg: using instead of typedef

Code/Editor/EditorFramework/Assets/AssetTableWriter.h
```diff
@@ -0,0 +1,74 @@
...
+
+struct ezAssetCuratorEvent;
+class ezTask;
+struct ezAssetInfo;
+
+/// \brief Asset table class. Persistent cache for an asset table.
+///
+/// The following assumptions need to be true for this cache to work:
+/// 1. ezAssetDocumentManager::AddEntriesToAssetTable does never change over time
+/// 2. ezAssetDocumentManager::GetAssetTableEntry never changes over the lifetime of an asset.
```
jankrassnigg, line 19: I don't understand what this means. If I modify an assert (e.g. change a material to be a visual shader), wouldn't that modify the asset table? (because I then have additional outputs)

Code/Editor/EditorFramework/Assets/AssetTableWriter.h
```diff
@@ -0,0 +1,74 @@
...
+struct ezAssetTable
+{
+  ezString m_sDataDir;
+  ezString m_sTargetFile;
+  const ezPlatformProfile* m_pProfile;
```
jankrassnigg: = nullptr

Code/Editor/EditorFramework/Assets/AssetTableWriter.h
```diff
@@ -0,0 +1,74 @@
...
+  bool m_bDirty = true;
+  bool m_bReset = true;
+  ezMap<ezString, ezString> m_GuidToPath;
+
+  ezResult WriteAssetTable();
```
jankrassnigg, line 29: const

Code/Editor/EditorFramework/Assets/Implementation/AssetTableWriter.cpp
```diff
@@ -0,0 +1,324 @@
...
+  {
+    ezLog::Error("Failed to open asset lookup table file '{0}'", m_sTargetFile);
+    return EZ_FAILURE;
+  }
+
+  m_bDirty = false;
+  return EZ_SUCCESS;
+}
+
+void ezAssetTable::Remove(const ezSubAsset& subAsset, ezStringBuilder& sTemp)
```
jankrassnigg: passing in the temp string builder seems overly pessimistic

Code/Editor/EditorFramework/Document/Implementation/GameObjectDocument.cpp
```diff
@@ -301,11 +301,22 @@ void ezGameObjectDocument::DetermineNodeName(const ezDocumentObject* pObject, co
...
+          sValue = pChild->GetTypeAccessor().GetValue(pProperty->GetPropertyName()).ConvertTo<ezString>();
+        }
+        else if (pProperty->GetCategory() == ezPropertyCategory::Array)
+        {
+          const ezInt32 iCount = pChild->GetTypeAccessor().GetCount(pProperty->GetPropertyName());
+          if (iCount > 0)
+          {
+            sValue = pChild->GetTypeAccessor().GetValue(pProperty->GetPropertyName(), 0).ConvertTo<ezString>();
+          }
+        }
```
jankrassnigg, line 319: what is this change about ?

Code/Editor/EditorFramework/EditorApp/EditorApp.moc.h
```diff
@@ -120,6 +123,9 @@ class EZ_EDITORFRAMEWORK_DLL ezQtEditorApp : public QObject
   /// \brief Returns true if the editor is started is run in test mode.
   bool IsInUnitTestMode() const { return m_StartupFlags.IsSet(StartupFlags::UnitTest); }
 
+  /// \brief Returns true if the editor is started is run in background mode.
```
jankrassnigg: typo in comment

Code/Editor/EditorEngineProcess/EngineProcGameApp.cpp
```diff
@@ -301,7 +311,10 @@ void ezEngineProcessGameApplication::EventHandlerIPC(const ezEngineProcessCommun
     }
     else if (pMsg1->m_sWhatToDo == "ReloadResources")
     {
-      ezResourceManager::ReloadAllResources(false);
+      if (pMsg1->m_sPayload == "ReloadAllResources")
```
jankrassnigg, line 314: Why is the DeleteAllCachedRenderData only in the one case? Are there other cases, or should there be an assert_not_implemented in the else case ?

jankrassnigg (approved):

Code/Editor/EditorFramework/Assets/AssetTableWriter.h
Sanakan8472: For each sub-asset we only write the main output (the one without any tags) into the asset table and the assumption is, that this output always points to the same path. In the case of the material, it will always be the ezMaterialBin regardless of how the material is generated.

Code/Editor/EditorFramework/Document/Implementation/GameObjectDocument.cpp
Sanakan8472: "Decals and other components that have arrays of assets will not have an auto-generated name in the scene graph referencing the first entry in the asset array."
Basically as the decal uses an array of assets it was not picked up so all decals were called "Decal". With this change the name now looks like this:
image

Code/Editor/EditorEngineProcess/EngineProcGameApp.cpp
Sanakan8472: So if you click "reload resources" in the menu, it will set the payload to "ReloadAllResources" and behave as before, i.e. calling:
```
ezResourceManager::ReloadAllResources(false);
ezRenderWorld::DeleteAllCachedRenderData();
```
Now if I incrementally reload resources in the background I still need to invalidate the cached render data at some point. I didn't want to do that on every resource so instead I fire the "ReloadResources" message but without the payload. Not sure if that is the ideal process of dealing with this problem.

Code/Editor/EditorFramework/Assets/AssetTableWriter.h
Sanakan8472: WriteAssetTable is not const as it will actually re-generate itself if `m_bReset` is true.

PR #850, opened 2023-01-15
ADS fixes

### LINKED ISSUE #532: [BUG] Regression: Document shortcuts don't work when switching tabs
CTRL+W should close a document. However, when switching between documents, it often doesn't work (for example closing multiple documents one after the other with the shortcut).

It only starts working after you once click into the document itself (not the tab at the top).

@Sanakan8472 I assume this is a side-effect of the shortcut change in #516

Sanakan8472:
* Closing a document will set the keyboard focus to the new window that is on top of the dock area (fixes #532).
* Restored ads::CDockManager::ConfigFlags which were changed to being static with some update.
* Enabled FocusHighlighting in ADS and tweaked the css style.
* Fixed compile error due to precision loss.

### DIFF
--- a/Code/ThirdParty/ads/stylesheets/focus_highlighting.css
+++ b/Code/ThirdParty/ads/stylesheets/focus_highlighting.css
@@ -2,7 +2,7 @@
  * Default style sheet on Windows Platforms with focus highlighting flag enabled
  */
 ads--CDockContainerWidget {
-	background: palette(dark);
+	background: palette(window);
 }
 ads--CDockContainerWidget > QSplitter{
         padding: 1 0 1 0;
@@ -10,7 +10,7 @@ ads--CDockContainerWidget > QSplitter{
 
 ads--CDockAreaWidget {
 	background: palette(window);
-	border: 1px solid white;
+	border: 1px solid black;
 }
 
 ads--CDockWidgetTab {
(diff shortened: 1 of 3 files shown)


Code/Engine/GameEngine/Effects/Shake/Implementation/CameraShakeComponent.cpp
```diff
@@ -67,7 +67,7 @@ void ezCameraShakeComponent::GenerateKeyframe()
 
   if (deviation > ezAngle())
   {
-    m_Rotation += ezAngle::Radian(pWorld->GetRandomNumberGenerator().DoubleMinMax(ezAngle::Degree(120).GetRadian(), ezAngle::Degree(240).GetRadian()));
```
jankrassnigg: I just relocated this file in my PR, so I have integrated your fix there. In case my PR goes in first, you can just remove this change from your PR.

jankrassnigg (approved):

Code/Engine/GameEngine/Effects/Shake/Implementation/CameraShakeComponent.cpp
Sanakan8472: Ok, I'll removed this change then.

PR #854, opened 2023-01-16
Added GameObject parent changed message

jankrassnigg:


Code/Engine/Core/World/Implementation/World.cpp
```diff
@@ -673,6 +682,15 @@ void ezWorld::UnlinkFromParent(ezGameObject* pObject)
 {
   if (ezGameObject* pParentObject = pObject->GetParent())
   {
+    if (pObject->m_Flags.IsSet(ezObjectFlags::ParentChangesNotifications))
```
C-Core: I think this message should be send after the parent unlinking? Same place as we send the ChildRemoved message below.

C-Core (approved):

PR #866, opened 2023-02-02
Fix decal asset resource reloading

### LINKED ISSUE #860: [BUG] Editing decal assets broken due to resource reloading change
If you change the texture that a decal resource references, the change is not visible in engine, until you execute "Reload Resources" from the menu.

1. In ezAssetTableWriter::MainThreadTick m_bNeedToReloadResources will be true, but m_GuidToPath.Contains fails, because the decal GUID is not part of the asset table.

2. In EngineProcGameApp.cpp -> "ReloadResource" (line 284) GetExistingResourceByType succeeds, but it is not enough. 
    ezResourceManager::ReloadAllResources(false); is also needed.
    Also that block should probably call ezRenderWorld::DeleteAllCachedRenderData(); after the reload.

Sanakan8472:
* Fixes #860
* **ezAssetDocumentManager::AddEntriesToAssetTable** now uses a callback instead of writing to a passed in container and also provides the resource type of the added entries.
* ezAssetTable keeps a copy of all manager resources provided by the above function around.
* If an asset is not represented by a resource in the ezAssetTable we assume it is represented by a manager resource. Currently we don't know how these relate, e.g. we don't know all "Decal" assets are represented by the "{ ProjectDecalAtlas }" resource. Therefore, we just reload all manager resources.
* Added `ezResourceManager::RegisterResourceForAssetType("Decal Atlas", ezGetStaticRTTI<ezDecalAtlasResource>());` so that decal atlas resources can be reloaded without knowing the type.


Code/Editor/EditorFramework/Assets/Implementation/AssetTableWriter.cpp
```diff
@@ -53,24 +52,21 @@ ezResult ezAssetTable::WriteAssetTable()
...
-  auto Write = [](ezMap<ezString, ezString>::ConstIterator it, ezDeferredFileWriter& file) {
-    const ezString& guid = it.Key();
-    const ezString& path = it.Value();
-
-    file.WriteBytes(guid.GetData(), guid.GetElementCount()).IgnoreResult();
+  auto Write = [](const ezString& sGuid, const ezString& sPath, ezDeferredFileWriter& file) {
+    file.WriteBytes(sGuid.GetData(), sGuid.GetElementCount()).IgnoreResult();
     file.WriteBytes(";", 1).IgnoreResult();
-    file.WriteBytes(path.GetData(), path.GetElementCount()).IgnoreResult();
+    file.WriteBytes(sPath.GetData(), sPath.GetElementCount()).IgnoreResult();
```
jankrassnigg, line 58: btw. most "IgnoreResult" calls should probably get changed to "AssertSuccess"

Not needed here, just saying, that it could be done.

jankrassnigg (approved):

Issue #580, opened 2022-05-01
[BUG] Editor Crashes when a surface is selected as itself for BaseMaterial


exoad:
**Bug description:**
The Editor refuses to launch and crashes if a Surface has its BaseMaterial Selected as itself.

**How to Reproduce:**
A clear description what steps to take to reproduce the issue.
1. Do this...
1. Do that...
1. etc...
Create a new Surface, then set the BaseMaterial property to itself.

I am on Windows 10 (x64)

exoad:
Ok I was able to kind of fix this by deleting / renaming / moving my current project directory

jankrassnigg:
Reopening, because this definitely needs a proper fix.

exoad:
> Reopening, because this definitely needs a proper fix.

Were you able to reproduce this?

jankrassnigg:
Unfortunately this can still happen.

Surfaces use only ezDependencyFlags::Package when referencing their base surface. Apparently this is not checked for circular dependencies and therefore it is possible to select something as its own base.

Materials on the other hand also use ezDependencyFlags::Thumbnail, which apparently prevents building circular dependencies.

PR #896, opened 2023-03-19
# Asset Dependency Overhaul

### LINKED ISSUE #814: Refactor asset dependency handling
There are currently two types of asset dependencies:
1. TransformDependencies (TD)
2. RuntimeDependencies (RD)

A property with ezFileBrowserAttribute is automatically added as a TD, a property with an ezAssetBrowserAttribute is added as an RD.

As far as I can tell, these are the rules:

1. If the value of any property changes, an asset is retransformed (and a new thumbnail is always generated ?).
2. If a file or asset changes that is referenced as a TD, the referencing asset is re-transformed (and a new thumbnail is always generated ?).
3. If a file or asset changes that is referenced as a RD, the referencing asset is NOT transformed, but it's thumbnail is regenerated, unless, of course, it doesn't have a thumbnail at all.

So in essence, if an input file changes (jpg, fbx) the asset that is built from that is re-transformed and all assets that transitively reference that asset, get a new thumbnail. If any property value of an asset changes, the same thing happens.

These rules are in most cases too pessimistic and way too many thumbnails may get updated unnecessarily, while in other cases a necessary re-transform won't be triggered.

I would argue that references have these properties
* Visual (V) vs Non-Visual (NV)
* Not-Affecting (NA) vs Directly-Affecting (DA) vs Transitively-Affecting (TA)

Consider textures, materials, surfaces, meshes, prefabs and collections.

The input file on a texture asset (PNG) is Visual and Directly-Affecting. Any change to the input file must trigger a re-transform and a thumbnail update.

The texture reference on a material is V and NA. If the texture asset changes visually, the material thumbnail must be updated as well, but the material itself does not need to be re-transformed.

The surface reference on a material is NV and NA. Any change to a surface won't trigger a re-transform or a thumbnail update on the material.

Surfaces only reference prefabs, all these references are NV and NA. Changes to the prefabs never need to trigger a re-transform of the surface (and it doesn't have a thumbnail anyway).

Collections may reference ANY other asset type. These references are NV but transitively-affecting (TA). Any change to any directly or transitively referenced asset must mark the collection as out-of-date. That's because the collections gather the transitive hull of references. So if a material changes what texture it uses, then a referencing mesh only needs to update its thumbnail, but a referencing collection must be re-transformed, to store the updated set of references.

I would propose to configure these via an attribute, so that for each property I can configure how that property affects the asset state. If such an attribute is not present, we determine a useful default from the ezFileBrowserAttribute and ezAssetBrowserAttribute as before. However, now I can for example use the new attribute on the surface reference of a material to make this a non-visual reference, and therefore prevent thumbnail updates when a surface is modified.

Similarly, we could add information to asset types. So for example for a collision mesh asset I could specify that by default, it is not visual, meaning if I change a collision mesh, I don't need to update the thumbnails of prefabs. This would just be for convenience, though, so that I don't need to repeat the same information through the new attribute over and over. E.g. when we fine an ezAssetBrowserAttribute, but not the more explicit attribute, we then ask the respective asset manager, what the reference would affect.

For tracking, we already have an asset hash and a thumbnail hash and AFAIK we chain those hashes for TDs and RDs. Instead we should chain them depending on the visual vs non-visual and x-affecting property. To get the transitive-affecting thing working, we probably need a third hash that tracks whether ANY property along an asset-reference chain was modified, even if none of them were DA or V.

Sanakan8472:
## Asset Package References
* fixes #814 
* Added **ezDependencyFlags** which defines how a reference in an asset is used:
  *  **Thumbnail**: This reference is a dependency to generating a thumbnail.
  *  **Transform**: This reference is a dependency to transforming this asset.
  *  **Package**: This reference is needs to be packaged as it is used at runtime by this asset.
* **ezAssetBrowserAttribute** now has a new parameter to set the `ezDependencyFlags` for it. By default, **ezDependencyFlags::Thumbnail | ezDependencyFlags::Package** is set.
* Fixed naming of various enums and member variables to match the new `transform`, `thumbnail` and `package` dependency naming pattern.

## Circular Dependencies
* Circular dependencies will no longer deadlock the editor and instead result in a new error code on one of the assets in the circle: **ezAssetInfo::TransformState::CircularDependency**.
* **ezAssetCurator::GenerateTransitiveHull** is no longer recursive and can handle circular dependencies.
* Added **ezAssetCurator::GenerateInverseTransitiveHull** which generates a set of all assets that depend in the given asset. Used by the new **ezAssetCurator::CheckForCircularDependencies** function.
* **ezAssetCurator::InvalidateAssetTransformState** is no longer recursive, using **GenerateInverseTransitiveHull** instead.
* **ezQtAssetPropertyWidget::InternalSetValue** will prevent the user now from creating circular dependencies.

## Misc
* Added **Asset.WriteDependencyDGML** action which creates a DGML file containing all transform and thumbnail dependencies this asset transitively depends on. Very useful to understand circular dependencies.
* Added **ezAssetActions::MapMenuActions** to put `Transform Asset` and `Write Dependency DGML` into the edit menu.
* **ezCollectionAssetDocument** are no longer updated automatically as their transitive naturure can't be tracked correctly by the currently defined `ezDependencyFlags`. Instead, they are set to manual transform and will be auto-generated on every scene export.

### DIFF
--- a/Code/Editor/EditorFramework/Assets/AssetCurator.h
+++ b/Code/Editor/EditorFramework/Assets/AssetCurator.h
@@ -67,8 +67,9 @@ struct EZ_EDITORFRAMEWORK_DLL ezAssetInfo
     NeedsTransform,
     NeedsThumbnail,
     TransformError,
-    MissingDependency,
-    MissingReference,
+    MissingTransformDependency,
+    MissingThumbnailDependency,
+    CircularDependency,
     COUNT,
   };
 
@@ -87,8 +88,9 @@ struct EZ_EDITORFRAMEWORK_DLL ezAssetInfo
 
   ezUniquePtr<ezAssetDocumentInfo> m_Info;
 
-  ezSet<ezString> m_MissingDependencies;
-  ezSet<ezString> m_MissingReferences;
+  ezSet<ezString> m_MissingTransformDeps;
+  ezSet<ezString> m_MissingThumbnailDeps;
+  ezSet<ezString> m_CircularDependencies;
 
   ezSet<ezUuid> m_SubAssets; ///< Main asset uses the same GUID as this (see m_Info), but is NOT stored in m_SubAssets
 
@@ -270,8 +272,6 @@ public:
   /// \brief Computes the combined hash for the asset and its references. Returns 0 if anything went wrong.
   ezUInt64 GetAssetReferenceHash(ezUuid assetGuid);
 
-  void GenerateTransitiveHull(const ezStringView sAssetOrPath, ezSet<ezString>* pDependencies, ezSet<ezString>* pReferences);
-
   ezAssetInfo::TransformState IsAssetUpToDate(const ezUuid& assetGuid, const ezPlatformProfile* pAssetProfile, const ezAssetDocumentTypeDescriptor* pTypeDescriptor, ezUInt64& out_uiAssetHash, ezUInt64& out_uiThumbHash, bool bForce = false);
   /// \brief Returns the number of assets in the system and how many are in what transform state
   void GetAssetTransformStats(ezUInt32& out_uiNumAssets, ezHybridArray<ezUInt32, ezAssetInfo::TransformState::COUNT>& out_count);
@@ -316,9 +316,23 @@ public:
 
   void NeedsReloadResources(const ezUuid& assetGuid);
 
+  void InvalidateAssetsWithTransformState(ezAssetInfo::TransformState state);
+
   ///@}
 
-  void InvalidateAssetsWithTransformState(ezAssetInfo::TransformState state);
+  /// \name Utilities
+  ///@{
+
+  /// \brief Generates one transitive hull for all the dependencies that are enabled. The set will contain dependencies that are reachable via any combination of enabled reference types.
+  void GenerateTransitiveHull(const ezStringView sAssetOrPath, ezSet<ezString>& deps, bool bIncludeTransformDebs = false, bool bIncludeThumbnailDebs = false, bool bIncludePackageDebs = false) const;
+
+  /// \brief Generates one transitive hull for all the dependencies that are enabled. The set will contain dependencies that are reachable via any combination of enabled reference types. As only assets can have dependencies, the inverse hull is always just asset GUIDs.
+  void GenerateInverseTransitiveHull(const ezAssetInfo* pAssetInfo, ezSet<ezUuid>& inverseDeps, bool bIncludeTransformDebs = false, bool bIncludeThumbnailDebs = false) const;
+
+  /// \brief Generates a DGML graph of all transform and thumbnail dependencies.
+  void WriteDependencyDGML(const ezUuid& guid, ezStringView sOutputFile) const;
+
+  ///@}
 
 public:
   ezEvent<const ezAssetCuratorEvent&> m_Events;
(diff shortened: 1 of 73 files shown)


Code/Editor/EditorFramework/Assets/AssetCurator.h
```diff
@@ -316,9 +316,23 @@ class EZ_EDITORFRAMEWORK_DLL ezAssetCurator
...
+  void InvalidateAssetsWithTransformState(ezAssetInfo::TransformState state);
+
   ///@}
 
-  void InvalidateAssetsWithTransformState(ezAssetInfo::TransformState state);
+  /// \name Utilities
+  ///@{
+
+  /// \brief Generates one transitive hull for all the dependencies that are enabled. The set will contain dependencies that are reachable via any combination of enabled reference types.
+  void GenerateTransitiveHull(const ezStringView sAssetOrPath, ezSet<ezString>& deps, bool bIncludeTransformDebs = false, bool bIncludeThumbnailDebs = false, bool bIncludePackageDebs = false) const;
```
jankrassnigg: nit: "Deps" not "Debs"

Code/Editor/EditorFramework/Assets/AssetCurator.h
```diff
@@ -316,9 +316,23 @@ class EZ_EDITORFRAMEWORK_DLL ezAssetCurator
...
+
+  /// \brief Generates one transitive hull for all the dependencies that are enabled. The set will contain dependencies that are reachable via any combination of enabled reference types. As only assets can have dependencies, the inverse hull is always just asset GUIDs.
```
jankrassnigg: The comment is mostly identical to the one above, which doesn't make sense to me.

Code/Editor/EditorFramework/Assets/Implementation/AssetUpdates.cpp
```diff
@@ -321,39 +327,75 @@ ezResult ezAssetCurator::EnsureAssetInfoUpdated(const char* szAbsFilePath)
...
   for (auto outputIt = pAssetInfo->m_Info->m_Outputs.GetIterator(); outputIt.IsValid(); ++outputIt)
   {
     const ezString sTargetFile2 = pAssetInfo->GetManager()->GetAbsoluteOutputFileName(pAssetInfo->m_pDocumentTypeDescriptor, pAssetInfo->m_sAbsolutePath, outputIt.Key());
-    it = m_InverseReferences.FindOrAdd(sTargetFile2);
+    it = m_InverseThumbnailDeps.FindOrAdd(sTargetFile2);
     it.Value().RemoveAndCopy(pAssetInfo->m_Info->m_DocumentID);
   }
 }
 
+ezResult ezAssetCurator::CheckForCircularDependencies(ezAssetInfo* pAssetInfo)
```
jankrassnigg, line 365: can the pointer be const or even const& ?

Code/Editor/EditorFramework/Assets/Implementation/AssetWatcher.cpp
```diff
@@ -258,7 +258,7 @@ void ezDirectoryUpdateTask::Execute()
   {
     CURATOR_PROFILE("FindReferencedFiles");
     // Find all currently known files that are under the given folder.
-    // TODO: is m_InverseDependency / m_InverseReferences covered by this?
+    // TODO: is m_InverseTransformDeps / m_InverseThumbnailDeps covered by this?
```
jankrassnigg: is this comment still relevant ?

Code/EditorPlugins/Scene/EditorPluginScene/Actions/SceneActions.cpp
```diff
@@ -332,6 +332,33 @@ void ezSceneAction::Execute(const ezVariant& value)
...
+        EZ_PROFILE_SCOPE("ConvertCollections");
+        ezAssetCurator* pCurator = ezAssetCurator::GetSingleton();
+        ezSet<ezUuid> collections;
+        {
+          ezAssetCurator::ezLockedAssetTable allAssets = pCurator->GetKnownAssets();
+
+          ezTempHashedString sCollection = "Collection";
+          for (auto it : *allAssets)
+          {
+            if (it.Value()->m_Info->m_sAssetsDocumentTypeName == sCollection)
```
jankrassnigg, line 348: Instead of hardcoding this for collections, we could also add a virtual function "NeedsTransformOnSceneExport". Then in the future other assets could also utilize this.
Not important though, can also be changed when the need really arises.

Code/Engine/Foundation/Reflection/Implementation/PropertyAttributes.h
```diff
@@ -336,6 +336,31 @@ class EZ_FOUNDATION_DLL ezConstrainPointerAttribute : public ezPropertyAttribute
   ezUntrackedString m_sConstantValueProperty;
 };
 
+/// \brief Defines how a reference set by ezFileBrowserAttribute is treated.
```
jankrassnigg: not only ezFileBrowserAttribute but also ezAssetBrowserAttribute

Code/Engine/Foundation/Reflection/Implementation/PropertyAttributes.h
```diff
@@ -336,6 +336,31 @@ class EZ_FOUNDATION_DLL ezConstrainPointerAttribute : public ezPropertyAttribute
...
+struct ezDependencyFlags
+{
+  typedef ezUInt8 StorageType;
+
+  enum Enum
+  {
+    None = 0,
+    Thumbnail = EZ_BIT(0), ///< This reference is a dependency to generating a thumbnail.
+    Transform = EZ_BIT(1), ///< This reference is a dependency to transforming this asset.
+    Package = EZ_BIT(2),   ///< This reference is needs to be packaged as it is used at runtime by this asset.
```
jankrassnigg: These are quite important to understand to use correctly, I think we should document this a bit more thoroughly. Also the None case. Maybe with a concrete example for each.

jankrassnigg (approved):

jankrassnigg:
This also fixes #580.

Code/Editor/EditorFramework/Assets/Implementation/AssetUpdates.cpp
Sanakan8472: No it has to write to the **m_CircularDependencies** member.

Code/Editor/EditorFramework/Assets/Implementation/AssetWatcher.cpp
Sanakan8472: Nope, I don't think they are relevant to the current code anymore.

Code/Engine/Foundation/Reflection/Implementation/PropertyAttributes.h
Sanakan8472: I've added a few examples to each element and to the class itself.

Code/EditorPlugins/Scene/EditorPluginScene/Actions/SceneActions.cpp
Sanakan8472: Added an #TODO_ASSET to this code for future reference.

jankrassnigg (approved):

PR #907, opened 2023-03-27
Fix #902: Compute Shader Histogram Duplicate Windows.

### LINKED ISSUE #902: [BUG] ComputeShader sample fails to launch
It complains about having two input systems (or two windows).

TheophilusE:
Fixed the Compute Shader Histogram sample recreating the main window and output target.


jankrassnigg:
Thank you very much for the fix :)

PR #939, opened 2023-05-11
Fixes and improvements for the C++ workflow

### LINKED ISSUE #817: [BUG] Editor sometimes removes plugin references
If you load the sample game project without having the DLL compiled, the editor removes the plugin reference from the ddl file.
This may also happen to other things under certain circumstances.
This should not happen.

jankrassnigg:
* Fixed #817: Project specific plugin bundles don't need to be copied to the Bin folder anymore. This prevents plugin references getting lost when switching between build types.
* When building the C++ code, the CMake solution is now automatically regenerated when the build type differs (Debug, Dev, Shipping).
* The editor warns now that the build may fail, when a debugger is attached.

### DIFF
--- a/Code/Engine/Foundation/IO/FileSystem/DeferredFileWriter.h
+++ b/Code/Engine/Foundation/IO/FileSystem/DeferredFileWriter.h
@@ -22,7 +22,7 @@ public:
 
   /// \brief Upon calling this the content is written to the file specified with SetOutput().
   /// The return value is EZ_FAILURE if the file could not be opened or not completely written.
-  ezResult Close(); // [tested]
+  ezResult Close(bool* out_bWasWrittenTo = nullptr); // [tested]
 
   /// \brief Calling this abandons the content and a later Close or destruction of the instance
   /// will no longer write anything to file.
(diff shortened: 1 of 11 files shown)


Code/Engine/Foundation/IO/FileSystem/DeferredFileWriter.h
```diff
@@ -22,7 +22,7 @@ class EZ_FOUNDATION_DLL ezDeferredFileWriter : public ezStreamWriter
 
   /// \brief Upon calling this the content is written to the file specified with SetOutput().
   /// The return value is EZ_FAILURE if the file could not be opened or not completely written.
-  ezResult Close(); // [tested]
+  ezResult Close(bool* out_bWasWrittenTo = nullptr); // [tested]
```
C-Core, line 25: Isn't out_bWasWrittenTo the same as the returned ezResult?

jankrassnigg: No, if the content is identical, the result will be true, but written to will be false.

PR #943, opened 2023-05-15
Fix some issues found by PVS-Studio in Foundation and Core 

dimhotepus:
Product page: https://www.viva64.com/en/pvs-studio/
How to use for free: https://www.viva64.com/en/b/0600/


jankrassnigg (commented):
Very nice, thanks a lot! I'll take a closer look later. From a rough glance there are definitely a few important fixes in there.

jankrassnigg:
I've made some cosmetic changes (mostly ran Clang Format) and I changed the ezTaskWorkerInfo layout some more, so that it is now 24 bytes on x64.

jankrassnigg (approved):

PR #945, opened 2023-05-18
PVS-Studio fixes for Renderer{Foundation | Core}

dimhotepus:
Product page: https://www.viva64.com/en/pvs-studio/
How to use for free: https://www.viva64.com/en/b/0600/

### DIFF
--- a/Code/Engine/RendererCore/AnimationSystem/AnimGraph/Implementation/AnimGraphPins.cpp
+++ b/Code/Engine/RendererCore/AnimationSystem/AnimGraph/Implementation/AnimGraphPins.cpp
@@ -57,7 +57,7 @@ void ezAnimGraphTriggerOutputPin::SetTriggered(ezAnimGraph& ref_graph, bool bTri
   const auto& map = ref_graph.m_OutputPinToInputPinMapping[ezAnimGraphPin::Trigger][m_iPinIndex];
 
 
-  const ezInt8 offset = bTriggered ? +1 : -1;
+  const ezInt8 offset = +1; // bTriggered ? +1 : -1;
 
   // trigger or reset all input pins that are connected to this output pin
   for (ezUInt16 idx : map)
--- a/Code/Engine/RendererCore/Meshes/Implementation/MeshBufferResource.cpp
+++ b/Code/Engine/RendererCore/Meshes/Implementation/MeshBufferResource.cpp
@@ -503,11 +503,6 @@ ezResult ezMeshBufferResourceDescriptor::RecomputeNormals()
 //////////////////////////////////////////////////////////////////////////
 //////////////////////////////////////////////////////////////////////////
 
-ezMeshBufferResource::ezMeshBufferResource()
-  : ezResource(DoUpdate::OnAnyThread, 1)
-{
-}
-
 ezMeshBufferResource::~ezMeshBufferResource()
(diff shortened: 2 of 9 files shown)


dimhotepus:
https://github.com/ezEngine/ezEngine/pull/945/commits/9a455d4eb70383e6aadedbf898bf09e3ffc8a5a8

^ This one requires attention. Looks like it fixes regression from 07ae33724330e6c26f3ee1de88eb75ea1ce84524.

Code/Engine/RendererCore/Meshes/Implementation/MeshBufferResource.cpp
```diff
@@ -505,6 +505,8 @@ ezResult ezMeshBufferResourceDescriptor::RecomputeNormals()
 
 ezMeshBufferResource::ezMeshBufferResource()
   : ezResource(DoUpdate::OnAnyThread, 1)
+  , m_uiPrimitiveCount(0)
```
C-Core: Would be even better to initialize in the header directly

C-Core (approved):

jankrassnigg:
> [9a455d4](https://github.com/ezEngine/ezEngine/commit/9a455d4eb70383e6aadedbf898bf09e3ffc8a5a8)
> 
> ^ This one requires attention. Looks like it fixes regression from [07ae337](https://github.com/ezEngine/ezEngine/commit/07ae33724330e6c26f3ee1de88eb75ea1ce84524).

Thanks for pointing it out. I'll need to think about this carefully. Not sure whether I'll have time for that today.

Code/Engine/RendererCore/Meshes/Implementation/MeshBufferResource.cpp
jankrassnigg: Indeed for constants that would be nicer. If you have the time to change that, please do, though I would accept the PR this way as well.

jankrassnigg: I wonder whether this is something that we could use Clang-tidy for? @Ingrater ?

Code/Engine/RendererCore/AnimationSystem/AnimGraph/Implementation/AnimGraphPins.cpp
```diff
@@ -51,12 +51,8 @@ void ezAnimGraphTriggerOutputPin::SetTriggered(ezAnimGraph& ref_graph, bool bTri
   if (m_iPinIndex < 0)
     return;
 
-  if (!bTriggered)
```
jankrassnigg: Looking at the rest of the logic, the current code is correct, and this change would break it.
There was a design change at some point, which meant that this code works differently now than originally.

So to get rid of the warning, you can change this line:
const ezInt8 offset = +1; // bTriggered ? +1 : -1;

but keep the if (!bTriggered) return;

It is unclear whether this piece of code will stay that way in the long run, but for now that's good enough.

Code/Engine/RendererCore/Meshes/Implementation/MeshBufferResource.cpp
dimhotepus: I've initialized class members directly in header.

PR #950, opened 2023-05-19
clang-tidy modernize use default member init

Ingrater:


Code/Editor/EditorEngineProcessFramework/EngineProcess/EngineProcessMessages.h
```diff
@@ -299,14 +299,13 @@ class EZ_EDITORENGINEPROCESSFRAMEWORK_DLL ezExportDocumentMsgToEngine : public e
 
 public:
   ezExportDocumentMsgToEngine()
```
jankrassnigg, line 301: Is there also a check that can remove empty constructors ?

Code/Engine/RendererCore/Pipeline/Implementation/Passes/BloomPass.cpp
```diff
@@ -29,10 +29,8 @@ EZ_END_DYNAMIC_REFLECTED_TYPE;
 
 ezBloomPass::ezBloomPass()
   : ezRenderPipelinePass("BloomPass", true)
-  , m_fRadius(0.2f)
-  , m_fThreshold(1.0f)
-  , m_fIntensity(0.3f)
-  , m_InnerTintColor(ezColor::White)
+  , 
+   m_InnerTintColor(ezColor::White)
```
jankrassnigg: These could also be assigned directly (not important though).

jankrassnigg: The question is why clang-tidy didn't do this automatically. Maybe we need more constexpr on ezColor ? :D

jankrassnigg (approved):

Code/Engine/RendererCore/Pipeline/Implementation/Passes/BloomPass.cpp
Ingrater: yes, it only pulls out constant expressions. For some reason it might think taht ezColor::White is not a constant

Code/Editor/EditorEngineProcessFramework/EngineProcess/EngineProcessMessages.h
Ingrater: even empty costructors should not be removed but instead be ` = default` in the cpp file so that the linker doesn't have to deduplicate them.

PR #954, opened 2023-05-20
Infrastructure and changes for new scripting

C-Core:
This PR contains all the infrastructure and the changes needed for the new scripting system:
* Base class of script class resource and new attributes like e.g. a script extension to add methods to a class that is not reflected (similar to C# extension methods)
* New script component that will work with new visual scripting but also with all other scripting solutions that use the new scripting infrastructure
* New script state machine state
* Added an explicit variant type dropdown to the variant widget instead of the "right click to change type" logic that is easy to miss
* Fixed that the variant types changes in double and int spinboxes
* Tweaked visuals of node graphs, added an optional subtitle to nodes
* Dynamic pin functionality: Helper functions in node manager and special object accessor to insert the correct commands for undo and redo when dynamic pins change
* Ddl files can now store invalid variants. This is important to keep variant arrays at the correct count.
* Support to pass ezVariantArray and ezVariantDictionary as argument in a reflected function
* Reflected a few more classes and methods to expose them in the new scripting, e.g. blackboard
* Splitted ezFormatString functionality to also allow dynamic arrays as arguments instead of only variadic templates (needed for scripting log implementation)


Code/Engine/Core/ResourceManager/ResourceHandle.h
```diff
@@ -240,7 +240,7 @@ class ezTypedResourceHandle
 template <typename T>
 struct ezHashHelper<ezTypedResourceHandle<T>>
 {
-  EZ_ALWAYS_INLINE static ezUInt32 Hash(const ezTypedResourceHandle<T>& value) { return value.GetResourceIDHash(); }
+  EZ_ALWAYS_INLINE static ezUInt32 Hash(const ezTypedResourceHandle<T>& value) { return ezHashingUtils::StringHashTo32(value.GetResourceIDHash()); }
```
jankrassnigg, line 243: why is this necessary ?

Code/Engine/Core/Scripting/Implementation/ScriptClassResource.cpp
```diff
@@ -0,0 +1,105 @@
...
+  RegisterType();
+
+  SetupParentHierarchy();
+  GatherDynamicMessageHandlers();
+}
+
+ezScriptRTTI::~ezScriptRTTI()
+{
+  UnregisterType();
+  m_szTypeName = nullptr;
```
jankrassnigg, line 42: the = nullptr is unnecessary

Code/Engine/Core/Scripting/Implementation/ScriptClassResource.cpp
```diff
@@ -0,0 +1,105 @@
...
+// clang-format on
+
+ezScriptClassResource::ezScriptClassResource()
+  : ezResource(DoUpdate::OnAnyThread, 1)
+{
+}
+
+ezScriptClassResource::~ezScriptClassResource() = default;
+
+void ezScriptClassResource::CreateScriptType(const char* szName, const ezRTTI* pBaseType, ezScriptRTTI::FunctionList&& functions, ezScriptRTTI::MessageHandlerList&& messageHandlers)
```
jankrassnigg: use string views ?

Code/Engine/Core/Scripting/Implementation/ScriptClassResource.cpp
```diff
@@ -0,0 +1,105 @@
...
+{
+  ezScriptRTTI::FunctionList sortedFunctions;
+  for (auto pFuncProp : pBaseType->GetFunctions())
+  {
+    auto pBaseClassFuncAttr = pFuncProp->GetAttributeByType<ezScriptBaseClassFunctionAttribute>();
+    if (pBaseClassFuncAttr == nullptr)
+      continue;
+
+    ezStringView sBaseClassFuncName = pFuncProp->GetPropertyName();
+    if (sBaseClassFuncName.StartsWith("Reflection_"))
```
jankrassnigg: this can also be done with TrimWordStart

Code/Engine/Core/Scripting/Implementation/ScriptComponent.cpp
```diff
@@ -0,0 +1,302 @@
...
+  SUPER::Deinitialize();
+
+  ClearInstance(false);
+}
+
+void ezScriptComponent::OnActivated()
+{
+  SUPER::OnActivated();
+
+  CallScriptFunction(ezComponent_ScriptBaseClassFunctions::OnActivated);
```
jankrassnigg, line 94: is this executed at editor time as well ?

Code/Engine/Core/Scripting/Implementation/ScriptComponent.cpp
```diff
@@ -0,0 +1,302 @@
...
+}
+
+void ezScriptComponent::RemoveParameter(const char* szKey)
+{
+  if (m_Parameters.RemoveAndCopy(ezTempHashedString(szKey)))
+  {
+  }
+}
+
+bool ezScriptComponent::GetParameter(const char* szKey, ezVariant& out_value) const
```
jankrassnigg, line 205: should this return ezResult ?

Code/EditorPlugins/Assets/EditorPluginAssets/PropertyAnimAsset/PropertyAnimObjectManager.cpp
```diff
@@ -37,21 +37,3 @@ ezStatus ezPropertyAnimObjectManager::InternalCanMove(
     return ezStatus("The structure of the context cannot be animated.");
   return ezStatus(EZ_SUCCESS);
 }
-
-bool ezPropertyAnimObjectManager::IsTemporary(const ezDocumentObject* pObject) const
```
jankrassnigg, line 41: why has this been removed ?

Code/Editor/EditorFramework/PropertyGrid/Implementation/ExposedParametersPropertyWidget.cpp
```diff
@@ -250,6 +250,8 @@ void ezQtExposedParameterPropertyWidget::InternalSetValue(const ezVariant& value
           m_pWidget->setParent(this);
           m_pLayout->addWidget(m_pWidget);
           m_pWidget->Init(m_pGrid, m_pObjectAccessor, type, prop);
+
+          UpdateTypeListSelection(commonType);
```
jankrassnigg, line 254: why is this needed here ?

Code/Engine/Core/Scripting/Implementation/ScriptComponent.cpp
```diff
@@ -0,0 +1,302 @@
...
+  if (it == ezInvalidIndex)
+    return false;
+
+  out_value = m_Parameters.GetValue(it);
+  return true;
+}
+
+void ezScriptComponent::InstantiateScript(bool bActivate)
+{
+  ClearInstance(true);
```
jankrassnigg: wouldn't the bool parameter depend on whether OnActivated was called before ?

Code/Engine/Core/Scripting/Implementation/ScriptComponent.cpp
```diff
@@ -0,0 +1,302 @@
...
+
+  pModule->AddScriptReloadFunction(m_hScriptClass,
+    [this]() {
+      InstantiateScript(true);
+    });
+}
+
+const ezAbstractFunctionProperty* ezScriptComponent::GetScriptFunction(ezUInt32 uiFunctionIndex)
+{
+  if (m_pScriptType != nullptr && m_pInstance != nullptr)
```
jankrassnigg, line 290: is the instance check really needed ?

Code/Engine/Core/Scripting/ScriptAttributes.h
```diff
@@ -0,0 +1,34 @@
...
+#include <Core/CoreDLL.h>
+#include <Foundation/Reflection/Reflection.h>
+
+class EZ_CORE_DLL ezScriptExtensionAttribute : public ezPropertyAttribute
+{
+  EZ_ADD_DYNAMIC_REFLECTION(ezScriptExtensionAttribute, ezPropertyAttribute);
+
+public:
+  ezScriptExtensionAttribute();
+  ezScriptExtensionAttribute(const char* szTypeName);
```
jankrassnigg, line 14: I think you can use string views here. Not important though.

Code/Engine/Core/Scripting/ScriptAttributes.h
```diff
@@ -0,0 +1,34 @@
+#pragma once
+
+#include <Core/CoreDLL.h>
+#include <Foundation/Reflection/Reflection.h>
+
+class EZ_CORE_DLL ezScriptExtensionAttribute : public ezPropertyAttribute
```
jankrassnigg, line 8: Some documentation would be nice for these classes.

Code/Engine/Core/Scripting/ScriptClassResource.h
```diff
@@ -0,0 +1,60 @@
+#pragma once
+
+#include <Core/ResourceManager/Resource.h>
+#include <Foundation/Containers/ArrayMap.h>
+
+class ezWorld;
+using ezScriptClassResourceHandle = ezTypedResourceHandle<class ezScriptClassResource>;
+
+class EZ_CORE_DLL ezScriptInstance
```
jankrassnigg, line 9: documentation comments please

Code/Engine/Core/Scripting/ScriptComponent.h
```diff
@@ -0,0 +1,83 @@
...
+  virtual void Initialize() override;
+  virtual void Deinitialize() override;
+  virtual void OnActivated() override;
+  virtual void OnDeactivated() override;
+  virtual void OnSimulationStarted() override;
+
+  /*virtual bool OnUnhandledMessage(ezMessage& msg, bool bWasPostedMsg) override;
+  virtual bool OnUnhandledMessage(ezMessage& msg, bool bWasPostedMsg) const override;
+
+  bool HandleUnhandledMessage(ezMessage& msg, bool bWasPostedMsg);*/
```
jankrassnigg, line 28: remove

Code/Engine/Core/Scripting/ScriptComponent.h
```diff
@@ -0,0 +1,83 @@
...
+protected:
+  // virtual bool HandlesEventMessage(const ezEventMessage& msg) const override;
+
+  //////////////////////////////////////////////////////////////////////////
+  // ezScriptComponent
+public:
+  ezScriptComponent();
+  ~ezScriptComponent();
+
+  void BroadcastEventMsg(ezEventMessage& msg);
```
jankrassnigg, line 42: Why is this here? Is this a message handler?

Code/Engine/Core/Scripting/ScriptWorldModule.h
```diff
@@ -0,0 +1,64 @@
...
+  ~ezScriptWorldModule();
+
+  virtual void Initialize() override;
+
+  void AddUpdateFunctionToSchedule(const ezAbstractFunctionProperty* pFunction, void* pInstance, ezTime updateInterval);
+  void RemoveUpdateFunctionToSchedule(const ezAbstractFunctionProperty* pFunction, void* pInstance);
+
+  using ReloadFunction = ezDelegate<void()>;
+  void AddScriptReloadFunction(ezScriptClassResourceHandle hScript, ReloadFunction function);
+  void RemoveScriptReloadFunction(ezScriptClassResourceHandle hScript, void* pInstance);
```
jankrassnigg, line 25: Out of curiosity, what is this for exactly ?

Code/Engine/Core/Utils/Blackboard.h
```diff
@@ -174,6 +174,14 @@ class EZ_CORE_DLL ezBlackboard : public ezRefCounted
   /// If deserialized entries overlap with existing ones, the deserialized entries will overwrite the existing ones (both values and flags).
   ezResult Deserialize(ezStreamReader& inout_stream);
 
+public:
+  // The following functions are for scripting reflection and shouldn't be called manually
```
jankrassnigg: they don't need to be public, I think

Code/Engine/Core/World/Implementation/GameObject.cpp
```diff
@@ -35,6 +35,14 @@ EZ_BEGIN_STATIC_REFLECTED_TYPE(ezGameObject, ezNoBase, 1, ezRTTINoAllocator)
     EZ_SET_ACCESSOR_PROPERTY("Components", Reflection_GetComponents, Reflection_AddComponent, Reflection_RemoveComponent)->AddFlags(ezPropertyFlags::PointerOwner),
   }
   EZ_END_PROPERTIES;
+  EZ_BEGIN_FUNCTIONS
+  {
+    EZ_SCRIPT_FUNCTION_PROPERTY(IsActive),
+
+//    EZ_SCRIPT_FUNCTION_PROPERTY(FindChildByName, In, "Name", In, "Recursive"),
```
jankrassnigg: what about the first function ?

Code/Engine/Foundation/Math/Implementation/AllClasses_inl.h
```diff
@@ -324,3 +324,30 @@ ezResult ezMat4Template<Type>::Invert(Type fEpsilon)
...
   return EZ_SUCCESS;
 }
+
+//////////////////////////////////////////////////////////////////////////
+
+// static
+template <typename T>
+bool ezComparisonOperator::Compare(ezComparisonOperator::Enum cmp, const T& a, const T& b)
+{
+  switch (cmp)
```
jankrassnigg, line 334: this makes my head hurt, do we have a unit test that validates this is correct, e.g. for floats ?

Code/Engine/Foundation/Reflection/Implementation/MessageHandler.h
```diff
@@ -11,6 +11,8 @@ class ezMessage;
 class EZ_FOUNDATION_DLL ezAbstractMessageHandler
 {
 public:
+  EZ_DECLARE_POD_TYPE();
```
jankrassnigg: is it a good idea to do this on a base class ?

Code/Engine/Foundation/Strings/FormatString.h
```diff
@@ -72,13 +72,12 @@ class EZ_FOUNDATION_DLL ezFormatString
...
-  static void SBClear(ezStringBuilder& sb);
-  static void SBAppendChar(ezStringBuilder& sb, ezUInt32 uiChar);
-  static const char* SBReturn(ezStringBuilder& sb);
+  /// \brief Helper function to build the formatted text with the given arguments.
+  ///
+  /// \note We can't use ezArrayPtr here because of include order.
+  const char* BuildFormattedText(ezStringBuilder& ref_sStorage, ezStringView* pArgs, ezUInt32 uiNumArgs) const;
 
+protected:
   const char* m_szString;
```
jankrassnigg: = nullptr, while we are at it

Code/Engine/GameEngine/StateMachine/StateMachine.h
```diff
@@ -127,13 +146,20 @@ class EZ_GAMEENGINE_DLL ezStateMachineInstance
...
   ezReflectedClass& GetOwner() { return m_Owner; }
+  ezWorld* GetOwnerWorld();
 
   void SetBlackboard(const ezSharedPtr<ezBlackboard>& pBlackboard);
   const ezSharedPtr<ezBlackboard>& GetBlackboard() const { return m_pBlackboard; }
 
   /// \brief Returns how long the state machine is in its current state
   ezTime GetTimeInCurrentState() const { return m_TimeInCurrentState; }
 
+public:
```
jankrassnigg: private ?

Code/Engine/GameEngine/VisualScript/Implementation/VisualScriptResource.cpp
```diff
@@ -326,6 +326,8 @@ void ezVisualScriptResourceDescriptor::PrecomputeMessageHandlers()
   {
     auto& node = m_Nodes[uiNode];
     const ezRTTI* pType = node.m_pType;
+    if (pType == nullptr)
```
jankrassnigg: redundant (same check below)

Code/Tools/Libs/GuiFoundation/PropertyGrid/Implementation/AddSubElementButton.cpp
```diff
@@ -141,32 +139,6 @@ void ezQtAddSubElementButton::onMenuAboutToShow()
     }
     m_SupportedTypes.Insert(pProp->GetSpecificType());
 
-    ezVariant constraintValue;
```
jankrassnigg, line 144: Why was this removed?

Code/Tools/Libs/ToolsFoundation/NodeObject/NodeCommandAccessor.h
```diff
@@ -0,0 +1,32 @@
...
+  virtual ezStatus RemoveValue(const ezDocumentObject* pObject, const ezAbstractProperty* pProp, ezVariant index = ezVariant()) override;
+  virtual ezStatus MoveValue(
+    const ezDocumentObject* pObject, const ezAbstractProperty* pProp, const ezVariant& oldIndex, const ezVariant& newIndex) override;
+
+private:
+  bool IsDynamicPinProperty(const ezDocumentObject* pObject, const ezAbstractProperty* pProp) const;
+
+  struct ConnectionInfo
+  {
+    const ezDocumentObject* m_pSource;
```
jankrassnigg: = nullptr

Code/Tools/Libs/ToolsFoundation/Reflection/Implementation/ReflectedTypeStorageAccessor.cpp
```diff
@@ -224,11 +224,11 @@ ezInt32 ezReflectedTypeStorageAccessor::GetCount(const char* szProperty) const
   if (m_pMapping->m_PathToStorageInfoTable.TryGetValue(szProperty, storageInfo))
   {
     if (storageInfo->m_Type == ezVariant::Type::Invalid)
-      return false;
+      return -1;
 
     const ezAbstractProperty* pProp = GetType()->FindPropertyByName(szProperty);
     if (pProp == nullptr)
-      return false;
+      return -1;
```
jankrassnigg, line 231: it looks weird that the type of these two return values has changed, but nothing else in this function

Data/Tools/ezEditor/Localization/en/StateMachineAsset.txt
```diff
@@ -9,7 +9,7 @@ ezStateMachineComponent;State Machine;;https://ezengine.net/pages/docs/ai/state-
 # Properties
 
 ezComparisonOperator::Equal;==
-ezComparisonOperator::NotEqual;<>
+ezComparisonOperator::NotEqual;!=
```
jankrassnigg, line 12: != is the programmer's thing, <> is typically used by mere mortals

Code/Engine/Core/ResourceManager/ResourceHandle.h
C-Core: GetResourceIDHash returns a 64bit hash but the Hash function for the HashTable etc. expects a 32bit hash. I guess we never used a resource handle as key in a hashtable before so this was never a problem.

Code/Engine/Core/Scripting/Implementation/ScriptClassResource.cpp
C-Core: Well technically it's not necessary but the type name storage will be invalid after the destructor so it makes it easier to detect a dangling scriptRTTI pointer if we set it to null here.

Code/Engine/Core/Scripting/Implementation/ScriptComponent.cpp
C-Core: yes it is

Code/Engine/Core/Scripting/Implementation/ScriptComponent.cpp
C-Core: This function is needed for reflection so the signature is fixed

Code/EditorPlugins/Assets/EditorPluginAssets/PropertyAnimAsset/PropertyAnimObjectManager.cpp
C-Core: It has been moved to ezDocumentObjectManager

Code/Editor/EditorFramework/PropertyGrid/Implementation/ExposedParametersPropertyWidget.cpp
C-Core: The exposed parameter widget derives from variant widget which has been refactored. This call is needed so the exposed parameter widget has the correct variant type preselected in its type dropdown.

Code/Engine/Core/Scripting/Implementation/ScriptComponent.cpp
C-Core: Yes, otherwise you can't call the returned function. You could add the same check in all the places where you use the returned function but it was simpler to do it this way.

Code/Engine/Core/Scripting/Implementation/ScriptComponent.cpp
C-Core: Indeed.

Code/Engine/Core/Scripting/ScriptClassResource.h
C-Core: This classes are still very work in progress. I will add comments once they are more finalized.

Code/Engine/Core/Scripting/ScriptComponent.h
C-Core: Message handling is not implemented yet so I will leave them in there until I know whether I need them or not.

Code/Engine/Core/Scripting/ScriptComponent.h
C-Core: Some as above, no message handling or sending yet, might be needed in the future.

Code/Engine/Core/Scripting/ScriptWorldModule.h
C-Core: Typically we an update function or something like that on the component manager for resource reloading but that would mean that I have to implemented resource reloading for e.g. state machine states separatly. Registering a callback on a world module is a more generic way to handle resource reloading. This might be something we could generalize even further, not sure yet.

Code/Engine/Core/Utils/Blackboard.h
C-Core: Unfortunately they need to be public for static reflected types, but I think I will move them to an extension class to clean up the interface here.

Code/Tools/Libs/ToolsFoundation/Reflection/Implementation/ReflectedTypeStorageAccessor.cpp
C-Core: The function returns an int and -1 is the error case so it was simply wrong before.

Code/Tools/Libs/GuiFoundation/PropertyGrid/Implementation/AddSubElementButton.cpp
C-Core: This feature was not used anywhere so I removed the corresponding attribute and the code here.

Data/Tools/ezEditor/Localization/en/StateMachineAsset.txt
C-Core: Unreal uses != in blueprints and it is also consistent with c-like languages e.g. typescript etc.
image

Code/Engine/Core/Utils/Blackboard.h
C-Core: Found a better way: EZ_ALLOW_PRIVATE_PROPERTIES

Code/Engine/Foundation/Math/Implementation/AllClasses_inl.h
C-Core: added a unit test

PR #963, opened 2023-05-27
File tracking refactor

Sanakan8472:
## Asset Curator
* Tracking files has been removed from the **ezAssetCurator** and moved into its own class **ezFileSystemModel** in ToolsFoundation.
* Rewrote most of the dependency tracking code when adding / removing assets, e.g. **EnsureAssetInfoUpdated**.
* To keep the number of changes small the ezAssetCurator still has various file tracking functions like **NotifyOfFileChange** which just redirect to **ezFileSystemModel**.

## File System Model
* **ezFileSystemModel** is now a sub-system in ToolsFoundation that tracks files and folder changes in all data directories and creates a model out of these changes.
* Changes can be observed via **ezFolderChangedEvent** and **ezFileChangedEvent**.
* The model keeps maps of all files and folders over all data directories via absolute paths.
* An  **ezFileSystemWatcher** instance is created to track files in all data directories. If for some reason a change was not observed by the watcher, updates can be forced via these functions:
  * **NotifyOfChange**: Force checking the filesystem for changes to the given file or folder. This function will handle file add/remove/change as well as folder add/remove.
  * **CheckFolder**: Check an existing folder recursively for changes.
  * **CheckFileSystem**: Updates all files and folders in all data directories. Expensive and triggers a model reset event at the end.
* Both function calls and event firing is thread-safe and can happen on any thread and even at the same time.
* **HashFile** is cached and atomic via file system locks so computing the hash will never return a stale hash.
* **ReadDocument** will return an **ezStreamReader** and the up-to-date hash in the same atomic fashion as **HashFile**.
* **LinkDocument** and **UnlinkDocument** can be used to have fast lookups from files to documents / assets.
* Use **GetFiles** or **GetFolders** to access all files / folders under a lock. ** FileSystemModelTest test most of the common file system operations on the model.

## File System Watcher
* **ezAssetWatcher** has been moved to ToolsFoundation and renamed to **ezFileSystemWatcher**
* It has been cleaned up and it now fires the event **ezFileSystemWatcherEvent** instead of hardcoding operations.
* Operations are done via the tasksytem and currently require **MainThreadTick** to be called at regular intervals.
* Add revents are delayed by 5 frames while delete operations are delayed by 10 to make sure renaming / moving files / folders will first observe the addition before the removal.

## Misc fixes
* **ezTestInt** now prints the full 64bit of the test parameters to the log message instead of just the upper 32bit.
* Started replacing const char* parameters ez ezStringView.


Code/Tools/Libs/ToolsFoundation/FileSystem/Declarations.h
```diff
@@ -0,0 +1,51 @@
...
+  enum class Status : ezUInt8
+  {
+    Unknown,    ///< Since the file has been tagged as 'Unknown' it has not been encountered again on disk (yet). Use internally to find stale entries in the model.
+    FileLocked, ///< The file is locked, i.e. reading is currently not possible. Try again at a later date.
+    Valid       ///< The file exists on disk.
+  };
+
+  ezFileStatus()
+  {
+    m_uiHash = 0;
```
jankrassnigg: nit: init var directly

Code/Tools/Libs/ToolsFoundation/FileSystem/Declarations.h
```diff
@@ -0,0 +1,51 @@
...
+    m_Status = Status::Unknown;
+  }
+
+  ezTimestamp m_Timestamp;
```
jankrassnigg: what is this timestamp? last file modification? last file status update? could have a better name

Code/Tools/Libs/ToolsFoundation/FileSystem/Implementation/Declarations.cpp
```diff
@@ -0,0 +1,16 @@
...
+
+// clang-format off
+EZ_BEGIN_STATIC_REFLECTED_TYPE(ezFileStatus, ezNoBase, 3, ezRTTIDefaultAllocator<ezFileStatus>)
+  {
+    EZ_BEGIN_PROPERTIES
+    {
+      EZ_MEMBER_PROPERTY("Timestamp", m_Timestamp),
+      EZ_MEMBER_PROPERTY("Hash", m_uiHash),
+      EZ_MEMBER_PROPERTY("DocumentID", m_DocumentID),
+    } EZ_END_PROPERTIES;
```
jankrassnigg: nit: formatting

Code/Tools/Libs/ToolsFoundation/FileSystem/Implementation/Declarations.cpp
```diff
@@ -0,0 +1,16 @@
+#include <ToolsFoundation/ToolsFoundationDLL.h>
+
+#include <ToolsFoundation/FileSystem/Declarations.h>
+
+// clang-format off
+EZ_BEGIN_STATIC_REFLECTED_TYPE(ezFileStatus, ezNoBase, 3, ezRTTIDefaultAllocator<ezFileStatus>)
```
jankrassnigg, line 6: version 3 already ?

Code/UnitTests/ToolsFoundationTest/FileSystem/FileSystemModelTest.cpp
```diff
@@ -0,0 +1,695 @@
...
+EZ_CREATE_SIMPLE_TEST_GROUP(FileSystem);
+
+namespace
+{
+  ezResult eztCreateFile(ezStringView path)
+  {
+    ezFileWriter FileOut;
+    EZ_SUCCEED_OR_RETURN(FileOut.Open(path));
+    EZ_SUCCEED_OR_RETURN(FileOut.WriteString("Test"));
+    EZ_SUCCEED_OR_RETURN(FileOut.Flush());
```
jankrassnigg: flush is redundant when you close right away anyway

Code/UnitTests/ToolsFoundationTest/FileSystem/FileSystemModelTest.cpp
```diff
@@ -0,0 +1,695 @@
...
+        break;
+      case ezFileChangedEvent::Type::FileAdded:
+        EZ_TEST_BOOL(ezFileSystemModel::GetSingleton()->FindFile(e.m_sPath, stat).Succeeded());
+        break;
+      case ezFileChangedEvent::Type::FileChanged:
+        EZ_TEST_BOOL(ezFileSystemModel::GetSingleton()->FindFile(e.m_sPath, stat).Succeeded());
+        break;
+      case ezFileChangedEvent::Type::DocumentLinked:
+        EZ_TEST_BOOL(ezFileSystemModel::GetSingleton()->FindFile(e.m_sPath, stat).Succeeded());
+        break;
```
jankrassnigg: these could just fall through to the same line

Code/UnitTests/ToolsFoundationTest/FileSystem/FileSystemModelTest.cpp
```diff
@@ -0,0 +1,695 @@
...
+  {
+    ezStringBuilder sFolderPath(sOutputFolder);
+    sFolderPath.AppendPath("Folder2");
+
+    ezStringBuilder sFilePath(sOutputFolder);
+    sFilePath.AppendPath("Folder2", "rootFile2.txt");
+
+    ezFileSystemModel::GetSingleton()->CheckFileSystem();
+
+    // #TODO_ASSET This FileChanged should be removed once the model is fixed to no longer require firing this after restoring the model from cache. See comment in ezFileSystemModel::HandleSingleFile.
```
jankrassnigg, line 455: still relevant?

Code/Tools/Libs/ToolsFoundation/FileSystem/FileSystemWatcher.h
```diff
@@ -0,0 +1,91 @@
...
+  struct WatcherResult
+  {
+    ezString sFile;
+    ezDirectoryWatcherAction action;
+    ezDirectoryWatcherType type;
+  };
+
+  struct PendingUpdate
+  {
+    ezString sAbsPath;
```
jankrassnigg: missing m_

Code/Tools/Libs/ToolsFoundation/FileSystem/FileSystemWatcher.h
```diff
@@ -0,0 +1,91 @@
...
+
+private:
+  static constexpr ezUInt32 s_AddedFrameDelay = 5;
+  static constexpr ezUInt32 s_RemovedFrameDelay = 10;
+
+  struct WatcherResult
+  {
+    ezString sFile;
+    ezDirectoryWatcherAction action;
+    ezDirectoryWatcherType type;
```
jankrassnigg: missing m_

Code/Tools/Libs/ToolsFoundation/FileSystem/FileSystemWatcher.h
```diff
@@ -0,0 +1,91 @@
...
+    ezString sAbsPath;
+    ezUInt32 m_uiFrameDelay = 0;
+  };
+
+  /// \brief Handles a single change notification by a directory watcher.
+  void HandleWatcherChange(const WatcherResult& res);
+  /// \brief Handles update delays to allow compacting multiple changes.
+  void NotifyChanges();
+  /// \brief Adds a change with the given delay to the container. If the entry is already present, only its delay is increased.
+  void AddEntry(ezDynamicArray<PendingUpdate>& container, const char* szAbsPath, ezUInt32 uiFrameDelay);
```
jankrassnigg: ezStringView

Code/Tools/Libs/ToolsFoundation/FileSystem/Implementation/FileSystemWatcher.cpp
```diff
@@ -0,0 +1,209 @@
...
+// ezAssetWatcher
+////////////////////////////////////////////////////////////////////////
+
+ezFileSystemWatcher::ezFileSystemWatcher(const ezApplicationFileSystemConfig& fileSystemConfig)
+{
+  m_FileSystemConfig = fileSystemConfig;
+}
+
+
+ezFileSystemWatcher::~ezFileSystemWatcher()
```
jankrassnigg: = default

Code/Tools/Libs/ToolsFoundation/FileSystem/FileSystemModel.h
```diff
@@ -0,0 +1,195 @@
...
+  /// \brief Updates all files and folders in the model by iterating over all data directories. This is very expensive and should be done on a worker thread.
+  void CheckFileSystem();
+
+  ///@}
+  /// \name File Meta Operations
+  ///@{
+
+  /// \brief Links a document Id to the given file. This allows for fast lookups whether a file is also a document.
+  /// \param sAbsolutePath Path to the document. Must be in the model.
+  /// \param documentId The Id of the document that should be linked to th file.
```
jankrassnigg: minor typo "th"

Code/Tools/Libs/ToolsFoundation/FileSystem/FileSystemModel.h
```diff
@@ -0,0 +1,195 @@
...
+  ezResult LinkDocument(ezStringView sAbsolutePath, const ezUuid& documentId);
+
+  /// \brief Unlinks a document from a file
+  /// \param sAbsolutePath Path to the document. Must be in the model.
+  /// \return Returns EZ_SUCCESS if the file existed.
+  ezResult UnlinkDocument(ezStringView sAbsolutePath);
+
+  /// \brief Creates a file reader to the given file. Will also link the document and hash it in a file-system-atomic operation.
+  /// \param sAbsolutePath Path to the document. Must be in the model.
+  /// \param callback Called once the file was opened and hashed. The ezFileStatus contains the up to date info for the file, including hash. This function should return th document Id as it will automatically call LinkDocument.
```
jankrassnigg: same typo

Code/Tools/Libs/ToolsFoundation/FileSystem/Implementation/FileSystemModel.cpp
```diff
@@ -0,0 +1,855 @@
...
+ezFileChangedEvent::ezFileChangedEvent(ezStringView sFile, ezFileStatus status, Type type)
+  : m_sPath(sFile)
+  , m_Status(status)
+  , m_Type(type)
+{
+}
+
+bool ezFileSystemModel::IsSameFile(ezStringView sAbsolutePathA, ezStringView sAbsolutePathB)
+{
+#if (EZ_ENABLED(EZ_SUPPORTS_CASE_INSENSITIVE_PATHS))
```
jankrassnigg: are these paths guaranteed to be "clean" ?

Code/Tools/Libs/ToolsFoundation/FileSystem/Implementation/FileSystemModel.cpp
```diff
@@ -0,0 +1,855 @@
...
+////////////////////////////////////////////////////////////////////////
+// ezAssetFiles
+////////////////////////////////////////////////////////////////////////
+
+ezFileSystemModel::ezFileSystemModel()
+  : m_SingletonRegistrar(this)
+{
+}
+
+ezFileSystemModel::~ezFileSystemModel()
```
jankrassnigg: = default

Code/Tools/Libs/ToolsFoundation/FileSystem/Implementation/FileSystemModel.cpp
```diff
@@ -0,0 +1,855 @@
...
+    {
+      if (ezFileSystem::ResolveSpecialDirectory(m_FileSystemConfig.m_DataDirs[i].m_sDataDirSpecialPath, sDataDirPath).Failed())
+      {
+        ezLog::Error("Failed to resolve data directory named '{}' at '{}'", m_FileSystemConfig.m_DataDirs[i].m_sRootName, m_FileSystemConfig.m_DataDirs[i].m_sDataDirSpecialPath);
+        m_DataDirRoots.PushBack({});
+      }
+      else
+      {
+        sDataDirPath.MakeCleanPath();
+        while (sDataDirPath.EndsWith("/"))
```
jankrassnigg: use Trim ?

Code/Tools/Libs/ToolsFoundation/FileSystem/Implementation/FileSystemModel.cpp
```diff
@@ -0,0 +1,855 @@
...
+  HandleSingleFile(sPath, true);
+}
+
+void ezFileSystemModel::CheckFileSystem()
+{
+  if (!m_bInitialized)
+    return;
+
+  EZ_PROFILE_SCOPE("CheckFileSystem");
+  ezStopwatch sw;
```
jankrassnigg: I think sw isn't used anymore.

Code/Tools/Libs/ToolsFoundation/FileSystem/Implementation/FileSystemModel.cpp
```diff
@@ -0,0 +1,855 @@
...
+  {
+    it = m_ReferencedFiles.Find(sPath);
+  }
+  else
+  {
+    // Data dir parent relative?
+    for (const auto& dd : m_FileSystemConfig.m_DataDirs)
+    {
+      ezStringBuilder sDataDir;
+      ezFileSystem::ResolveSpecialDirectory(dd.m_sDataDirSpecialPath, sDataDir).IgnoreResult();
```
jankrassnigg: AssertSuccess instead of IgnoreResult

Code/Tools/Libs/ToolsFoundation/FileSystem/Implementation/FileSystemModel.cpp
```diff
@@ -0,0 +1,855 @@
...
+        break;
+    }
+
+    if (!it.IsValid())
+    {
+      // Data dir relative?
+      for (const auto& dd : m_FileSystemConfig.m_DataDirs)
+      {
+        ezStringBuilder sDataDir;
+        ezFileSystem::ResolveSpecialDirectory(dd.m_sDataDirSpecialPath, sDataDir).IgnoreResult();
```
jankrassnigg: AssertSuccess

Code/Tools/Libs/ToolsFoundation/FileSystem/Implementation/FileSystemModel.cpp
```diff
@@ -0,0 +1,855 @@
...
+    }
+    else
+    {
+      return EZ_FAILURE;
+    }
+  }
+
+  if (bDocumentLinkChanged)
+  {
+    FireFileChangedEvent(sAbsolutePath, fileStatus, ezFileChangedEvent::Type::DocumentLinked);
```
jankrassnigg: Is there a DocumentUnlinked event type that it should fire here?

Code/Tools/Libs/ToolsFoundation/FileSystem/Implementation/FileSystemModel.cpp
```diff
@@ -0,0 +1,855 @@
...
+  }
+  return EZ_SUCCESS;
+}
+
+ezResult ezFileSystemModel::HashFile(ezStringView sAbsolutePath, ezFileStatus& out_stat)
+{
+  if (!m_bInitialized)
+    return EZ_FAILURE;
+
+  // #TODO_ASSET Sanity check that file path is under m_FileSystemConfig
```
jankrassnigg: I think the code below already deals with that. Since it checks m_ReferencedFiles and then fails if it doesn't find anything.

Code/Tools/Libs/ToolsFoundation/FileSystem/Implementation/FileSystemModel.cpp
```diff
@@ -0,0 +1,855 @@
...
+    }
+  }
+
+  // We can only hash files that are tracked.
+  if (out_stat.m_Status == ezFileStatus::Status::Unknown)
+  {
+    out_stat = HandleSingleFile(sAbsolutePath, statDep, false);
+    if (out_stat.m_Status == ezFileStatus::Status::Unknown)
+    {
+      ezLog::Error("Failed to update file '{0}'", sAbsolutePath);
```
jankrassnigg: update or hash ?

Code/Tools/Libs/ToolsFoundation/FileSystem/Implementation/FileSystemModel.cpp
```diff
@@ -0,0 +1,855 @@
...
+      return EZ_FAILURE;
+    }
+
+    // We need to request the stats again wile while we have shared read access or we might trigger a race condition of writes to the file between the last stat call and the current file open.
+    if (ezOSFile::GetFileStats(sAbsolutePath, statDep).Failed())
+    {
+      ezLog::Error("Failed to retrieve file stats '{0}'", sAbsolutePath);
+      return EZ_FAILURE;
+    }
+    out_stat.m_Timestamp = statDep.m_LastModificationTime;
```
jankrassnigg: so m_Timestamp is the last modification time...

Code/Tools/Libs/ToolsFoundation/FileSystem/Implementation/FileSystemModel.cpp
```diff
@@ -0,0 +1,855 @@
...
+  ezUInt8 cachedBytes[1024 * 10];
+
+  while (true)
+  {
+    const ezUInt64 uiRead = InputStream.ReadBytes(cachedBytes, EZ_ARRAY_SIZE(cachedBytes));
+
+    if (uiRead == 0)
+      break;
+
+    hsw.WriteBytes(cachedBytes, uiRead).IgnoreResult();
```
jankrassnigg: AssertSuccess

Code/Tools/Libs/ToolsFoundation/FileSystem/Implementation/FileSystemModel.cpp
```diff
@@ -0,0 +1,855 @@
...
+}
+
+
+void ezFileSystemModel::CheckFolder(ezStringView sAbsolutePath)
+{
+  ezStringBuilder sAbsolutePath2 = sAbsolutePath;
+  sAbsolutePath2.MakeCleanPath();
+  EZ_ASSERT_DEV(ezPathUtils::IsAbsolutePath(sAbsolutePath2), "Only absolute paths are supported for directory iteration.");
+
+  while (sAbsolutePath2.EndsWith("/"))
```
jankrassnigg: Trim

Code/Tools/Libs/ToolsFoundation/FileSystem/Implementation/FileSystemModel.cpp
```diff
@@ -0,0 +1,855 @@
...
+    }
+
+    if (!bExisted)
+    {
+      FireFolderChangedEvent(sAbsolutePath, ezFolderChangedEvent::Type::FolderAdded);
+      if (bRecurseIntoFolders)
+        CheckFolder(sAbsolutePath);
+    }
+
+    return {};
```
jankrassnigg: is this default result intended ?

Code/Tools/Libs/ToolsFoundation/FileSystem/Implementation/FileSystemModel.cpp
```diff
@@ -0,0 +1,855 @@
...
+      auto it = m_ReferencedFiles.FindOrAdd(sAbsolutePath, &bExisted);
+      ezFileStatus& value = it.Value();
+      bFileChanged = !value.m_Timestamp.Compare(FileStat.m_LastModificationTime, ezTimestamp::CompareMode::Identical);
+      if (bFileChanged)
+      {
+        value.m_uiHash = 0;
+      }
+
+      // If the state is unknown, we loaded it from the cache and need to fire FileChanged to update dependent systems.
+      // #TODO_ASSET This behaviors should be changed once the asset cache is stored less lossy.
```
jankrassnigg, line 814: still todo ?

Code/Tools/Libs/ToolsFoundation/FileSystem/Implementation/FileSystemModel.cpp
```diff
@@ -0,0 +1,855 @@
...
+    m_LockedFiles.Insert(sAbsolutePath);
+  }
+}
+
+void ezFileSystemModel::FireFileChangedEvent(ezStringView sFile, ezFileStatus fileStatus, ezFileChangedEvent::Type type)
+{
+  // We queue up all requests on a thread and only return once the list is empty. The reason for this is that:
+  // A: We don't want to allow recursive event calling as it creates limbo states in the model and hard to debug bugs.
+  // B: If a user calls NotifyOfChange, the function should only return if the event and any indirect events that were triggered by the event handlers have been processed.
+  static thread_local ezHybridArray<ezFileChangedEvent, 2, ezStaticAllocatorWrapper> m_PostponedFiles;
```
jankrassnigg: static thread_local might not be allowed on all compilers. Move the variable out of the function and just make it a thread_local global variable.

Code/Tools/Libs/ToolsFoundation/FileSystem/Implementation/FileSystemModel.cpp
```diff
@@ -0,0 +1,855 @@
...
+  ezFileChangedEvent& e = m_PostponedFiles.ExpandAndGetRef();
+  e.m_sPath = sFile;
+  e.m_Status = fileStatus;
+  e.m_Type = type;
+
+  static thread_local bool bInFileBroadcast;
```
jankrassnigg: same here, also give it a default value

Code/Tools/Libs/ToolsFoundation/FileSystem/Implementation/FileSystemModel.cpp
```diff
@@ -0,0 +1,855 @@
...
+
+void ezFileSystemModel::FireFolderChangedEvent(ezStringView sFile, ezFolderChangedEvent::Type type)
+{
+  // See comment in FireFileChangedEvent.
+  static thread_local ezHybridArray<ezFolderChangedEvent, 2, ezStaticAllocatorWrapper> m_PostponedFolders;
+  ezFolderChangedEvent& e = m_PostponedFolders.ExpandAndGetRef();
+  e.m_sPath = sFile;
+  e.m_Type = type;
+
+  static thread_local bool bInFolderBroadcast;
```
jankrassnigg: same + default value

Code/Editor/EditorFramework/Assets/Implementation/AssetCurator.cpp
```diff
@@ -1553,39 +1496,112 @@ ezSubAsset* ezAssetCurator::GetSubAssetInternal(const ezUuid& assetGuid)
...
+    }
+  }
+}
+
+void ezAssetCurator::OnAssetFilesEvent(const ezFileChangedEvent& e)
+{
+  switch (e.m_Type)
+  {
+    case ezFileChangedEvent::Type::DocumentLinked:
+    {
```
jankrassnigg: unnecessary braces

Code/Editor/EditorFramework/Assets/Implementation/AssetCurator.cpp
```diff
@@ -1553,39 +1496,112 @@ ezSubAsset* ezAssetCurator::GetSubAssetInternal(const ezUuid& assetGuid)
...
+          for (const ezUuid& guid : it2.Value())
+          {
+            InvalidateAssetTransformState(guid);
+          }
+        }
+      }
+
+      // Assets should never be in an AssetCache folder.
+      const char* szNeedle = e.m_sPath.FindSubString("AssetCache/");
+      if (szNeedle != nullptr && e.m_sPath.GetData() != szNeedle && szNeedle[-1] == '/')
```
jankrassnigg: accessing szNeedle at -1 looks weird. I see that you check against the problems, but not sure whether all static analyzers would get that. Maybe there is a better way to write this.

Code/Editor/EditorFramework/Assets/Implementation/AssetCurator.cpp
```diff
@@ -1553,39 +1496,112 @@ ezSubAsset* ezAssetCurator::GetSubAssetInternal(const ezUuid& assetGuid)
...
-      auto it = m_InverseTransformDeps.Find(sAbsolutePath);
+      // check that this is an asset type that we know
+      ezStringBuilder sExt = ezPathUtils::GetFileExtension(e.m_sPath);
+      sExt.ToLower();
+      if (!m_ValidAssetExtensions.Contains(sExt))
+      {
+        return;
+      }
+
+      // #TODO_ASSET -> wrong? This will update the timestamp for assets.
```
jankrassnigg: todo still valid ?

Code/Editor/EditorFramework/Assets/Implementation/AssetCurator.cpp
```diff
@@ -1553,39 +1496,112 @@ ezSubAsset* ezAssetCurator::GetSubAssetInternal(const ezUuid& assetGuid)
...
+    {
+      ezUuid guid0 = e.m_Status.m_DocumentID;
+      if (guid0.IsValid())
+      {
+        // this is a bit tricky:
+        // when the document is deleted on disk, it would be nicer not to close it (discarding modifications!)
+        // instead we could set it as modified
+        // but then when it was only moved or renamed that means we have another document with the same GUID
+        // so once the user would save the now modified document, we would end up with two documents with the same GUID
+        // so, for now, since this is probably a rare case anyway, we just close the document without asking
```
jankrassnigg, line 1590: Did you test this recently? A long time ago I added this feature to close documents that were modified on disk, but recently I noticed that it often didn't work anymore. Does it work better now?

Code/Editor/EditorFramework/Assets/Implementation/AssetCurator.cpp
```diff
@@ -1603,89 +1619,10 @@ void ezAssetCurator::HandleSingleFile(const ezString& sAbsolutePath)
...
-    return;
-
-  // store the folder of the asset
-  {
-    ezStringBuilder sAssetFolder = sAbsolutePath;
-    sAssetFolder = sAssetFolder.GetFileDirectory();
-
-    m_AssetFolders.Insert(sAssetFolder);
+    break;
+    default:
```
jankrassnigg: should we just ignore all cases, or assert that we covered all known cases ?

jankrassnigg (approved):

Code/UnitTests/ToolsFoundationTest/FileSystem/FileSystemModelTest.cpp
Sanakan8472: Yes. Right now this is a quirk that is required to make the asset curator work. Once it stores its entire state in the cache this will no longer be necessary.

Code/Tools/Libs/ToolsFoundation/FileSystem/Implementation/FileSystemModel.cpp
Sanakan8472: Yes. Only used in one place so far. I've extended the description to make this clear.

Code/Tools/Libs/ToolsFoundation/FileSystem/Implementation/FileSystemModel.cpp
Sanakan8472: There is now. The event is supposed to be used by the asset brower view later on but that isn't implemented yet.

Code/Tools/Libs/ToolsFoundation/FileSystem/Implementation/FileSystemModel.cpp
Sanakan8472: Actually this as still missing. As this is a public function it needs to check against files outside its data dirs. Otherwise it would call the internal HandleSingleFile function which is not guarded against files outside.

Code/Tools/Libs/ToolsFoundation/FileSystem/Implementation/FileSystemModel.cpp
Sanakan8472: Renamed.

Code/Tools/Libs/ToolsFoundation/FileSystem/Implementation/FileSystemModel.cpp
Sanakan8472: Ups! No, not at all, thanks!

Code/Tools/Libs/ToolsFoundation/FileSystem/Implementation/FileSystemModel.cpp
Sanakan8472: Yep, will need to change the asset cache first to serialize the full state to cache before this quirk can be changed.

Code/Editor/EditorFramework/Assets/Implementation/AssetCurator.cpp
Sanakan8472: I can't remember why I implemented it like this. Now the input is always an absolute path, so if AssetCache is in the path, it will be in the form of "/AssetCache/" so I just search for that now.

Code/Editor/EditorFramework/Assets/Implementation/AssetCurator.cpp
Sanakan8472: I don't think so, assets no longer store timestamps. They are all in the ezFileSystemModel now.

Code/Editor/EditorFramework/Assets/Implementation/AssetCurator.cpp
Sanakan8472: I haven't tested it extensively no. EnsureDocumentIsClosedInAllManagers in itself should work fine. If it failed in the past it must have been due to bugs in the file tracking. This should be much more stable now (and tested).

Code/Tools/Libs/ToolsFoundation/FileSystem/Implementation/FileSystemModel.cpp
Sanakan8472: Turns out this is more complicated then expected. The previous implementation was hapily tracking and caching files outside the data dirs which we relied on for stuff in the Data\Tools\ezEditor path. Preventing hashing from working on those files makes the editor fail. So I now allow hashing files outside the data dirs but the hashes are cached in a separate transient map.

Code/Editor/EditorFramework/Assets/Implementation/AssetCurator.cpp
Sanakan8472: Added assert

Code/Tools/Libs/ToolsFoundation/FileSystem/Implementation/FileSystemModel.cpp
jankrassnigg: Why do we need those files to be hashed?... Oh is it for the Visual shader stuff?

Code/Tools/Libs/ToolsFoundation/FileSystem/Implementation/Declarations.cpp
Sanakan8472: This was moved from the asset curator code which previous owned this class. Technically we could change the version as it is only serialized in the cache which is reset with this commit but I think it would be a bad idea to move the version back in those cases in cases.

jankrassnigg: Not important.

jankrassnigg (approved):

Code/Tools/Libs/ToolsFoundation/FileSystem/Implementation/FileSystemModel.cpp
Sanakan8472: Yes visual shader stuff. It makes sense to be hashed but I don't know if we want to expose all of that as a data directory so for now I simply allow hashing files outside data dirs.

PR #967, opened 2023-05-29
More migrations to ezStringView

jankrassnigg:


Code/Engine/Foundation/Types/Implementation/Variant.cpp
```diff
@@ -76,6 +76,19 @@ ezVariant::ezVariant(const ezTypedObject& value)
   m_bIsShared = true;
 }
 
+ezVariant::ezVariant(const ezStringView& value, bool bCopyString)
+{
+  if (bCopyString)
+  {
+    const ezRTTI* pType = ezGetStaticRTTI<ezString>();
```
C-Core: InitShared(ezString(value)) would be cleaner instead of copying the inner logic here

PR #973, opened 2023-06-05
Use PowerShell to download version HTML file

### LINKED ISSUE #969: [BUG] Checking for latest EZ version broke due to HTTPS enforcement
The editor regularly checks for the latest version and shows a message box, if a new release is available.
Since our Qt version doesn't have support for HTTPS, this only worked because the documentation webpage that it downloads to read the version was also accessible via HTTP.
Recently we force enabled HTTPS on the webpage, which now means that the editor can't read that URL anymore.

The code to read the EZ version is in ezQtVersionChecker::Check() (Code\Editor\EditorFramework\EditorApp\CheckVersion.cpp).

It tries to read http://ezengine.net/pages/getting-started/binaries.html (note that HTTP prefix).

The question is how to solve this.

1. We could disable HTTPS on the webpage again (https://discord.com/channels/806092968346779699/806092968346779702/1108426687005335642).
2. We could store the latest version somewhere else, where we have no encryption. Though I have no idea where.
3. We could somehow get encryption support into our C++ code. This would either require a new Qt version (big upgrade) or we need to use some other C++ library to do so. In my experience this is very painful in C++.
4. Thinking out of the box, maybe one could use a PowerShell or Python script or some other system call to let someone else do the heavy lifting.

jankrassnigg:
Fixes #969.

Attempts to run a PowerShell command to download the webpage where the latest released version can be found. This is used so that we don't need support for encryption (OpenSSL) just for this task.

Also fixes that the Utf-16 Byte-Order-Marks for Big Endian and Little Endian were swapped.

### DIFF
--- a/Code/Editor/EditorFramework/EditorApp/CheckVersion.cpp
+++ b/Code/Editor/EditorFramework/EditorApp/CheckVersion.cpp
@@ -18,13 +18,13 @@ static ezString GetVersionFilePath()
   return sTemp;
 }
 
-PageDownloader::PageDownloader(const QString& url)
+PageDownloader::PageDownloader(const QString& sUrl)
 {
 #if EZ_ENABLED(EZ_PLATFORM_WINDOWS_DESKTOP)
   QStringList args;
 
   args << "-Command";
-  args << QString("(Invoke-webrequest -URI \"%1\").Content > \"%2\"").arg(url).arg(GetVersionFilePath().GetData());
+  args << QString("(Invoke-webrequest -URI \"%1\").Content > \"%2\"").arg(sUrl).arg(GetVersionFilePath().GetData());
 
   m_pProcess = EZ_DEFAULT_NEW(QProcess);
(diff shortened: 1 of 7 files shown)


Code/Editor/EditorFramework/EditorApp/CheckVersion.cpp
```diff
@@ -9,29 +9,53 @@
...
-  if (e != QNetworkReply::NetworkError::NoError)
+  ezOSFile file;
+  if (file.Open(GetVersionFilePath(), ezFileOpenMode::Read).Failed())
+    return;
+
+  ezDataBuffer content;
+  file.ReadAll(content);
+
+  content.PushBack('\0');
+  content.PushBack('\0');
```
Ingrater, line 49: Is it intentional that this line is done twice?

Ingrater (approved):

Code/Editor/EditorFramework/EditorApp/CheckVersion.cpp
jankrassnigg: Yes, in case it is a Utf-16 string.

PR #976, opened 2023-06-06
Proposal for changing math class construction to static functions

jankrassnigg:
This has been inspired by how the math classes are designed in Jolt and I find this quite clean. It solves the problem that construction functions are not as easy to find (they all start with lower case s). It also removes the need for different constructors, that all have the same name, but may take very different parameters, and are thus difficult to know from a glance what they are doing.
It also allows to construct in different ways from the same kind of data, and use a different function name to differentiate.

It also makes it easier to construct a const variable, which is not possible if you need to call a non-const member function to configure a variable

It also gives you more options to be explicit. For example:

```
ezVec3 v; // < NaN in debug builds, uninitialized in release
ezVec3 v2 = ezVec3::sNaN(); // clearly always initialized to NaN
```

The only downside (apart from the huge change) might be that this would now always make a copy (though maybe the compilers are good enough to prevent that).

There is obviously lots of room for dicussions and opinions. This PR is mainly to get a rough idea how this could look and to then have a heated discussion about it.

### DIFF
--- a/Code/BuildSystem/CMake/CMakeUtils/ezUtilsCppFlags.cmake
+++ b/Code/BuildSystem/CMake/CMakeUtils/ezUtilsCppFlags.cmake
@@ -149,7 +149,7 @@ function(ez_set_build_flags_msvc TARGET_NAME)
 	target_compile_options(${TARGET_NAME} PRIVATE /wd5240)
     
     # Disable deprecation warnings (qt, etc)
-    target_compile_options(${TARGET_NAME} PRIVATE /wd4996)
+    # target_compile_options(${TARGET_NAME} PRIVATE /wd4996)
 
 endfunction()
 
--- a/Code/Engine/Core/Graphics/Implementation/Camera.cpp
+++ b/Code/Engine/Core/Graphics/Implementation/Camera.cpp
@@ -35,8 +35,8 @@ EZ_END_STATIC_REFLECTED_ENUM;
 
 ezCamera::ezCamera()
 {
-  m_vCameraPosition[0].SetZero();
-  m_vCameraPosition[1].SetZero();
+  m_vCameraPosition[0] = ezVec3::sZero();
+  m_vCameraPosition[1] = ezVec3::sZero();
   m_mViewMatrix[0].SetIdentity();
   m_mViewMatrix[1].SetIdentity();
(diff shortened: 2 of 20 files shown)


C-Core:
I think it makes sense as a replacement for the SetSomething functions, because for them you always had to write 2 lines before and the additional copy should be optimized away be the compiler. I'm still a bit unsure about the s prefix, because we have not used any kind of prefixes for functions before, so it feels a bit out of place.
I don't like it for things like CompMin/CompMul etc. It just feels unnecessary and requires more typing the just of writing "a." and then get a list to autocomplete.

jankrassnigg:
> I think it makes sense as a replacement for the SetSomething functions

Yep, I've never been too happy that there are both constructors and Set functions, so this should become more consistent.

> I'm still a bit unsure about the s prefix, it feels a bit out of place.

Agreed, I'm not sure about it either. Jolt uses this, and after having used Jolt for a while I find it extremely useful, because when you type "Type::s" you immediately see all the construction functions. It is really helpful to easily see all the ways that a type can be initialized. For instance on ezVec3 you have over 60 functions, but only a few are for construction purposes, it's easy to miss that there is a better way to do things, when they aren't readily apparent.

That said, I could live without a prefix. Or we could use a different one. You could also use "Construct" as prefix (e.g. "ConstructFromMat3"), but it's very cumbersome, especially for things like ezVec3::ConstructZero() etc.

Frankly, if we were to adopt a short prefix for such construction functions, we could also enforce via clang-tidy, that those functions always need to be static and return the own type. So it would be useful to have a common prefix.

I'm not advocating to prefix ALL static functions, only the ones that construct an instance.

> I don't like it for things like CompMin/CompMul etc.

Also agreed. For some functions it is a bit hard to say whether they are "constructors". You can rewrite a lot of functions in this style (e.g. also Dot and Cross), but I think we shouldn't. I've included those in the PR for discussion, since Jolt does this with a lot of functions, and for some I haven't made up my mind either.

jankrassnigg:
Btw I would argue that regular constructors should still be used for the most basic component-wise initializations. So e.g. ezVec3(x, y, z), ezTransform(position, rotation, scale) or ezColor(r, g, b, a). I think these are so obvious, that using a static creator function is contra-productive.

PR #984, opened 2023-06-18
More clang-tidy static analysis fixes

Ingrater:


jankrassnigg (approved):
Did you give those Qt mouse position changes a try to check that these functions don't return values in a different space ?

Ingrater:
> Did you give those Qt mouse position changes a try to check that these functions don't return values in a different space ?

All the Qt related changes are deprecation warnings. For them I take the implementation of the deprecated method and just paste it at the call site. As a result I'm very certain that this will not cause any behavioral changes. The interesting part is, that the microsoft compiler seems to not issue any of these deprecation warnings, while clang does. So even though we turned deprecation warnings back on for the microsoft compiler, it still just flat out ignores most of the QT deprecation warnings.

I'm currently on vacation with a laptop so testing these changes in the editor is difficult.

C-Core (approved):

PR #989, opened 2023-06-22
Animation graphs now store instance state separately

jankrassnigg:
The split between program state and instance state is not finished, yet, because some code requires larger changes. However, the infrastructure is now in, and several nodes already use it. That means that nodes are currently still allocated per instance, and not yet shared, but that's a simple change, once all nodes are migrated.


C-Core (approved):
Looks good to me

PR #1000, opened 2023-06-29
Screenspace effects

### LINKED ISSUE #824: [FEATURE] Component to add a fullscreen effect
We could use a component that, when present in a scene, renders a fullscreen quad.

It should be configurable at what point it is rendered (pre / post tonemapping etc). Need a way to define the sorting order.
You can select which material to use.
It should expose a color or some other way to input custom parameters.

For maximum flexibility it may also be useful to be able to rotate and offset the quad, so that one can place blotches at different places in screenspace and rotate them around the screen axis. Then one would also need a scale/size value.

This can be used for:
* Full screen pictures (loading screen background)
* Fade in / out masks
* Some HUD elements

C-Core:
* Added depth offset that only affects the sorting order of meshes. This can be used to force render order for transparent objects that are actually at the same position, e.g. inside and outside of a bottle.
* Added a damage indicator screen space effect to the Testing chambers project which uses a special shader to render the mesh in screenspace at the correct place in the render pipeline (setting RenderDataCategory to SimpleTransparent).
* Fixed the quaternion <-> euler angle convertions to use the correct euler angle order for our coordinate system of ZYX aka 3-2-1 and extended the unit tests to test for that.

Fixes #824

Also 1000 🥳 🎉


JailbreakPapa:
Nice, this brings in room for SS effects like SSR,SSS,SSGI,etc. been waiting for this one for a while now

jankrassnigg (approved):

PR #1013, opened 2023-07-21
Refactoring classes to use static Make* construction functions

jankrassnigg:
Notable changes:

* MakeXYZ functions for the math classes and some others
* ezQuat doesn't use an ezVec3 internally anymore for the 'vector part'. So far the tests look good, but I am not 100% sure this can't create any problems with the editor data.

This PR only adds the new functions and marks up which functions would be deprecated in the future. It doesn't make any changes to the rest of the code yet, as that would be quite a large PR that's difficult to review, so that will come later.


C-Core (commented):
I think this naming convention with the "Make" prefix is much better than the "s" prefix and more in line with the "ez style" of naming things. So green light from my side.

Code/Engine/Foundation/Time/Time.h
```diff
@@ -15,33 +15,40 @@ struct EZ_FOUNDATION_DLL ezTime
   static ezTime Now(); // [tested]
 
   /// \brief Creates an instance of ezTime that was initialized from nanoseconds.
-  EZ_ALWAYS_INLINE constexpr static ezTime Nanoseconds(double fNanoseconds) { return ezTime(fNanoseconds * 0.000000001); }
+  [[nodiscard]] EZ_ALWAYS_INLINE constexpr static ezTime MakeNanoseconds(double fNanoseconds) { return ezTime(fNanoseconds * 0.000000001); }
```
C-Core: Shouldn't that be MakeFromNanoseconds etc. to follow the naming scheme of the other classes?

Code/UnitTests/FoundationTest/SimdMath/SimdTransformTest.cpp
```diff
@@ -38,6 +38,29 @@ EZ_CREATE_SIMPLE_TEST(SimdMath, SimdTransform)
       EZ_TEST_BOOL((t.m_Scale == ezSimdVec4f(1)).AllSet<3>());
     }
 
+    {
+      ezSimdQuat qRot;
```
C-Core, line 42: Use new MakeFromAxisAndAngle function

C-Core (approved):

Code/UnitTests/FoundationTest/SimdMath/SimdTransformTest.cpp
jankrassnigg: All these changes come in a separate PR.

Code/Engine/Foundation/Time/Time.h
jankrassnigg: Theoretically yes, but that makes it even longer. I don't know, how do you feel about it ?

C-Core: I would prefer consistency in this case

jankrassnigg: Done

PR #1014, opened 2023-07-22
UI improvements

PlasmaDev5:
Backport of some feature from my personal branch by request of Jan.

First of all iv added the ability to set display a color for components in the properties window. This allows for quickly understanding what system a component belongs to. For example you could make all physics components blue. I have not setup any components in this pr as im not sure what colors you would like. 
To set the color just add the ezColorAttribute to your component.
image

Secondly iv provided the darker theme i use in my personal branch. This is a little softer on the eyes for long sessions. In some places it also provides a higher level of contrast. Long terms i feel the best solution would moving all the theming information into a stylesheet and allow for switching stylesheet in the editor preferences. 
image


jankrassnigg:
@PlasmaDev5 could you change the target branch of the PR to "user/plasmadev/ui" ? Then we can merge the PR immediately and then decide whether we make additional changes on top, before we merge it further to dev.

jankrassnigg:
> @PlasmaDev5 could you change the target branch of the PR to "user/plasmadev/ui" ? Then we can merge the PR immediately and then decide whether we make additional changes on top, before we merge it further to dev.

Turns out GitHub actually allowed me to do this already, so all good!

PlasmaDev5:
nice, was unsure if i had permission to make branches on the ez side

jankrassnigg:
> nice, was unsure if i had permission to make branches on the ez side

No you don't. Unless you are a registered user of a team in GitHub, you never have those permissions on a repository. I had to create that branch myself first.

PR #1016, opened 2023-07-23
Require Qt 6.3 as minimum

Ingrater:
Qt 6.3 contains a important bug fix for the 3D view port of the Editor on Linux. With this bug fix the editor 3d view port finally fully works and does so without requiring shared textures.

### DIFF
--- a/Data/Base/RenderPipelines/EditorRenderPipeline.ezRenderPipelineAsset
+++ b/Data/Base/RenderPipelines/EditorRenderPipeline.ezRenderPipelineAsset
@@ -11,7 +11,7 @@ o
 		s %AssetType{"RenderPipeline"}
 		VarArray %Dependencies{}
 		Uuid %DocumentID{u4{5189305147899033520,5264929536831011917}}
-		u4 %Hash{1394242861671256051}
+		u4 %Hash{11687044204642216029}
 		VarArray %MetaInfo{}
 		VarArray %Outputs{}
 		VarArray %PackageDeps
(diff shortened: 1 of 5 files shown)


Data/Base/RenderPipelines/EditorRenderPipeline.ezRenderPipelineAsset
```diff
@@ -11,7 +11,7 @@ o
 		s %AssetType{"RenderPipeline"}
 		VarArray %Dependencies{}
 		Uuid %DocumentID{u4{5189305147899033520,5264929536831011917}}
-		u4 %Hash{1394242861671256051}
+		u4 %Hash{11687044204642216029}
```
Ingrater: TODO: Revert (MSAA does not work on linux yet)

jankrassnigg:
What about windows? Do we need 6.3 there as well? Would be best to use the same version everywhere.

jankrassnigg (approved):

Ingrater:
> What about windows? Do we need 6.3 there as well? Would be best to use the same version everywhere.

As usual the answer is complicated. On windows we have been using Qt 6.4.0 so far. On Linux we attempted to use whatever Qt version did come with the package manager, as this is typically the best maintained version and the least amount of hassle to install. Unfortunately that version is to old on Ubuntu 22.04 (LTS) and thus we had to change paths. Our Linux CI Agent is now using Qt 6.4.3. The minimum requirement of 6.3 does not dictate which version to use, it just means 6.3 or newer. So we now require at least Qt 6.3.0 on all platforms. Given our cmake setup everyone is free to use whatever version of Qt they want anyway, so having some difference in used Qt versions might not hurt. I still think it would make sense to update Qt on windows to 6.4.3 as well. I'll talk to @Sanakan8472 about that.

PR #1027, opened 2023-08-12
Renderer cleanup and unit tests for texture readback

Sanakan8472:
## Renderer
 * Removed stream output support.
 * Added `ezGALResourceFormatSupport` flags (Sample, Render, VertexAttribute) which can be queried for each format via `ezGALDeviceCapabilities::m_FormatSupport`.
 * Removed `ezGALDeviceCapabilities::m_bB5G6R5Textures` as it is replaced by `ezGALResourceFormatSupport`.
 * Added `ezGALResourceFormat::IsIntegerFormat`.

 ## ezImage
 * Fix wrong mappings in GalFormatToImageFormat
 * Added image format conversions to make it possible to convert anything to `ezImageFormat::R32G32B32A32_FLOAT`.

 ## Vulkan
 * If both discard and clear is set on a `ezGALRenderingSetup`, clear is used instead of discard to match the DX11 behaviour which does not support discard.
 * Device capabilities are clamped to their target type to prevent overflow, e.g. for `m_uiUAVCount`.

 ## Unit Test Framework
 * On windows, added **optipng** compression to the generation of failure images.
 * Reordered test framework init to allow tests to init the renderer during SetupSubTests.

 ## Renderer Unit Tests
 * Moved lots of duplicated code into the base class.
 * Reference images are now rendered in such a way that U goes from left to right and V from bottom to top in the reference images.
 * Made the GenerateMipMaps reference image for sRGB textures that correctly interpolates in linear space the main reference image as the implementations that are correct outnumber for broken ones that interpolate in gamma space.
 * Added override folder for **D3D11AMD**.
 * Added new `ezRendererTestReadback`. This test reads back all renderable texture formats, uploads them again to a second texture and compares both the readback and re-upload to reference images.


jankrassnigg:
> On windows, added optipng compression to the generation of failure images.

Why? We don't want to check them in, so why waste time on extra compression?

Sanakan8472:
> > On windows, added optipng compression to the generation of failure images.
> 
> Why? We don't want to check them in, so why waste time on extra compression?

I can't use the CopyReferenceImages functionality as renderer images tend to branch of into multiple override folders to accommodate various renderer quirks like line rendering. So I exclusively copy the files by hand. The same applies when I write unit tests. I may end up with a bunch of images, but only some are correct, the rest is running into bugs that need to be fixed first.

The code is only run when an image comparison fails or no reference image is available so it doesn't impact normal test runs. The time it takes to run the tool in the test failure case is neglectable and shouldn’t be noticeable to the end user.

PR #1028, opened 2023-08-19
Asset Curator fixes for external file operations

Sanakan8472:
## Asset Curator
* Copy, rename, overwrite and move file are now correctly handled and tested.
* Added **AssetMoved** event type.
* Renaming a file that is open as a document updates the document in place now.
* As the file system model is multi-threaded, any code that was touching **ezDocument** was moved to the main thread.

## File System Model
* Fixed deadlock due to events firing under the lock by changing to **ezCopyOnBroadcastEvent** type.
* Linking a document that is already linked will first fire **DocumentUnlinked** event before firing **DocumentLinked**.
* **ReadDocument** no longer requires setting the document GUID, this was removed as it caused linkege that needed to be immediately removed again, e.g. during duplicate asset detection. Now linking is always done outside the code when neccessary.
* Fixed a crash when **g_PostponedFiles** was resized while firing events.

## Misc Changes
* Fixed debug asset due to missing unregister of **ezHashedString** property grid GUI factory.
* ezDocument now has a **DocumentRenamed** function and event type.
* Sometimes moving a file triggers a modified watcher event on the source file name. This would make the removal of the file detected before the addition at the new location. To mitigate, modified events are now also queued up for 10 frames before firing like delete events.

### DIFF
--- a/Code/Editor/EditorFramework/Assets/Implementation/AssetUpdates.cpp
+++ b/Code/Editor/EditorFramework/Assets/Implementation/AssetUpdates.cpp
@@ -118,14 +118,15 @@ static ezResult PatchAssetGuid(ezStringView sAbsFilePath, ezUuid oldGuid, ezUuid
   if (ezDocumentManager::FindDocumentTypeFromPath(sAbsFilePath, true, pTypeDesc).Failed())
     return EZ_FAILURE;
 
-  if (ezDocument* pDocument = pTypeDesc->m_pManager->GetDocumentByPath(sAbsFilePath))
-  {
-    pTypeDesc->m_pManager->CloseDocument(pDocument);
-  }
-
   ezUInt32 uiTries = 0;
-  ezStringBuilder sTemp = sAbsFilePath;
-  while (pTypeDesc->m_pManager->CloneDocument(sTemp, sTemp, newGuid).Failed())
+
+  ezStringBuilder sTemp;
+  ezStringBuilder sTempTarget = ezOSFile::GetTempDataFolder();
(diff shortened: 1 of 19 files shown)


Code/Editor/EditorFramework/Assets/Implementation/AssetUpdates.cpp
```diff
@@ -208,31 +212,38 @@ ezResult ezAssetCurator::EnsureAssetInfoUpdated(ezStringView sAbsFilePath, const
...
-    SetAssetExistanceState(*pCurrentAssetInfo, ezAssetExistanceState::FileModified);
+  if (newExistanceState != ezAssetExistanceState::FileAdded)
+    SetAssetExistanceState(*pCurrentAssetInfo, newExistanceState);
   UpdateSubAssets(*pCurrentAssetInfo);
 
   InvalidateAssetTransformState(newGuid);
-
+  if (pFiles->LinkDocument(sAbsFilePath, pCurrentAssetInfo->m_Info->m_DocumentID).Failed())
+  {
+    // #TODO_ASSET
```
jankrassnigg: still todo ?

jankrassnigg (approved):

Code/Editor/EditorFramework/Assets/Implementation/AssetUpdates.cpp
Sanakan8472: Probably not necessary anymore. This can only fail of the file is no longer part of the model, but as we are in all callback for this specific file, this shouldn't be possible.

PR #1031, opened 2023-09-06

C-Core:
* Math changes for PhysX
* Removed center parameter on MakeInvalid function on bounding box sphere class, the initial center is NOT included when calling ExpandToInclude
* Fixed UI colors in node editor
* Include the type when hashing and comparing variants as keys in a hash table
* Ensure that functions with a zero interval are called every frame when using the interval scheduler
* Made abstract property SetValuePtr function const so you can set a value with a const property (setting a value does not change the property itself)
* Better search filter in node creation menu, component creation menu and asset browser search bar. It now basically works like the symbol/file search in Visual Assist.


jankrassnigg (approved):
Very nice.

PR #1035, opened 2023-09-09
Shared texture support and other side-effects of linux editor development

Sanakan8472:
# Foundation
* Some Natvis fixes.
* Format string argument `ezArgErrno` added for linux errors. Works just like `ezArgErrorCode` on Windows and prints the error code and human readable error message. Pass in `errno` into it.
* `ezIpcChannel` was partially split into `ezIpcProcessMessageProtocol` and unit tests were added for the former. `ezIpcProcessMessageProtocol` is tested indirectly by the editor and shared texture rendering tests.
* `ezIpcChannel` can now be stored in an `ezUniquePtr`.
* `ezIpcChannelEvent::Type` enum was reduced to just `Connected` instead of `ConnectedToClient` and `ConnectedToServer`, same for `Disconnected`.
* Return value of `ezIpcChannel::Send` is now passed upwards into the editor abstraction layers.
* Fixed crash in `~ezQtEngineViewWidget` while the engine process is dead.

# Editor
* Engine process now renders the exact content requested before the `ezSyncWithProcessMsgToEngine` message. Before this the message loop could potentially have already processed messages from the next frame, causing editor and engine to desync slightly, causing stuttering as frames were skipped. To make this cleaner `ezGameApplicationBase::Run` was split into `ezGameApplicationBase::RunOneFrame` so it can be called independently.

# Renderer
* Shared texture support added. Allows 2D textures to be shared between two DX11 or Vulkan devices across processes.
  * Check for support by querying `ezGALDeviceCapabilities::m_bSharedTextures`.
  * Use `ezGALDevice::CreateSharedTexture` to create one. Then call `ezGALDevice::GetSharedTexture` on the handle to get the interface `ezGALSharedTexture`. Use `ezGALSharedTexture::GetSharedHandle` to get a `ezGALPlatformSharedHandle` which can be sent to another process to then call `ezGALDevice::OpenSharedTexture` on it. Shared textures need to be destroyed via `ezGALDevice::DestroySharedTexture`.
  * Use `ezGALSharedTexture::WaitSemaphoreGPU` and `ezGALSharedTexture::SignalSemaphoreGPU` to syncronize access to the shared resource. The starting semaphore value of a texture is always 0.
  * Added `ezGALSharedTextureSwapChain` to simplify rendering into a set of shared textures.
  * Added a unit test that renders out of process and composits the result into a window swapchain.
* Cleaned up texture creation logic under both DX11 and Vulkan.
* Extracted reflection of RendererFoundation into `RendererReflection.h`. Reflection should not be needed for normal rendering code. Only include this if you need to write tools code or log nice enum names.
* Render pipeline outputs passes that could not be initialized.
* Vulkan
  * `ezMemoryAllocatorVulkan` was extended to support shared textures which require a dedicated pool.
  * Special care need to be taken when destroying shared resources. `ezGALDeviceVulkan` logic for pending deletions has been extended to support external memory as well as file descriptors under linux.
  * Added `ezGALDeviceVulkan::AddWaitSemaphore` and `ezGALDeviceVulkan::AddSignalSemaphore` these functions replace the additional parameters for syncronization on `ezGALDeviceVulkan::Submit` which is now a void parameter function.
  * Added `ezVulkanDispatchContext`. This contains extension function dispatch targets. Eventually all extension methods will have their implementation moved here instead of `DeviceVulkan.cpp`.


Code/Engine/Foundation/Communication/IpcProcessMessageProtocol.h
```diff
@@ -0,0 +1,39 @@
...
+
+#include <Foundation/Basics.h>
+#include <Foundation/Communication/RemoteInterface.h>
+#include <Foundation/Communication/RemoteMessage.h>
+#include <Foundation/Types/UniquePtr.h>
+
+class ezIpcChannel;
+class ezMessageLoop;
+
+class EZ_FOUNDATION_DLL ezIpcProcessMessageProtocol
```
jankrassnigg, line 17: A comment would be nice, what this protocol thing is supposed to do. From the interface and members, it just looks like a queue.

Code/Engine/Foundation/Strings/Implementation/FormatStringArgs.h
```diff
@@ -146,6 +146,19 @@ EZ_FOUNDATION_DLL ezStringView BuildString(char* szTmp, ezUInt32 uiLength, const
 
 #endif
 
+#if EZ_ENABLED(EZ_PLATFORM_LINUX)
+struct ezArgErrno
```
jankrassnigg, line 157: Comment for when and where to use this ?

jankrassnigg (approved):

Code/Engine/Foundation/Strings/Implementation/FormatStringArgs.h
Sanakan8472: I added docs to both the windows and linux version.

Code/Engine/Foundation/Communication/IpcProcessMessageProtocol.h
Sanakan8472: Extended comments on ezIpcChannel and this class. This was basically split out from ezIpcChannel to allow different protocols to be sent over the raw byte channel.

PR #1036, opened 2023-09-10
Added different types of following modes for FollowPathComponent

### LINKED ISSUE #947: FollowPathComponent: Mode for position only
Currently the follow path component (https://ezengine.net/pages/docs/animation/paths/follow-path-component.html) sets the full transform on the object.

For things like moving platforms it would be useful to have a mode that only sets the position, but keeps the rotation unchanged.
Another nice feature would be to set the position and the rotation, but always keep the object upright.

So for example for a moving platform, you want a platform to stay flat, so that one can stand on it, but when it goes around a curve, the platform should rotate around the up-axis.

See this branch for some preliminary work: [user/jk/follow-path-ext](https://github.com/ezEngine/ezEngine/commits/user/jk/follow-path-ext)

DavidJones011:
- PositionOnly (move the object along the path with only the position being altered)
- AlignUpZ (keeps the object aligned with the path up direction)
- FullRotation (moves the object along the path, matching its transform with the sampled path transform)

Related to Issue: **FollowPathComponent: Mode for position only #947**

https://github.com/ezEngine/ezEngine/assets/90158160/eed0df36-4390-4871-977c-186490f95a84

### DIFF
--- a/Code/Engine/GameEngine/Animation/FollowPathComponent.h
+++ b/Code/Engine/GameEngine/Animation/FollowPathComponent.h
@@ -67,13 +67,13 @@ public:
   /// \brief Sets the reference to the game object on which an ezPathComponent should be attached.
   void SetPathObject(const char* szReference); // [ property ]
 
-  ezEnum<ezPropertyAnimMode> m_Mode;     ///< [ property ] How the path should be traversed.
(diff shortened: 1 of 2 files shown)


jankrassnigg:
Very nice! Thanks for taking care of this.
Fixes: #947

PR #1042, opened 2023-09-15
MSFT contributions September 2023

mafaldag1:
- Improvements to the test framework
- Other various small fixes

### DIFF
--- a/Code/Engine/Core/System/Implementation/Win/Window_win32.inl
+++ b/Code/Engine/Core/System/Implementation/Win/Window_win32.inl
@@ -183,7 +183,7 @@ ezResult ezWindow::Initialize()
   SetWindowLongPtrW(windowHandle, GWLP_USERDATA, reinterpret_cast<LONG_PTR>(this));
 
   // show window and activate if required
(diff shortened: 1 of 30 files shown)


jankrassnigg (approved):

Ingrater:
@mafaldag1 the linux build fails with a memory corruption `malloc(): smallbin double linked list corrupted` very likely this memory corruption is also present in windows builds, but doesn't lead to a test failure there. I would suggest that you turn on the guarding allocator in UserConfig.h and do a run on windows to find any potential memory corruptions.

jankrassnigg:
I can take another look next week. Thanks for the tip.

Ingrater:
@jankrassnigg I think the memory corruption is only a side effect. It seems that the task system test deadlocks. This only happens on CI and not locally, due to the number of processors available. I think it would be best if we could debug this directly on the CI machine. @Sanakan8472

jankrassnigg:
@Ingrater I think this branch needs to be rebased, because we've already submitted the change to the JSON reader. If this PR doesn't produce conflicts, it will most likely still do unintended changes to that interface.

PR #1044, opened 2023-09-17
Clang Tidy dependency analysis

Ingrater:
* When a change contains header files, we now use the `DependencyAnalysis` tool in the clang-tidy CI to find all `.cpp` files which directly or indirectly contain this changed header files. Clang Tidy then checks all the found `.cpp` files on top of all changed `.cpp` files. This way Clang Tidy no longer misses certain errors, if only header files changed.
* The `DependencyAnalysis` tools was written to quickly find all header dependencies of every `.cpp` file. Doing so would typically require a full rebuild and then reading the dependency files which results from this. Instead of a full rebuild the `DependencyAnalysis` tool collects all header dependencies in one second (on my dev machine).
* Fixed `ezTokenizer` so it can parse integer literals like `100'000`
* Enable ezJSONParser and ezJSONReader to parse json documents which start with an array. Also add unittests for this case. This conflicts with the microsoft contributions PR.
* ezConversionUtils now correctly parse integers and floats with c++ digit separators: `100'000` and `100'000.0`


Code/BuildSystem/CMake/BuildFilters/PrecompiledTools.BuildFilter
```diff
@@ -27,4 +27,8 @@ set_property(GLOBAL PROPERTY EZ_BUILD_FILTER_PrecompiledTools
 	Utilities
 	ads
 	tinyexr
+    DependencyAnalysis
+    Lua
+    Duktape
+    Imgui
```
jankrassnigg, line 33: Why do we have these dependencies in the precompiled tools? Was that always the case?

Code/Engine/Foundation/CodeUtils/Implementation/Tokenizer.cpp
```diff
@@ -367,7 +367,7 @@ void ezTokenizer::HandleNumber()
   {
     NextChar();
 
-    while (ezStringUtils::IsDecimalDigit(m_uiCurChar))
+    while (ezStringUtils::IsDecimalDigit(m_uiCurChar) || m_uiCurChar == '\'') // integer literal: 100'000
```
jankrassnigg, line 370: isn't that very C++ specific ? where did you run into this being a problem ?

jankrassnigg (approved):

PR #1045, opened 2023-09-17
Remove a use of regex compare on path in CMake

ananace:
This could cause issues with building - and using - the engine under paths that contain regex tokens.

As an example, CMake entirely fails to generate a build system if the path contains repeated matching characters; (`*?+`)
```console
-- Shared Library: Duktape
RegularExpression::compile(): Nested *?+.
RegularExpression::compile(): Error in compile.
CMake Error at Code/BuildSystem/CMake/ezUtils.cmake:229 (if):
  if given arguments:

    "/Components/C++/ezEngine/Code/ThirdParty/Duktape" "MATCHES" "/Components/C++/ezEngine/"

  Regular expression "/Components/C++/ezEngine/" cannot compile
Call Stack (most recent call first):
  Code/BuildSystem/CMake/CMakeUtils/ezUtilsTarget.cmake:98 (ez_set_project_ide_folder)
  Code/ThirdParty/Duktape/CMakeLists.txt:10 (ez_create_target)


-- Configuring incomplete, errors occurred!
```


jankrassnigg:
Thanks for the fix :)

jankrassnigg:
@ananace I had to revert this change, because it changed the folder structure in the Visual Studio solution.  I assumed the changed CMake code would have the same result, but that's apparently not so. Maybe you can have another look.

ananace:
Interesting, I'll have a look when I have a moment, it really should've had the same result after all.

PR #1046, opened 2023-09-18
Fixes when EZ is used as submodule

C-Core:
* Added a EZ_ROOT cmake variable that should be used instead of ${CMAKE_SOURCE_DIR}/${EZ_SUBMODULE_PREFIX_PATH}
* Also changed ezCMakeConfig.cmake so it doesn't need modifications when ez is used as submodule


ezCMakeConfig.cmake
```diff
@@ -11,7 +11,7 @@ set_property(GLOBAL PROPERTY EZ_DEV_BUILD_LINKERFLAGS "DEBUG")
 set_property(GLOBAL PROPERTY EZ_CMAKE_RELPATH "Code/BuildSystem/CMake")
 set_property(GLOBAL PROPERTY EZ_CMAKE_RELPATH_CODE "Code")
 
-set_property(GLOBAL PROPERTY EZ_CONFIG_PATH_7ZA "${CMAKE_SOURCE_DIR}/Data/Tools/Precompiled/7z.exe")
```
Ingrater, line 14: Is the absolute path not required here?

Ingrater (approved):

ezCMakeConfig.cmake
C-Core: I make it absolute in ezUtils.cmake where it is used: 
set(FULL_7ZA_PATH "${EZ_ROOT}/${EZ_CONFIG_PATH_7ZA}")
		execute_process(COMMAND "${FULL_7ZA_PATH}"

PR #1049, opened 2023-09-19
Less regex compares on path in CMake, v2

ananace:
A follow-up on #1045 - now instead using CMake's `cmake_path(IS_PREFIX)` function (added in 3.20, which is helpfully this projects minimum required) to check if `PROJECT_SOURCE_DIR` is under `CMAKE_SOURCE_DIR`.


ananace:
And looks like my editor stripped away some trailing whitespace, missed that.

jankrassnigg:
Gave this a try and it seems to work as desired. Thanks!

PR #1063, opened 2023-09-24
Fixed #1062: Blackboard interface improvement 

### LINKED ISSUE #1062: [BUG] Blackboard interface for RegisterEntry + SetEntryValue needs rethinking
RegisterEntry ONLY changes the value when the entry didn't exist yet.
SetEntryValue FAILS, if the entry didn't exist yet.

Thus a lot of code will either need to call both, or potentially fails to do the desired thing.

The separation has been done because of hashedstring vs temphashedstring interfaces. Maybe this needs to be thought over again.

jankrassnigg:


Code/EditorPlugins/VisualScript/EditorPluginVisualScript/VisualScriptClassAsset/VisualScriptClassAsset.cpp
```diff
@@ -20,7 +20,7 @@ EZ_BEGIN_DYNAMIC_REFLECTED_TYPE(ezVisualScriptClassAssetProperties, 1, ezRTTIDef
 }
 EZ_END_DYNAMIC_REFLECTED_TYPE;
 
-EZ_BEGIN_DYNAMIC_REFLECTED_TYPE(ezVisualScriptClassAssetDocument, 1, ezRTTINoAllocator)
+EZ_BEGIN_DYNAMIC_REFLECTED_TYPE(ezVisualScriptClassAssetDocument, 2, ezRTTINoAllocator)
```
C-Core, line 23: Why was this increase necessary?

Code/Engine/Core/Utils/Blackboard.h
```diff
@@ -130,22 +132,26 @@ class EZ_CORE_DLL ezBlackboard : public ezRefCounted
   ///  \brief Removes all entries.
   void UnregisterAllEntries();
```
C-Core: I would rename this function to RemoveAllEntries, the UnregisterEntry function to RemoveEntry and get rid of the RegisterEntry function. This is just confusing otherwise.

Code/Engine/Core/Utils/Blackboard.h
```diff
@@ -130,22 +132,26 @@ class EZ_CORE_DLL ezBlackboard : public ezRefCounted
...
+  /// \brief Sets the value of the named entry. If the entry doesn't exist, yet, it will be created with default flags.
   ///
   /// If the 'OnChangeEvent' flag is set for this entry, OnEntryEvent() will be broadcast.
-  /// However, if the new value is no different to the old, no event will be broadcast, unless 'force' is set to true.
+  /// However, if the new value is no different to the old, no event will be broadcast.
   ///
-  /// Returns EZ_FAILURE, if the named entry hasn't been registered before.
-  ezResult SetEntryValue(const ezTempHashedString& sName, const ezVariant& value, bool bForce = false);
+  /// For new entries, no OnEntryEvent() is sent.
+  void SetEntryValue(ezStringView sName, const ezVariant& value);
```
C-Core, line 141: This should take an ezHashedString for the name. This would allow you to prevent the string hashing by caching the keys in your code.

Code/Engine/Core/Utils/Blackboard.h
```diff
@@ -180,8 +186,9 @@ class EZ_CORE_DLL ezBlackboard : public ezRefCounted
   static ezBlackboard* Reflection_GetOrCreateGlobal(const ezHashedString& sName);
   static ezBlackboard* Reflection_FindGlobal(ezTempHashedString sName);
   void Reflection_RegisterEntry(const ezHashedString& sName, const ezVariant& initialValue, bool bSave, bool bOnChangeEvent);
-  bool Reflection_SetEntryValue(ezTempHashedString sName, const ezVariant& value);
+  void Reflection_SetEntryValue(ezStringView sName, const ezVariant& value);
```
C-Core, line 193: No need for this reflection wrapper anymore. This was only necessary because of the ezResult return value.

C-Core (changes requested):

Code/EditorPlugins/VisualScript/EditorPluginVisualScript/VisualScriptClassAsset/VisualScriptClassAsset.cpp
jankrassnigg: The type change for the script functions means that script assets get serialized differently. I also think that they MAY need to be retransformed (not 100% sure) to work at runtime. At least some of my scripts seemed not to work without it. This just ensures that this happens.

Code/Engine/Core/Utils/Blackboard.h
jankrassnigg: I've added an overload for both. The problem is, usually people access existing strings. If they don't cache their hashed string, they'll take the very slow string construction path every time. With the ezStringView overload, this is still reasonably fast.
Now, if they do want to optimize it, they can use a hashed string. I'd say that's the best of both worlds.

Code/Samples/RtsGamePlugin/GameMode/EditLevelMode/EditLevelMode.cpp
```diff
@@ -27,11 +27,11 @@ RtsEditLevelMode::RtsEditLevelMode()
   m_pBlackboard = ezBlackboard::Create();
   m_pBlackboard->SetName("EditLevelModel");
 
-  m_pBlackboard->RegisterEntry(s_sTeam, 0);
-  m_pBlackboard->RegisterEntry(s_sShipType, 0);
-  m_pBlackboard->RegisterEntry(s_sSelectKey, ezVariant());
-  m_pBlackboard->RegisterEntry(s_sCreateKey, ezVariant());
-  m_pBlackboard->RegisterEntry(s_sRemoveKey, ezVariant());
+  m_pBlackboard->SetEntryValue(s_sTeam, 0);
```
C-Core, line 30: This should be AssignEntryValue

Code/Samples/RtsGamePlugin/GameMode/EditLevelMode/EditLevelMode.cpp
```diff
@@ -64,9 +64,9 @@ void RtsEditLevelMode::SetupEditUI()
 {
   // Set blackboard values
   {
-    m_pBlackboard->SetEntryValue(s_sSelectKey, ezInputManager::GetInputSlotDisplayName(ezInputSlot_MouseButton0)).IgnoreResult();
-    m_pBlackboard->SetEntryValue(s_sCreateKey, ezInputManager::GetInputSlotDisplayName("EditLevelMode", "PlaceObject")).IgnoreResult();
-    m_pBlackboard->SetEntryValue(s_sRemoveKey, ezInputManager::GetInputSlotDisplayName("EditLevelMode", "RemoveObject")).IgnoreResult();
+    m_pBlackboard->SetEntryValue(s_sSelectKey, ezInputManager::GetInputSlotDisplayName(ezInputSlot_MouseButton0));
```
C-Core, line 67: Here also AssignEntryValue

Code/UnitTests/GameEngineTest/StateMachineTest/StateMachineBuiltinsTest.cpp
```diff
@@ -231,15 +231,15 @@ void ezGameEngineTestStateMachine::RunBuiltinsTest()
 
     {
       ezSharedPtr<ezBlackboard> pBlackboard = ezBlackboard::Create();
-      pBlackboard->RegisterEntry(sTestVal, 2);
-      pBlackboard->RegisterEntry(sTestVal2, 0);
+      pBlackboard->SetEntryValue(sTestVal, 2);
```
C-Core, line 234: And again AssignEntryValue

Code/Engine/Core/Utils/Blackboard.h
C-Core: Makes sense, but I don't like the different function names, since SetEntryValue now also works with a HashedString but you would be pick the "very slow" path this way. You would need to add the wrapper function again to resolve the overloads for reflection.

C-Core (approved):

PR #1065, opened 2023-09-25
Added a "global blackboard" component

### LINKED ISSUE #1061: [BUG] Global Blackboards retrain state across editor lifetime
Global blackboards are never cleared, thus state easily leaks between simulation runs.

Maybe we should clear them after one run.

jankrassnigg:
This component allows to modify global blackboards.

It will get or create a named global blackboard and then add or modify its values. You can choose whether it should just ensure that the entries exist, are always reset to a fixed state, or whether even the entire blackboard gets cleared, including other entries that aren't part of the blackboard template.

Since global blackboards aren't tied to scenes and don't get cleared during simulation runs or level switches, this component allows to bring them into a known state.

I originally intended to make this a mode of the ezBlackboardComponent, but it was quite difficult to ensure it wouldn't be buggy or with weird side effects when editing the settings in the editor, since you could switch between modes, change the name of the blackboard (which should switch to a different global blackboard) and so on. So I decided to use a dedicated component, where the effects of modifications are more obvious.

Also I've removed the option to set initial values. Global blackboards should use templates only. Since multiple components may modify the same global blackboard (ie in every level), it is less error prone to define the values in a common asset.

It would also be possible to remove the "initial value" feature from the local blackboard as well, and instead add a "blackboard override component", which modifies another blackboard (local or global).

Fixes #1061


C-Core (approved):

Code/Engine/GameEngine/Gameplay/BlackboardComponent.h
```diff
@@ -83,38 +79,132 @@ class EZ_GAMEENGINE_DLL ezBlackboardComponent : public ezComponent
...
+
+  ezBlackboardTemplateResourceHandle m_hTemplate;
+};
+
+//////////////////////////////////////////////////////////////////////////
+
+using ezLocalBlackboardComponentManager = ezComponentManager<class ezLocalBlackboardComponent, ezBlockStorageType::Compact>;
+
+/// \brief This component creates its own ezBlackboard, and thus locally holds state.
+class EZ_GAMEENGINE_DLL ezLocalBlackboardComponent : public ezBlackboardComponent
```
C-Core, line 104: I guess you also need to rename the icons for the local blackboard component?

PR #1071, opened 2023-09-30
Convert ezToolsFoundation to use ezStringView instead of const char*

Sanakan8472:
* Added `ezMakeQString` to easily convert an ezStringView into a QString
* Fixed OpenXR compile issues.


jankrassnigg (approved):
Boooring.
But thanks for making this a separate PR.

PR #1072, opened 2023-09-30
Substance texture integration

C-Core:
Integration of substance textures into the Editor's asset pipeline: The substance package asset references the sbs file and extracts all graphs and there outputs. It can be then configured which outputs should be exported as textures. The transform output is just regular texture 2Ds. There is no substance texture generation at runtime, the integration is only on the Editor side.
You need to have a valid installation of Substance Designer otherwise the asset transform will fail.


C-Core:
@Sanakan8472 Can you review the changes I did to the asset infrastructure?

Code/Editor/EditorFramework/Assets/Implementation/AssetDocument.cpp
```diff
@@ -578,12 +578,12 @@ ezTransformStatus ezAssetDocument::InternalTransformAsset(const char* szTargetFi
 
 ezString ezAssetDocument::GetThumbnailFilePath() const
 {
-  return static_cast<ezAssetDocumentManager*>(GetDocumentManager())->GenerateResourceThumbnailPath(GetDocumentPath());
+  return GetAssetDocumentManager()->GenerateResourceThumbnailPath(GetDocumentPath());
 }
 
-void ezAssetDocument::InvalidateAssetThumbnail() const
+void ezAssetDocument::InvalidateAssetThumbnail(ezStringView sThumbnailFile) const
```
Sanakan8472: Why not add `ezStringView sSubAssetName` as a parameter here as well as on `GetThumbnailFilePath` above to make it consistent with the functions on the asset docment manager?

Code/Editor/EditorFramework/Assets/Implementation/AssetCurator.cpp
```diff
@@ -854,14 +854,14 @@ ezAssetInfo::TransformState ezAssetCurator::UpdateAssetTransformState(ezUuid ass
         state = ezAssetInfo::TransformState::UpToDate;
         if (pTypeDescriptor->m_AssetDocumentFlags.IsSet(ezAssetDocumentFlags::SupportsThumbnail))
         {
-          if (!ezAssetDocumentManager::IsThumbnailUpToDate(sAssetFile, out_ThumbHash, pTypeDescriptor->m_pDocumentType->GetTypeVersion()))
+          if (!pManager->IsThumbnailUpToDate(sAssetFile, "", out_ThumbHash, pTypeDescriptor->m_pDocumentType->GetTypeVersion()))
```
Sanakan8472, line 863: This change and the one in `ClearAssetCaches` ignore the fact that sub-assets can have thumbnails. Is this intended? If not, we could add a flag the the asset type that its sub-assets have thumbnails as well and check all of them to determine the transform state?

Code/Editor/EditorFramework/Assets/Implementation/AssetDocumentManager.cpp
```diff
@@ -66,11 +66,20 @@ ezStatus ezAssetDocumentManager::ReadAssetDocumentInfo(ezUniquePtr<ezAssetDocume
...
-  ezStringBuilder sProjectDir = ezAssetCurator::GetSingleton()->FindDataDirectoryForAsset(sDocumentPath);
+  ezStringBuilder sRelativePath;
+  if (sSubAssetName.IsEmpty())
+  {
+    sRelativePath = sDocumentPath;
+  }
+  else
+  {
+    sRelativePath = sDocumentPath.GetFileDirectory();
+    sRelativePath.AppendPath(sSubAssetName);
```
Sanakan8472: There is no guarantee that the sub asset name is a valid path. I'd suggest adding ezPathUtils::MakeValidFilename to it to make sure this works on any sub-asset.

Code/Editor/EditorFramework/EditorApp/Projects.cpp
```diff
@@ -55,13 +55,18 @@ ezResult ezQtEditorApp::CreateOrOpenProject(bool bCreate, const char* szFile)
...
   {
-    ezStringBuilder sdkDir;
-    if (ezFileSystem::FindFolderWithSubPath(sdkDir, szFile, "Data/Base", "ezSdkRoot.txt").Succeeded())
+    ezStringBuilder sdkDirFromProject;
+    if (ezFileSystem::FindFolderWithSubPath(sdkDirFromProject, szFile, "Data/Base", "ezSdkRoot.txt").Succeeded())
     {
-      sdkDir.MakeCleanPath();
-      if (sdkDir != ezFileSystem::GetSdkRootDirectory())
+      sdkDirFromProject.MakeCleanPath();
+      sdkDirFromProject.TrimWordEnd("/");
```
Sanakan8472: As Jan recently mentioned to me, `TrimWordEnd` really only removes one word. In this case, it would be better to use `Trim(nullptr, "/")` to remove all slashes at the end.

Code/Editor/EditorFramework/IPC/EngineProcessConnection.cpp
```diff
@@ -431,15 +431,16 @@ void ezEditorEngineProcessConnection::Update()
 
 bool ezEditorEngineConnection::SendMessage(ezEditorEngineDocumentMsg* pMessage)
 {
-#ifdef __GNUC__
+#if EZ_ENABLED(EZ_COMPILER_GCC)
```
Sanakan8472: I added these to shutup clang under linux, but then windows started complaining about the GCC pragmas so I added more pragmas. However, the GCC pragmas are supported by GCC and clang, so ideally we would use a macro that works on both clang and gcc, do you have something clean in mind that we can use in general for these cases?

Sanakan8472 (approved):

PR #1073, opened 2023-10-01
Added ezDataDirPath to cache various file queries

Sanakan8472:
## ezDataDirPath
* The `ezDataDirPath` class stores the absolute path to a file or folder as well as a cache to various sub-paths: `GetDataDirParentRelativePath`, `GetDataDirRelativePath`, `GetDataDir` and `GetDataDirIndex`. This facilitiates and accelerates various path handlings for files inside data directories.
* This replaces the various paths inside the `ezAssetInfo` of the `ezAssetCurator`.
* Migration is pretty straight forward as the function naming is the same as the old members inside `ezAssetInfo`. The class also implicitly converts to `ezStringView` to remove the requirement to write lots of `GetAbsolutePath` calls when dealing with just the main part of the class.
* Added `ezCompareDataDirPath` which first sorts case-insensitive and then case-sensitive if necessary for a unique ordering. Use this comparator when sorting e.g. files on disk like they would appear in a windows explorer.
  * Special care needs to be taken when using lower-bound on such a sorted comparer to find all paths inside a case-matching folder for example as entries of different casings can interleave, see FileSystemModel.cpp.
* Serializing an de-serializing of the asset cache has become a bit more complicated as the paths need to be checked for validity and potentially rebuild after loading, see FileSystemModel.cpp. This updates the asset cache version to allow serializing this new class.


Code/EnginePlugins/OpenXRPlugin/init.cmake
```diff
@@ -2,7 +2,7 @@
 ### OpenXR support
 ######################################
 
-set (EZ_BUILD_OPENXR OFF CACHE BOOL "Whether support for OpenXR should be added")
+set (EZ_BUILD_OPENXR ON CACHE BOOL "Whether support for OpenXR should be added")
```
jankrassnigg, line 5: Is this change intended?

Code/Tools/Libs/ToolsFoundation/FileSystem/Implementation/FileSystemModel.cpp
```diff
@@ -490,20 +531,24 @@ ezResult ezFileSystemModel::ReadDocument(ezStringView sAbsolutePath, const ezDel
   if (!m_bInitialized)
     return EZ_FAILURE;
 
+  ezStringBuilder sAbsolutePath2(sAbsolutePath);
+  sAbsolutePath2.MakeCleanPath();
+  sAbsolutePath2.TrimWordEnd("/");
```
jankrassnigg: I wouldn't use this, but just Trim. TrimWord only removes one instance IIRC.

jankrassnigg (approved):
Looks good.

Code/EnginePlugins/OpenXRPlugin/init.cmake
Sanakan8472: Yes, OpenXR plugin is built by default now (which includes local builds and CI) so that it doesn't deteriorate anymore.

Code/Tools/Libs/ToolsFoundation/FileSystem/Implementation/FileSystemModel.cpp
Sanakan8472: Oh good point!

PR #1077, opened 2023-10-04
Default cpp sources compile fix and unit test

Sanakan8472:


Code/UnitTests/EditorTest/Project/Project.cpp
```diff
@@ -76,3 +95,38 @@ ezTestAppRun ezEditorTestProject::RunSubTest(ezInt32 iIdentifier, ezUInt32 uiInv
...
+
+ezTestAppRun ezEditorTestProject::CreateCppSolution()
+{
+  ezCppSettings cpp;
+  EZ_TEST_BOOL(cpp.Load().Failed());
+
+  EZ_TEST_BOOL(!ezCppProject::ExistsSolution(cpp));
+
+  cpp.m_sPluginName = "TestPlugin";
+#if defined(_MSC_VER)
```
jankrassnigg: Don't we have some ez macro for this ?

Also you don't need the ifdef, you can directly do the number comparison, macros that aren't defined always return the value zero.

jankrassnigg (approved):
Very nice, thank you.

Code/UnitTests/EditorTest/Project/Project.cpp
Sanakan8472: Replaced it with `#if EZ_ENABLED(EZ_COMPILER_MSVC)`.
I'm keeping the outer ifdef in preparation for linux generator support.

PR #1079, opened 2023-10-15
Added support for fetching sample projects from other git repos

jankrassnigg:
A sample project may be added as an "ezRemoteProject" file which specifies the URL from which to clone the project data.

The user can then select where to clone the project to and then it will be opened like any other regular project.


jankrassnigg:
Currently this just clones the default branch e.g. 'main'. I'm wondering whether I should add an option to clone either a specific branch or even a specific commit. Or maybe a 'tag' ? (which AFAIK is also just a branch).

Not sure whether any of that would really be used. Currently this would just always check out the latest version on the default branch.

PR #1082, opened 2023-10-24
Lensflares

C-Core:
Added a lens flare component:
image

Cleaned up sprite renderer a bit
Added shader files and other editor specific data files to the visual studio solution


jankrassnigg:
Very cool! I wanted this feature for ages 🙂
Now if you want to do me a favor, you could go over the classes and structures and see whether you could write some more documentation comments here and there. Then maybe I can copy and paste done of them for the docs pages. Also our rendering code could generally use some more documentation.

PR #1090, opened 2023-10-30
Various fixes

C-Core:
* Fixed rmlui blackboard binding with blackboard templates
* Use the world's reload functions for scripts instead of duplicating the same logic in the script world module. Also extra proxy allocator for scripting stuff.
* Fixed script update function running when the component was deactivated


Code/Engine/GameEngine/Gameplay/Implementation/BlackboardComponent.cpp
```diff
@@ -512,9 +512,9 @@ ezGlobalBlackboardComponent::ezGlobalBlackboardComponent(ezGlobalBlackboardCompo
 ezGlobalBlackboardComponent::~ezGlobalBlackboardComponent() = default;
 ezGlobalBlackboardComponent& ezGlobalBlackboardComponent::operator=(ezGlobalBlackboardComponent&& other) = default;
 
-void ezGlobalBlackboardComponent::OnActivated()
+void ezGlobalBlackboardComponent::Initialize()
 {
-  SUPER::OnActivated();
+  SUPER::Initialize();
```
jankrassnigg, line 537: This change would mean that a deactivated global BB still changes the state.

PR #1093, opened 2023-11-02
Force transforming collection assets is now handled in a more general way

jankrassnigg:
This code is now also executed on TransformAll not only when a scene is exported.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -12,7 +12,7 @@ indent_style = space
 trim_trailing_whitespace = true
 
 # MSVC specific
-cpp_generate_documentation_comments = doxygen_triple_slash
+cpp_generate_documentation_comments = none
 
 # additional settings for code files
 [*.{cpp,h,inl,c,d,qrc,lua,cs,ps1,sh}]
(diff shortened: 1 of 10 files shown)


.editorconfig
```diff
@@ -12,7 +12,7 @@ indent_style = space
 trim_trailing_whitespace = true
 
 # MSVC specific
-cpp_generate_documentation_comments = doxygen_triple_slash
+cpp_generate_documentation_comments = none
```
Sanakan8472, line 15: Why this change?

Sanakan8472 (approved):

.editorconfig
jankrassnigg: Because the default then is that when you type /// VS will auto complete it to /// @brief.
That was also the intention. However, we use /// \brief. So now I always need to modify it again, which is annoying.
So either we agree that using @ is also fine, then it is useful, or we just continue doing it manually. And I didn't have the energy for yet another discussion about these kind of things, I had enough of them the last weeks :D

Sanakan8472: Well Ok I have no strong feeling one way or another. We could just use `@`, the end result will be the same.

PR #1101, opened 2023-11-17
Updated Jolt and enabled shared objects on Linux

### LINKED ISSUE #949: Get JoltPhysics to compile as a shared object on Linux
It already works with GCC or Clang, but not the other (which we use on Linux).

Needs some additions to the export macros for friend declarations.
This can be prototyped in EZ, but should then be back-merged to JoltPhysics.

jankrassnigg:
Fixes #949


Sanakan8472 (approved):

Ingrater (approved):
Tested jolt on linux with these changes, everything works fine.

PR #1102, opened 2023-11-17
Re-implemented "Sort by Recently Used" asset browser feature

jankrassnigg:
This got removed during the asset browser refactor.


Code/Editor/EditorFramework/Assets/Implementation/AssetBrowserModel.cpp
```diff
@@ -59,6 +60,14 @@ struct FileComparer
       sSortB = b.m_sAbsFilePath.GetAbsolutePath().GetFileNameAndExtension();
     }
 
+    if (m_bSortByRecentlyUsed && pInfoA && pInfoB)
```
Sanakan8472: This will only work if all items are assets. What you probably want to do is:
1. If both assets, do this.
2. If one asset AND the asset has a valid LastAccess time, its smaller in order.

PR #1103, opened 2023-11-17
Fixed #1100: Newly created assets are now functional right away

### LINKED ISSUE #1100: [BUG] Newly created assets are broken
**Bug description:**
When creating a new asset via right click->new the asset document opens but it is in a weird state. Property changes have no effect on the underlying asset, e.g. the asset properties are always the default value.

**How to Reproduce:**
A clear description what steps to take to reproduce the issue.
1. Create a new texture asset via right click-> new... in the asset browser and give it a name.
2. The texture asset window opens, assign it a texture file and hit transform
3. The transform will fail stating that no input texture file has been assigned
4. Debugging into ezTextureAssetDocument::RunTexConv you can see that the ezTextureAssetProperties are still all default values, e.g. no input texture file has been assigned allthough the UI shows an input file assigned.
5. After closing and re-opening the asset file all your properties changes are gone but this time it seems to behave as usual, you can change properties and they are correctly reflected in the ezTextureAssetProperties object.

jankrassnigg:


Code/Editor/EditorFramework/Assets/Implementation/AssetBrowserWidget.cpp
```diff
@@ -923,12 +923,19 @@ void ezQtAssetBrowserWidget::OnFileEditingFinished(const QString& sAbsPath, cons
...
-    if (iNewIndex != -1)
-    {
-      QModelIndex idx = m_pModel->index(iNewIndex, 0);
-      on_ListAssets_doubleClicked(idx);
-    }
+
+    // I don't understand why, but if we immediately execute on_ListAssets_doubleClicked() here, then the opened document will be in a broken state
+    // maybe something else has to happen on the main thread first ?
+    // using a delay (even of 0) fixes the problem
+    QTimer::singleShot(100, this, [this, sNewPath]()
```
Sanakan8472: Ok the actual problem is as follows:
1. We create an asset, a doc is created under the original filename.
2. We rename the assset right after.
3. The `ezAssetCurator` now has an asset renamed event queued but no main thread tick has flushed it yet.
4. `on_ListAssets_doubleClicked` is called which will eventuall call `ezQtEditorApp::GetSingleton()->OpenDocumentQueued`.

So we now have two conflicting events queued: Rename of old to new name, and open of new name.
If open of new name runs first, `ezDocumentManager::CreateOrOpenDocument` will think we are opening a doc that isn't open yet, because the rename hasn't propagated and thus we end up with another document under the same name.

A proper fix would be to call `ezAssetCurator::GetSingleton()->MainThreadTick(true);` right after `m_pModel->OnFileSystemUpdate();` inside `ezQtAssetBrowserWidget::OnFileEditingFinished`.

Not sure how we can prevent this race alltogether.

Sanakan8472 (changes requested):

PR #1108, opened 2023-11-19
GUIFoundation: Add Document Close Window Options.

TheophilusE:
Adds options to **Close All** and **Close All But This** window to document windows. There is a bug (not sure if its reported), but the last closed window is not properly serialized, which causes it to open undocked on reopening project. Adequate short cuts have been included (which I think are easy to use), feel free to suggest changes or future improvements.


jankrassnigg:
Looks good, thank you :)

PR #1118, opened 2023-11-26
Support converting between EZ and STL containers

jankrassnigg:


Code/BuildSystem/CMake/CMakeUtils/ezUtilsCppFlags.cmake
```diff
@@ -307,7 +307,7 @@ endfunction()
 function(ez_set_build_flags TARGET_NAME)
 	ez_pull_compiler_and_architecture_vars()
 
-	set_property(TARGET ${TARGET_NAME} PROPERTY CXX_STANDARD 17)
+	set_property(TARGET ${TARGET_NAME} PROPERTY CXX_STANDARD 20)
```
Ingrater, line 313: Using C++20 immediately pulls in all the `<ranges>` stuff, so our compile times are going up significantly. I would not do this, without doing appropriate counter measures.

Code/Engine/Foundation/Strings/StringView.h
```diff
@@ -2,6 +2,7 @@
 
 #include <Foundation/Strings/Implementation/StringIterator.h>
 
+#include <string_view>
```
Ingrater, line 5: The advantage of ez so far was, that we used almost no stl, and thus we were not affected form any of the STL code bloat. If we now start including things like <string_view> in a header file as central as StringView.h, we suddenly get all the STL code bloat. If we absolutely need this, it should be behind a compile flag, so it can be turned off, if you don't need it.

Ingrater: My main fear is, that this is undoing most of my past efforts to make ez compile quickly.

jankrassnigg: We already include <string> in those low level files. <string_view> should be relatively unproblematic. A compile flag is in practice not maintainable.

Code/BuildSystem/CMake/CMakeUtils/ezUtilsCppFlags.cmake
jankrassnigg: We have to upgrade at some point. It is unrealistic to stay at 17 forever. We could even upgrade to 23 now, though one thing at a time. I don't know what counter measures you are talking about. We include as little of the STL as possible already, but there are limits to what is practical.

jankrassnigg: Also just for reference, with C++ 17 a clean rebuild of EZ takes on my machine 2 minutes and 32 seconds. With these changes (C++ 20 + string_view include) it takes 2 minutes and 32 seconds.

Ingrater: C++20 does not give us any significant benefits other than the spaceship operator. But it pulls in all the `<ranges>` stuff immediately, because that was added in C++20. To keep sensible compile times, you would have to move ez to modules, but that doesn't work properly until C++23, and even with the current state of C++23 not all compilers support modules. In my personal opinion we have to wait until all compilers we use support modules in C++23 properly.

Code/Engine/Foundation/Strings/StringView.h
Ingrater: No, unfortunately not. As soon as you switch C++20 on, **everything** from STL include `<ranges>` and bloats your compile times.

Code/BuildSystem/CMake/CMakeUtils/ezUtilsCppFlags.cmake
Ingrater: Take a look at this: https://www.reddit.com/r/cpp/comments/o94gvz/what_happened_with_compilation_times_in_c20/

Code/Engine/Foundation/Strings/StringView.h
Sanakan8472: Can't we make this a separate file and use the `MakeX` pattern for interop? Like `ezSTL::MakeString`, `ezSTL::MakeStlString`, `ezSTL::MakeStringView`, `ezSTL::MakeStlStringView` etc. There shouldn't be a need for putting this inside the class outside of implicent conversion convenience?

jankrassnigg: That's exactly what I'm going for (convenience).

Code/BuildSystem/CMake/CMakeUtils/ezUtilsCppFlags.cmake
jankrassnigg: I don't see this happening in practice.

Ingrater: I would still greatly appreciate it if you pull the C++20 switch out into a sepearte PR, given how many things it breaks at the moment. Essentially everything but windows is now broken due to this change. And for that it seems a bit strange to "hide" it in a PR that doesn't even have C++20 in the title.

jankrassnigg: This is a draft to see what happens. C++ 20 is unfortunately needed for std::span, which I also intend to add support for. Also C++ 20 breaks all sort of stuff because suddenly C++ has a notion of Unicode. 30 years too late, and obviously horrible (they added char8_t which just sucks). It also broke all the operator== stuff. So I'm looking into this.

Overall C++ is just awful and getting worse with every version. But not upgrading is also not a solution.

I just installed WSL to compile Linux locally. Man, I'm already pissed at C++, but working with Linux is even worse.

Sanakan8472: What about platform support? Are we sure all platforms actually support C++ 20 ?
Wouldn't this also force all consumers of EZ engine to also upgrade their own projects and all of their dependencies, if that is even possible in many cases?
I'd rather not force everyone onto a new standard just because. There seems to be very little to gain here for an astronimical price.

jankrassnigg: I can probably do this with a compiler switch. But maybe you wait for when this is not a draft anymore, because I only opened the draft PR to see what the compiler result is on all platforms.

PR #1121, opened 2023-12-01
Improvements and fixes

C-Core:
* Better sampling patterns for procedural placement
* Improvements and fixes in the Substance integration
* Reduced lag between virtual character and presence body by one frame (they are still one frame off)
* Grab point debug visualization
* Fixed asset property widget text with subassets
* Fixed asset property widget text when the asset is missing
* Fixed deadlock in saving assets with subassets


C-Core:
@Sanakan8472 can you take a look at the asset related changes?

Code/Editor/EditorFramework/PropertyGrid/Implementation/QtAssetLineEdit.cpp
```diff
@@ -100,10 +100,26 @@ void ezQtAssetLineEdit::paintEvent(QPaintEvent* e)
...
+
+    ezStringView sFinalText = sText;
+
+    if (m_pOwner->IsValidAssetType(sText))
+    {
+      if (const char* szPipe = sFinalText.FindLastSubString("|"))
+      {
+        sFinalText = ezStringView(szPipe + 1);
+      }
+      sFinalText = sFinalText.GetFileName();
```
Sanakan8472: Shouldn't this go into the else case? If we find a pipe, the rest is the sub-asset name so on that name the `GetFileName` operation might have unintended results?

Code/Engine/GameEngine/Gameplay/Implementation/GrabbableItemComponent.cpp
```diff
@@ -27,9 +35,16 @@ EZ_BEGIN_COMPONENT_TYPE(ezGrabbableItemComponent, 1, ezComponentMode::Static)
 {
   EZ_BEGIN_PROPERTIES
   {
+    EZ_ACCESSOR_PROPERTY("DebugShowPoints", GetDebugShowPoints, SetDebugShowPoints),
     EZ_ARRAY_ACCESSOR_PROPERTY("GrabPoints", GrabPoints_GetCount, GrabPoints_GetValue, GrabPoints_SetValue, GrabPoints_Insert, GrabPoints_Remove),
```
Sanakan8472: Nit: If none of the functions actually do anything besides the operation on the array, you might as well use `EZ_ARRAY_MEMBER_PROPERTY("GrabPoints", m_GrabPoints),` and remove all the functions.

Code/Engine/GameEngine/StateMachine/Implementation/StateMachine.cpp
```diff
@@ -350,6 +350,9 @@ ezResult ezStateMachineInstance::SetState(ezStateMachineState* pState)
     return SetState(pState->GetNameHashed());
   }
 
+  if (m_pCurrentState == pState)
+    return EZ_SUCCESS;
```
Sanakan8472: Why is this not moved to the start of the function? From the looks of it, we will always execute `SetState` no matter what, wo it's not a real early out? I also don't quite get this function, it seems to be doing most of the stuff that `SetStateInternal` does, but only if `pState == nullptr` in which case, should we be doing anything at all? Or if `m_pDescription == nullptr` in which case, shouldn't we fail the function immediately instead of returning success?

Code/Editor/EditorFramework/PropertyGrid/Implementation/QtAssetLineEdit.cpp
C-Core: Might be the safer option, done that.

Code/Engine/GameEngine/StateMachine/Implementation/StateMachine.cpp
C-Core: Moved it to the beginning of the function. A state machine without description is a valid use case. You will have a description when setting up your state machine through an asset, but you can also create your states and the state machine manually in code.

Sanakan8472 (approved):

PR #1122, opened 2023-12-05
Code fixes and C++ 20 preparations

jankrassnigg:
This is the minimum set of changes needed to make it possible to compile EZ with C++ 20.
This PR does NOT enable C++ 20. Nor is this planned. However, if someone decides to use C++ 20 in their own project, it should still compile. Which it wouldn't do without these fixes. Because C++ 20 does away with all pretense to be a backward compatible language, it just flat out breaks everything.


Code/Engine/Foundation/Containers/ArrayBase.h
```diff
@@ -31,18 +31,16 @@ class ezArrayBase
...
   bool operator==(const ezArrayBase<T, Derived>& rhs) const; // [tested]
-
-  /// \brief Compares this array to another contiguous array type.
-  bool operator!=(const ezArrayBase<T, Derived>& rhs) const; // [tested]
+  EZ_ADD_DEFAULT_OPERATOR_NOTEQUAL(const ezArrayBase<T, Derived>&);
 
   /// \brief Compares this array to another contiguous array type.
   bool operator<(const ezArrayBase<T, Derived>& rhs) const; // [tested]
 
+#if EZ_DISABLED(EZ_USE_CPP20_OPERATORS)
```
jankrassnigg, line 39: In C++ 20 having all these overloads actually creates ambiguous situations. In theory C++ 20 is much cleaner in this regard. Unfortunately for C++ 17 it breaks without these overloads.

However, the C++ 20 situation still requires more investigation. Stuff's complicated.

Code/Engine/Foundation/Strings/HashedString.h
```diff
@@ -167,7 +163,7 @@ class EZ_FOUNDATION_DLL ezTempHashedString
 
   /// \brief Creates an ezTempHashedString object from the given string. Computes the hash of the given string during runtime, which might
   /// be slow.
-  ezTempHashedString(ezStringView sString); // [tested]
+  explicit ezTempHashedString(ezStringView sString); // [tested]
```
jankrassnigg, line 166: This was an oversight, that became apparent when I compiled with C++ 20 enabled.

Code/Engine/Foundation/Types/ArrayPtr.h
```diff
@@ -202,6 +202,7 @@ class ezArrayPtr
   }
 
   /// \brief Compares the two arrays for equality.
+  template <typename = typename std::enable_if<std::is_const<T>::value == false>>
```
jankrassnigg, line 205: Needed to make this one mutually exclusive with the other function that also uses enable_if.

Code/Engine/Foundation/Types/ArrayPtr.h
```diff
@@ -213,11 +214,26 @@ class ezArrayPtr
...
+#if EZ_DISABLED(EZ_USE_CPP20_OPERATORS)
+  template <typename = typename std::enable_if<std::is_const<T>::value == false>>
+  inline bool operator!=(const ezArrayPtr<const T>& other) const // [tested]
   {
     return !(*this == other);
   }
+#endif
+
+  /// \brief Compares the two arrays for equality.
+  inline bool operator==(const ezArrayPtr<T>& other) const // [tested]
```
jankrassnigg, line 226: Need an overload for comparing with non-const T. Not sure why we didn't need this before.

Code/UnitTests/FoundationTest/Basics/Types/IdTest.cpp
```diff
@@ -36,6 +36,7 @@ using LargeTestId = ezGenericId<32, 10>;
 EZ_WARNING_PUSH();
 EZ_WARNING_DISABLE_MSVC(4068);
 EZ_WARNING_DISABLE_GCC("-Wbitfield-constant-conversion");
```
jankrassnigg: Are GCC and Clang warnings identical? Or just some of them ?

Code/Engine/Foundation/Strings/HashedString.h
Ingrater: Can you get in a bit more detail on this please? Why is it good that you now have to manually call this constructor always?

Code/Engine/Foundation/Memory/Implementation/MemoryUtils_inl.h
```diff
@@ -377,19 +377,19 @@ EZ_ALWAYS_INLINE void ezMemoryUtils::Destruct(T* pDestination, size_t uiCount, e
...
 template <typename T>
-EZ_ALWAYS_INLINE void ezMemoryUtils::Destruct(T* pDestination, size_t uiCount, ezTypeIsClass)
+void ezMemoryUtils::Destruct(T* pDestination, size_t uiCount, ezTypeIsClass)
 {
   EZ_CHECK_CLASS(T);
 
-#define EZ_GCC_WARNING_NAME "-Waggressive-loop-optimizations"
-#include <Foundation/Basics/Compiler/GCC/DisableWarning_GCC.h>
+  EZ_WARNING_PUSH()
+  // EZ_WARNING_DISABLE_GCC("-Waggressive-loop-optimizations")
```
Ingrater: why is this commented out?

Code/UnitTests/FoundationTest/Basics/Types/IdTest.cpp
```diff
@@ -2,8 +2,12 @@
 
 #include <Foundation/Types/Id.h>
 
-#define EZ_MSVC_WARNING_NUMBER 4463
-#include <Foundation/Basics/Compiler/MSVC/DisableWarning_MSVC.h>
+EZ_WARNING_PUSH()
```
Ingrater, line 5: what about the matching `EZ_WARNING_POP()` ?

Ingrater (approved):

Code/Engine/Foundation/Strings/HashedString.h
jankrassnigg: Creating a TempHashedString isn't free (because of the hashing). Just implicitly converting to it shouldn't happen, because sometimes you may have other options (or you could just store a TempHashedString to begin with). So I think people should be aware of this.

Code/UnitTests/FoundationTest/Basics/Types/IdTest.cpp
jankrassnigg: There was already one at the end of the file. You just don't see it in the diff.

Code/Engine/Foundation/Memory/Implementation/MemoryUtils_inl.h
jankrassnigg: This was a test (successful!). The warning push/pop etc can be removed now.

Issue #1119, opened 2023-11-26
Support for "Platform Plugin" like feature for ezEngine


JailbreakPapa:
**Is your feature request related to a problem? Please describe:**
Platform Plugins would be a amazing feature to add with the engine.

While adding **Platform Blue Support (NDA Sorry)**, Manually searching through each foundation and core file to see what i would need to implement for Platform Blue support was very tiresome. and to do the same for both **Platform Green & Red** was also very inconvenient.  

**Describe the solution you'd like:**
A Plugin that Authorized Developers for said console can create/use that allows them to add all needed functionality to the engine for the platform, in a separate Project/Target. the engine should load this plugin first before any other system, just how any foundational system gets loaded within the engine.

**Additional context:**
Reference how Unreal Engine handles console platforms, they give you a folder that contains the platform support code, support for UBT for said platform, and allow you to create a "Platform" folder where it configures everything for you through UBT CS Files/Build-Time. the same could be done for CMake.

jankrassnigg:
The way it is currently done you should actually be able to find all places fairly easily just by compiling the code. Once you add a certain platform, the compiler generates errors for all places that aren't handled yet. So frankly I don't understand why it was very inconvenient. I recently did it for another platform and it took me maybe two hours to go through all those places.

The problem with all the platforms you mentioned is, that we have no access to them. We could probably make things a bit easier, but we don't know the pitfalls of each platform.

TheophilusE:
You can easily search for platform dependent code, the engine is modular enough that you only need to tackle your platform bits per library. I think its unreasonable to try and abstract more platform things that may not agree with the code structure.

PlasmaDev5:
I'm in favor of such a feature assuming it its not to destructive to what is there. I'm mainly thinking for distribution of NDA platforms. 

Say there is a user with less technical knowledge who built a simple game using visual script. Hypothetically the engine supports console and they are looking to port what they made. The last thing they want to compile engine source every time they add an NDA platform. 

Having these platforms as simple drag and drop plugins was vastly streamline this process. It would allow each of the platforms to be distributed easily and independently. I feel there is also an argument that it could help NDA security. None of there api is allowed public along with other contract restrictions, this would ensure its I never mixed with the public code preventing accidents. 

I'd imagine this would likely need some work on plugins to support external plugins easily among other changes

jayrulez:
You can look at O3DE as an example of how something like this could be organized.

[...]

PR #1124, opened 2023-12-07
Move platform implementations to central folder

jankrassnigg:
This is an idea, how we could make porting to different platforms easier. In response to https://github.com/ezEngine/ezEngine/issues/1119

Basically we set up include paths to the correct platform folder in the build system.

Additionally, to be able to still name the files "Stuff_Win.h", so that they are easy to distinguish, a macro is used to concatenate the include filename with a platform shortname.

We would then move all platform specific files to a single folder in each library and name them consistently.


Ingrater:
I'm not a big fan of hiding includes in preprocessor stuff. This makes tooling significantly harder and prevents searching directly for the file name in source code. I know of IDEs, code completion and static analysis tools that can't deal with such includes. Although for visual studio it should be fine.

If we decide to go through with this we need to fix the two of our tools this breaks: HeaderChecker and DependencyAnalysis

jankrassnigg:
The alternative would be to reuse the same name. E.g. have 5 different "Platform.h" files, just in different folders. I don't really like having same-named files, because it breaks tooling in other ways (jump to file, search for all platform X specific files) but it would be the canonical way.

Code/Engine/Foundation/FoundationPCH.cpp
```diff
@@ -121,11 +110,107 @@ EZ_STATICLINK_LIBRARY(Foundation)
...
+  EZ_STATICLINK_REFERENCE(Foundation_Platform_Android_ConditionVariable_Android);
+  EZ_STATICLINK_REFERENCE(Foundation_Platform_Android_CrashHandler_Android);
+  EZ_STATICLINK_REFERENCE(Foundation_Platform_Android_DirectoryWatcher_Android);
+  EZ_STATICLINK_REFERENCE(Foundation_Platform_Android_EnvironmentVariableUtils_Android);
+  EZ_STATICLINK_REFERENCE(Foundation_Platform_Android_GuardedAllocation_Android);
+  EZ_STATICLINK_REFERENCE(Foundation_Platform_Android_MemoryMappedFile_Android);
+  EZ_STATICLINK_REFERENCE(Foundation_Platform_Android_MiniDumpUtils_Android);
+  EZ_STATICLINK_REFERENCE(Foundation_Platform_Android_OSFile_Android);
+  EZ_STATICLINK_REFERENCE(Foundation_Platform_Android_OSThread_Android);
+  EZ_STATICLINK_REFERENCE(Foundation_Platform_Android_PageAllocator_Android);
```
Ingrater: All of these should not be neccessary. Maybe we need to teach the static link util to ignore platform specific files?

Code/Engine/Foundation/Platform/Android/CrashHandler_Android.cpp
```diff
@@ -0,0 +1,9 @@
+#include <Foundation/FoundationPCH.h>
+
+#if EZ_ENABLED(EZ_PLATFORM_ANDROID)
+#  include <Foundation/Platform/Posix/CrashHandler_Posix.h>
+#endif
+
+
+EZ_STATICLINK_FILE(Foundation, Foundation_Platform_Android_CrashHandler_Android);
```
Ingrater: I don't like that for every platform we our now pulling in all other platform cpp files through this mechanism. This just bloats the final binary size for no reason.

Code/Engine/Foundation/FoundationPCH.cpp
Ingrater: In general this static link stuff is only required for things that are only available through reflection. As soon as you have methods / classes which are dllexport they always end up in the final binary. BEcause this is the case for all these platform specific things, the staticlink reference stuff here is just additional bloat without function.

Code/Engine/Foundation/Platform/Android/Screen_Android.cpp
```diff
@@ -0,0 +1,15 @@
+#include <Foundation/FoundationPCH.h>
+
+#if EZ_ENABLED(EZ_PLATFORM_ANDROID)
```
Ingrater, line 3: What I don't like about this approach is that this results in many compilation units that are plain empty and serve no function, but will still be compiled. I would prefer a hybrid approach, where we do the preprocessor decision but also have something in cmake to not compile the cpp in the first place.

Ingrater (changes requested):
We could merge this like this because it allows external platform implementations, but I don't like that this new approach has significant disavantges over our old approach.

Code/Engine/Foundation/FoundationPCH.cpp
jankrassnigg: Deactivated for these files now.

Code/Engine/Foundation/Platform/Android/CrashHandler_Android.cpp
jankrassnigg: Not the case anymore.

Code/Tools/StaticLinkUtil/StaticLinkUtil.cpp
```diff
@@ -723,13 +712,35 @@ class ezStaticLinkerApp : public ezApplication
...
           }
+
+          if (sFileContent.FindSubString("EZ_STATICLINK_FILE_DISABLE"))
+            continue;
+
+          m_FilesToModify.Insert(sFile);
+
+          bool bContainsGlobals = false;
+
+          bContainsGlobals = bContainsGlobals || (sFileContent.FindSubString("EZ_STATICLINK_LIBRARY") != nullptr);
```
jankrassnigg, line 732: @Ingrater Reflection isn't the only thing that needs to be considered. We have a couple more globals, like CVars, plugin references and the startup blocks. Those are all that I could find atm. Of course now that means we need to keep an eye out for additional globals and make sure they get included as well.

PR #1125, opened 2023-12-11
Updated some const to use constexpr instead

adriengivry:
From the coding guidelines ("Static members" section):
> If the member is a constant, it should be marked _constexpr_

In alignment with the coding guidelines, this pull request focuses on optimizing code quality by identifying and addressing instances where certain `cons`t variables should be marked as `constexpr`. The changes introduced ensure consistency with the established coding standards and enhance the efficiency of the affected code segments.


jankrassnigg:
Thanks!

PR #1132, opened 2023-12-28
Prevent mouse-wheel from changing property values

RiederAlex:
Prevent mouse-wheel from changing property values
This just applies the fix done in 5f8a23d2e0b83deb179e806f610b25a205d2acd4 to more PropertyWidgets


jankrassnigg (approved):

jankrassnigg:
Good catch! Thanks for the contribution!

PR #1135, opened 2023-12-29
Fix assert in ezDocumentManager when modifying render pipeline assets

### LINKED ISSUE #1134: [BUG] Adding / Removing pins in render pipeline assets is broken
Changing the render pipeline graph connections crashes.

It works in visual scripts and visual shaders.

Ingrater:
Fixes issue #1134


Ingrater:
@Sanakan8472 can you please take a look. This fixes issue #1134 I'm not quite sure though if the fix is optimal.

PR #1138, opened 2023-12-30
Fix segfault in RecastAgentComponent

x0r04rg:
The `dtNavMeshQuery::raycast` method expects `dtRaycastHit::maxPath` field to be the size of a buffer given by `dtRaycastHit::path`. So if `maxPath` is > 0, it tries to write to `path`. But here `maxPath` is uninitialized (as well as `path`) which leads to a segfault.


jankrassnigg:
Thanks!

PR #1143, opened 2024-01-07
Support Editor Generated C++ Plugins on Linux

Ingrater:
* In addition to VS2022 we now also support clang as compiler for Editor generated C++ projects on Windows
* In addition to Visual Studio, we now also support Visual Studio Code as IDE for editing Editor generated C++ projects. Visual Studio Code works on both Linux and Windows.
* Instead of passing arguments directly to cmake.exe we now write a `CMakeUserPresets.json` which makes integrating with any IDE much simpler. Also the generation can be trivially triggered from the command line by executing `cmake --preset ezEngine`
* Support for Editor generated C++ plugins on Linux. Both Gcc and Clang are supported.
* Various pieces of code to auto detect the needed compiler executables.
* Code to check for correct compiler settings. The compiler must match the compiler the SDK was build with.
* Moved compiler and IDE settings to the editor preferences. This way they are machine local and no longer checked in alongside the project. Making it easier to share a project between different machines or even Linux / Windows.

image
image


Ingrater:
@jankrassnigg since the compiler settings are now stored in the preferences, how should the tests behave to not modify the users preferences?

jankrassnigg:
> @jankrassnigg since the compiler settings are now stored in the preferences, how should the tests behave to not modify the users preferences?

In test mode the flag ezQtEditorApp::StartupFlags::UnitTest is set, which already prevents the editor from saving preferences.

Data/Samples/PacMan/RuntimeConfigs/Plugins.ddl
```diff
@@ -23,8 +23,3 @@ Plugin
 	string %Path{"ezParticlePlugin"}
 	bool %LoadCopy{false}
 }
-Plugin
-{
-	string %Path{"ezXBoxControllerPlugin"}
-	bool %LoadCopy{false}
-}
```
jankrassnigg, line 30: This obviously removes a feature from the Windows version, which isn't great. Either we keep it and just ignore that on Linux you'll see a plugin load error message, or you make the plugin compile on Linux, even if it doesn't do anything, so that it's not a problem to keep this reference.
OR you make that plugin actually work on Linux, but no idea how complicated that is.

Code/UnitTests/FoundationTest/System/ProcessTest.cpp
```diff
@@ -9,7 +9,7 @@ EZ_CREATE_SIMPLE_TEST_GROUP(System);
 
 EZ_CREATE_SIMPLE_TEST(System, Process)
 {
-  EZ_TEST_BLOCK(ezTestBlock::Enabled, "Command Line")
```
jankrassnigg: Are all these changes intentional ?

Code/Editor/EditorFramework/EditorApp/Projects.cpp
```diff
@@ -304,19 +304,31 @@ void ezQtEditorApp::ProjectEventHandler(const ezToolsProjectEvent& r)
 
       if (m_StartupFlags.AreNoneSet(ezQtEditorApp::StartupFlags::Headless | ezQtEditorApp::StartupFlags::SafeMode | ezQtEditorApp::StartupFlags::UnitTest | ezQtEditorApp::StartupFlags::Background))
       {
-        if (ezCppProject::IsBuildRequired())
```
jankrassnigg, line 307: run formatter ?

Code/Editor/EditorFramework/EditorApp/Projects.cpp
```diff
@@ -304,19 +304,31 @@ void ezQtEditorApp::ProjectEventHandler(const ezToolsProjectEvent& r)
...
+        if(ezCppProject::ExistsProjectCMakeListsTxt())
         {
-          const auto clicked = ezQtUiServices::MessageBoxQuestion("<html>Compile this project's C++ plugin?<br><br>\
+          ezStatus compilerStatus = ezCppProject::TestCompiler();
+          if(compilerStatus.Failed())
+          {
+            ezQtUiServices::MessageBoxWarning(ezFmt("<html>The compiler preferences are invalid.<br><br>\
+              This project has <a href='https://ezengine.net/pages/docs/custom-code/cpp/cpp-project-generation.html'>a dedicated C++ plugin</a> with custom code.<br><br>\
+              The compiler set in preferences does not appear to work, as a result the plugin can not be compiled <br><br><b>Error:</b> {}</html>", compilerStatus.m_sMessage.GetView()));
```
jankrassnigg: Typo: "cannot" instead of "can not" 
"set in THE preferences"

Code/Editor/EditorFramework/CodeGen/CppProject.h
```diff
@@ -2,10 +2,92 @@
...
+{
+  EZ_ADD_DYNAMIC_REFLECTION(ezCppProject, ezPreferences);
+
+  struct MachineSpecificCompilerPaths
+  {
+    ezString sNiceName;
+    ezEnum<ezCompiler> eCompiler;
+    ezString sCCompiler;
+    ezString sCppCompiler;
+    bool bIsCustom;
```
jankrassnigg: Not typical ez coding style.

Code/Editor/EditorFramework/CodeGen/CompilerPreferencesWidget.cpp
```diff
@@ -0,0 +1,242 @@
+#include <EditorFramework/EditorFrameworkPCH.h>
+
+#include <EditorFramework/CodeGen/CompilerPreferencesWidget.moc.h>
+#include <EditorFramework/CodeGen/CppProject.h>
+#include <GuiFoundation/PropertyGrid/PropertyGridWidget.moc.h>
+#include <GuiFoundation/PropertyGrid/PropertyMetaState.h>
+#include <GuiFoundation/Widgets/CollapsibleGroupBox.moc.h>
+#include <ToolsFoundation/Object/ObjectAccessorBase.h>
+
+// TODO change class to derive from ezQtPropertyTypeWidget and only add combo box in constructor
```
jankrassnigg: Several TODOs in here that should be reviewed.

Data/Samples/PacMan/RuntimeConfigs/Plugins.ddl
Ingrater: This is not a intentional change. The editor simply removes this entry from the `Plugins.ddl` on Linux, because this plugin does not exist on Linux. Building a ezXBoxControllerPlugin doesn't make sense for Linux, as glfw already coveres all gamepad & joystick input. We need a concept of platform specific plugins.

Code/UnitTests/FoundationTest/System/ProcessTest.cpp
Ingrater: No, need to revert this.

Code/Editor/EditorFramework/CodeGen/CppProject.h
Ingrater: What exactly do you mean? We discussed member naming rules in depth, and public members don't need to start with `m_` we also do this in other places.

Data/Samples/PacMan/RuntimeConfigs/Plugins.ddl
jankrassnigg: Yeah, something like that. But I guess not in this PR. I would revert this change, though.

Code/Editor/EditorFramework/CodeGen/CppProject.h
jankrassnigg: I guess you are right, though I thought we only allow this because it would be idiotic for things like vec3. For more complex types we typically use m_. But I don't care that much TBH.

Ingrater: No strong feelings either, added `m_`

Code/Editor/EditorFramework/CodeGen/CompilerPreferencesWidget.cpp
```diff
@@ -0,0 +1,167 @@
...
+  gridLayout->setColumnMinimumWidth(1, 5);
+  gridLayout->setColumnStretch(2, 2);
+  gridLayout->setContentsMargins(0, 0, 0, 0);
+  gridLayout->setSpacing(0);
+
+  ezStringBuilder fmt;
+  QLabel* versionText = new QLabel(ezMakeQString(ezFmt("This SDK was compiled with {} version {}. Select a compatible compiler.",
+    ezCppProject::CompilerToString(ezCppProject::GetSdkCompiler()),
+    ezCppProject::GetSdkCompilerMajorVersion())
+                                                   .GetText(fmt)));
```
jankrassnigg: interesting formatting

Code/Editor/EditorFramework/CodeGen/CompilerPreferencesWidget.cpp
```diff
@@ -0,0 +1,167 @@
...
+      sCppCompiler = varCppCompiler.Get<decltype(sCppCompiler)>();
+    }
+
+    ezInt32 selectedIndex = -1;
+    const auto& machineSpecificCompilers = ezCppProject::GetMachineSpecificCompilers();
+    // first look for non custom compilers
+    for (ezUInt32 i = 0; i < machineSpecificCompilers.GetCount(); ++i)
+    {
+      const auto& curCompiler = machineSpecificCompilers[i];
+      if (curCompiler.bIsCustom == false && curCompiler.eCompiler == eCompiler && curCompiler.sCCompiler == sCCompiler && curCompiler.sCppCompiler == sCppCompiler)
```
jankrassnigg: more brackets! they have little runtime overhead ;-)

Code/Editor/EditorFramework/CodeGen/CompilerPreferencesWidget.cpp
```diff
@@ -0,0 +1,167 @@
...
+      {
+        selectedIndex = static_cast<int>(i);
+        break;
+      }
+    }
+
+    if (selectedIndex == -1)
+    {
+      // If we didn't find a system default compiler, look for custom compilers next
```
jankrassnigg, line 96: what is a custom compiler? and if the user specified one, wouldn't we always prefer that ?

Code/Editor/EditorFramework/CodeGen/CompilerPreferencesWidget.cpp
```diff
@@ -0,0 +1,167 @@
...
+void ezQtCompilerPreferencesWidget::on_compiler_preset_changed(int index)
+{
+  auto compilerPresets = ezCppProject::GetMachineSpecificCompilers();
+
+  if (index >= 0 && index < compilerPresets.GetCount())
+  {
+    const auto& preset = compilerPresets[index];
+
+    const auto& selection = m_pTypeWidget->GetSelection();
+    EZ_ASSERT_DEV(selection.GetCount() == 1, "This Widget does not support multi selection");
```
jankrassnigg, line 126: these asserts shouldn't be necessary, if you configure the Qt selection mode as single selection, which I guess you did

Code/Editor/EditorFramework/CodeGen/CompilerPreferencesWidget.cpp
```diff
@@ -0,0 +1,167 @@
...
+    m_pObjectAccessor->SetValue(obj, "CustomCompiler", preset.bIsCustom).AssertSuccess();
+    m_pObjectAccessor->SetValue(obj, "CCompiler", preset.sCCompiler).AssertSuccess();
+    m_pObjectAccessor->SetValue(obj, "CppCompiler", preset.sCppCompiler).AssertSuccess();
+    m_pObjectAccessor->FinishTransaction();
+  }
+}
+
+void ezCompilerPreferences_PropertyMetaStateEventHandler(ezPropertyMetaStateEvent& e)
+{
+  static const ezRTTI* pRtti = ezGetStaticRTTI<ezCompilerPreferences>();
```
jankrassnigg: No need for the static. Neither is this performance critical, nor is the lookup that slow.

jankrassnigg: Actually, this isn't even a lookup, ezGetStaticRTTI should be a compile time thing already.

Code/Editor/EditorFramework/CodeGen/CppProject.cpp
```diff
@@ -78,11 +237,161 @@ ezString ezCppProject::GetSolutionPath(const ezCppSettings& cfg)
...
+#  error Unsupported compiler
+#endif
+}
+
+ezStatus ezCppProject::TestCompiler()
+{
+  const ezCppProject* preferences = ezPreferences::QueryPreferences<ezCppProject>();
+  if (preferences->m_CompilerPreferences.m_eCompiler != GetSdkCompiler())
+  {
+    return ezStatus(ezFmt("The currently configured compiler is incompatible with this SDK. The SDK was build with {} but the currently configured compiler is a {}.",
```
jankrassnigg: Typo: "was built" instead of "was build" (also a couple more times below)
Also I think I would remove the last "a" before the brackets, and you could also put the brackets in single quotes '{}', if you want the compiler name to be more prominent.

Code/Editor/EditorFramework/CodeGen/CppProject.cpp
```diff
@@ -117,6 +426,79 @@ ezResult ezCppProject::CheckCMakeCache(const ezCppSettings& cfg)
...
+    ezFileReader fileReader;
+    if (fileReader.Open(configureJsonPath).Failed())
+    {
+      ezLog::Error("Failed to open '{}' for reading", configureJsonPath);
+      return ModifyResult::FAILURE;
+    }
+    ezJSONReader reader;
+    if (reader.Parse(fileReader).Failed())
+    {
+      ezLog::Error("Failed to parse json of '{}'", configureJsonPath);
```
jankrassnigg: typo: "JSON" not "json"

Code/Editor/EditorFramework/CodeGen/CppProject.cpp
```diff
@@ -321,54 +704,60 @@ ezResult ezCppProject::CompileSolution(const ezCppSettings& cfg)
...
-  ezLog::Dev("Running MSBuild: {}", sMsBuildCmd);
-
-  ezInt32 iReturnCode = 0;
+  ezStringBuilder sCMakeBuildCmd;
+  po.BuildCommandLineString(sCMakeBuildCmd);
+  ezLog::Dev("Running {} {}", cmakePath, sCMakeBuildCmd);
   if (ezProcess::Execute(po, &iReturnCode).Failed())
   {
-    ezLog::Error("MSBuild failed to run.");
+    ezLog::Error("Failed to start cmake.");
```
jankrassnigg: In user visible texts we should write out the names properly, ie "CMake", "JSON", "HoloLens" etc
In code, I don't care that much.

Code/Editor/EditorFramework/CodeGen/CppProject.cpp
```diff
@@ -515,3 +901,241 @@ bool ezCppProject::IsBuildRequired()
 
   return false;
 }
+
+namespace
+{
+  template <typename T>
+  T* Expect(ezVariantDictionary& inout_json, ezStringView sName)
```
jankrassnigg, line 908: that's quite useful, maybe we should add something like this directly to ezVariant

Code/Editor/EditorFramework/CodeGen/CppProject.cpp
```diff
@@ -515,3 +901,241 @@ bool ezCppProject::IsBuildRequired()
...
+#endif
+
+#if EZ_ENABLED(EZ_PLATFORM_LINUX) && EZ_ENABLED(EZ_COMPILER_GCC)
+  s_MachineSpecificCompilers.PushBack({"Gcc (Custom)", ezCompiler::Gcc, "", "", true});
+#endif
+
+  if (preferences->m_CompilerPreferences.m_eCompiler != sdkCompiler)
+  {
+    s_MachineSpecificCompilers.PushBack(
+      {"SDK Incompatible previously configured compiler",
```
jankrassnigg: Is this missing a "with" in the text ?

jankrassnigg (approved):

Code/Editor/EditorFramework/CodeGen/CompilerPreferencesWidget.cpp
Ingrater: A custom compiler is a compiler where the user manually specified the compiler path. This code is not about what we prefer though. For simplicity there is a dropdown in the widget which doesn't get serialized (because it's contents can change on very editor restart, so it doesn't make sense to serialize it). What this code does is, it reconstructs the dropdowns selected index from the information we did actually serialize. So this has nothing to do with what we prefer in terms of a compiler.

Code/Editor/EditorFramework/CodeGen/CompilerPreferencesWidget.cpp
Ingrater: Certinaly true, still better safe then sorry ;-)

Code/Editor/EditorFramework/CodeGen/CompilerPreferencesWidget.cpp
Ingrater: We never throw away the users settings. Even if the compiler he configured earlier is no longer avaiable, we still retain the settings he made.

Code/Editor/EditorFramework/CodeGen/CppProject.cpp
Ingrater: No. This piece of code auto detects available compilers on the system. If the currently serialized compiler is different from the compiler the SDK was built with we add a dummy compiler which exactly represents the serialized state. This is done so the user doesn't loose the configuration upon exiting the editor no matter if that compiler still exists or not. So what this means is "A compiler that was previously configured but is now incompatible with the current SDK build" -> "SDK incompatible previously configured compiler"

Code/Editor/EditorFramework/CodeGen/CppProject.cpp
Ingrater: Let's discuss this and I'll do it as a follow up PR

Code/Editor/EditorFramework/CodeGen/CppProject.cpp
jankrassnigg: This text still doesn't make sense to me. Wouldn't it then be "SDK COMPATIBLE previously configured compiler" ? As in the previously configured compiler WAS compatible (not INcompatible).
Maybe it would be better to rephrase this entirely: "Missing: Compiler the SDK was built with"
Or something similar.

Code/Editor/EditorFramework/CodeGen/CompilerPreferencesWidget.cpp
Ingrater: This is just your typical copy & paste mistake. I copied this from another event handler that used dynamic RTTI.

Code/Editor/EditorFramework/CodeGen/CppProject.cpp
Ingrater: Imagine the following scenario:

1) User builds Editor with Vs2022. Goes into preferences and selects VS2022 as compiler. User closes Editor.
2) User builds Editor with Clang. Opens Editor goes into preferences. The previously selected VS2022 compiler is now no longer compatible with the Editor. This is why there will now be a entry in the dropdown saying "SDK incompatible previously configured compiler". The user can either choose to just keep that value. For example if the user only wanted to quickly test something with the clang Editor without needing C++ Plugins. This way the compiler preferences don't change. Or the user can switch from the incompatible compiler to a Clang version that was auto detected.
3) If the user didn't change the compiler in preferences and opens the Editor build with VS2022 again, the dropdown will again read VS2022, because now the compiler is compatible with the Editor again.

The same can also happen on linux with clang / gcc. Although most likely most users will never see this option because they will always build with the same compiler.

Code/Editor/EditorFramework/CodeGen/CppProject.h
```diff
@@ -2,10 +2,92 @@
...
+  };
+};
+
+EZ_DECLARE_REFLECTABLE_TYPE(EZ_EDITORFRAMEWORK_DLL, ezCompiler);
+
+struct EZ_EDITORFRAMEWORK_DLL ezCompilerPreferences
+{
+  ezEnum<ezCompiler> m_eCompiler;
+  bool m_bCustomCompiler;
+  ezString m_CppCompiler;
```
Sanakan8472: Shouldn't the strings be prepended with s?

PR #1150, opened 2024-01-09
Fixed #1146: Added physics collision geometry visualization

### LINKED ISSUE #1146: [FEATURE] Runtime Jolt collider visualization
Add a way to visualize all physics collider geometry.

jankrassnigg:
Adds these CVars:
1. Jolt.Visualize.Geometry
2. Jolt.Visualize.Exclusive
3. Jolt.Visualize.Distance

Variable 1. enables visualization of physics collider geometry:

image

This is just rendered together with the regular geometry, which helps to figure out whether the physics geometry fits the render geometry. For example here the tree collision mesh is not imported the same way as the graphics mesh:

image

This gives a lot of z-fighting, if the meshes are identical and can be distracting. So variable 2 allows to disable rendering of regular geometry:

image

Different types use different colors. In the image above:
* light blue = static geometry
* dark blue = kinematic
* yellow = dynamic bodies
* green = query shapes (hitboxes)
* pink = ragdolls + ropes
* red = soft bodies
* transparent = triggers

Geometry is only updated in a certain radius around the camera (to reduce the performance impact). Variable 3 configures this distance. Overall the rendering is very efficient, though, since it simply uses the CustomMeshComponent to add meshes to the scene and thus the typical rendering and culling is done on them, this is not using the ezDebugRenderer.

### DIFF
--- a/Code/EnginePlugins/JoltPlugin/System/JoltWorldModule.cpp
+++ b/Code/EnginePlugins/JoltPlugin/System/JoltWorldModule.cpp
@@ -774,89 +774,70 @@ void ezJoltWorldModule::UpdateConstraints()
 
 ezAtomicInteger32 s_iColMeshVisGeoCounter;
 
-ezView* ezJoltWorldModule::GetMainView() const
-{
-  for (ezViewHandle hView : ezRenderWorld::GetMainViews())
-  {
-    ezView* pView;
-    if (!ezRenderWorld::TryGetView(hView, pView))
-      continue;
-
-    if (pView->GetWorld() != GetWorld())
-      continue;
-
-    if (pView->GetCameraUsageHint() == ezCameraUsageHint::EditorView ||
-        pView->GetCameraUsageHint() == ezCameraUsageHint::MainView)
-    {
-      return pView;
-    }
-  }
-
-  return nullptr;
-}
-
 struct DebugVis
 {
   const char* m_szMaterial = nullptr;
   ezColor m_Color;
 };
 
-// "{ e6367876-ddb5-4149-ba80-180af553d463 }" = Data/Base/Materials/Common/PhysicsColliders.ezMaterialAsset
+const char* szMatSolid = "{ e6367876-ddb5-4149-ba80-180af553d463 }";       // Data/Base/Materials/Common/PhysicsColliders.ezMaterialAsset
+const char* szMatTransparent = "{ ca43dda3-a28c-41fe-ae20-419182e56f87 }"; // Data/Base/Materials/Common/PhysicsCollidersTransparent.ezMaterialAsset
+const char* szMatTwoSided = "{ b03df0e4-98b7-49ba-8413-d981014a77be }";    // Data/Base/Materials/Common/PhysicsCollidersSoft.ezMaterialAsset
 
 static const DebugVis s_Vis[ezPhysicsShapeType::Count][2] =
   {
     // Static
     {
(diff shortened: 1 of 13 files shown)


Data/Tools/ezEditor/VisualShader/Textures.ddl
```diff
@@ -342,7 +342,14 @@ SamplerState $prop0_AutoSampler;
     string %Type { "float" }
     string %DefaultValue { "1" }
   }
-  
+
```
jankrassnigg, line 345: @C-Core I wanted to make a material that uses the pattern 3-way mapping in object space. So that the pattern doesn't change when moving it. I failed quite miserably, and think this is currently not possible, but don't understand why. The necessary Visual Shader nodes exist, but don't seem to work properly. I kept this change in, because I think it is generally correct, but it would be helpful, if you could have a look some time, whether this is really not possible.

Code/EnginePlugins/JoltPlugin/System/JoltWorldModule.cpp
```diff
@@ -762,5 +772,292 @@ void ezJoltWorldModule::UpdateConstraints()
   m_RequireUpdate.Clear();
 }
 
+ezAtomicInteger32 s_iColMeshVisGeoCounter;
+
+ezView* ezJoltWorldModule::GetMainView() const
```
C-Core: ezRenderWorld::GetViewByUsageHint does exactly what you have done manually here

Code/EnginePlugins/JoltPlugin/System/JoltWorldModule.cpp
```diff
@@ -762,5 +772,292 @@ void ezJoltWorldModule::UpdateConstraints()
...
+
+    // Dynamic
+    {
+      {"{ e6367876-ddb5-4149-ba80-180af553d463 }", ezColor::Gold},      // non-kinematic
+      {"{ e6367876-ddb5-4149-ba80-180af553d463 }", ezColor::DodgerBlue} // kinematic
+    },
+
+    // Query
+    {
+      {"{ ca43dda3-a28c-41fe-ae20-419182e56f87 }", ezColor::GreenYellow.WithAlpha(0.5f)}, // non-kinematic
```
C-Core: Would be nice to know what this material is as you have documented it for the static/dynamic material already. Same for cloth below.

Code/EnginePlugins/JoltPlugin/System/JoltWorldModule.cpp
```diff
@@ -762,5 +772,292 @@ void ezJoltWorldModule::UpdateConstraints()
...
+
+  for (const JPH::TransformedShape& ts : collector.mHits)
+  {
+    DebugBodyShapeKey key;
+    key.m_uiBodyID = ts.mBodyID.GetIndexAndSequenceNumber();
+    key.m_pShapePtr = ts.mShape.GetPtr();
+
+    bool bExisted = false;
+    auto& geo = m_DebugDrawComponents.FindOrAdd(key, &bExisted).Value();
+    geo.m_uiLastSeenCounter = m_uiDebugGeoLastSeenCounter;
```
C-Core, line 937: Where is this last seen counter used? Shouldn't there be some code that deletes old entries after some time?

C-Core (approved):

Code/EnginePlugins/JoltPlugin/System/JoltWorldModule.cpp
jankrassnigg: At the end of DebugDrawGeometry() (see function above) the shapes are removed.

PR #1152, opened 2024-01-10
Make it possible to manually specify a precompiled tools

Ingrater:
Given how complicated it is to build Linux executables which will work on all Linux distributions we can't realistically ship precompiled binary tools for Linux like we do on Windows.

This ads a value to the editor preferences allowing the user to precompile the binaries and then specify where the precompiled binaries are located. This is for example useful when running a debug build Editor on Linux but still wanting fast asset transforms with the dev / shipping tools. The current behavior on windows doesn't change, by default the Editor will use the checked in precompiled tools.

### DIFF
--- a/Code/Editor/EditorFramework/EditorApp/ExternalTools.cpp
+++ b/Code/Editor/EditorFramework/EditorApp/ExternalTools.cpp
@@ -43,8 +43,7 @@ ezString ezQtEditorApp::FindToolApplication(const char* szToolName)
     sTool = folder;
     sTool.AppendPath(szToolName);
 
-    ezFileStats toolStats;
-    if (ezFileSystem::GetFileStats(sTool, toolStats).Succeeded() && !toolStats.m_bIsDirectory)
+    if (ezOSFile::ExistsFile(sTool))
       return sTool;
   }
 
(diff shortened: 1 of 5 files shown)


bss-jan-krassnigg (approved):

jankrassnigg (approved):

jankrassnigg (approved):

Code/Editor/EditorFramework/EditorApp/ExternalTools.cpp
```diff
@@ -18,24 +18,34 @@ ezString ezQtEditorApp::FindToolApplication(const char* szToolName)
...
 
   if (pPref->m_bUsePrecompiledTools)
   {
-    ezMath::Swap(bFolders[0], bFolders[1]);
+    if(!pPref->m_sCustomPrecompiledToolsFolder.IsEmpty() && ezOSFile::ExistsDirectory(pPref->m_sCustomPrecompiledToolsFolder))
+    {
+      ezStringBuilder customToolsFolder = pPref->m_sCustomPrecompiledToolsFolder;
+      customToolsFolder.MakeCleanPath();
+      sFolders.PushBack(customToolsFolder);
+      ezMath::Swap(sFolders[0], sFolders[2]);
```
Sanakan8472, line 32: Wouldn't it make more sense if the order was custom, true, false in this case?

Ingrater: This is exactly what the implementation currently does.

PR #1154, opened 2024-01-11
Fixed #1147: Added Custom Data asset type for configuring game-specific C++ structs

### LINKED ISSUE #1147: [FEATURE] Add a "Configuration Asset" similar to Unity's ScriptableObject
This should make it easy and typesafe to setup a C++ configuration struct for sharing state.

x0r04rg:
Allows users to define their own custom asset types in runtime plugin. Similar to ScriptableObject in Unity or UDataAsset in Unreal. (#1147)


x0r04rg:
"Create Documents" test is failing here, but it's kind of intended behavior - "empty" custom data fails to transform, because, well, it's empty.

jankrassnigg (approved):

PR #1158, opened 2024-01-14
Linux background asset processing fixes

Ingrater:
* Add a state machine to ezProcessTask which takes into account that a connection to the EditorProcessor might not be open instantly. The current code made this assumption because it is true for windows IPC channels. But this is not the case for network connections or Linux IPC channels.
* Implement tracelogging on linux. Off by default. Need to set `EZ_3RDPARTY_TRACELOGGING_LTTNG_SUPPORT` to enable it.
* Various tracelogging fixes, make sure all relevant log messages end up in tracelogging.
* Add the Microsoft tracelogging lttng compatibility library as a third party library
* Only create copy of engine plugins for code reloading on platforms which actually require it. On Linux no copy is required and hot reloading still works. Doing a copy on Linux results in a race in the file system which might end up with a corrupted plugin file on disk during asset background processing.

### DIFF
--- a/Code/Editor/EditorEngineProcessFramework/IPC/ProcessCommunicationChannel.h
+++ b/Code/Editor/EditorEngineProcessFramework/IPC/ProcessCommunicationChannel.h
@@ -23,7 +23,7 @@ public:
   using WaitForMessageCallback = ezDelegate<bool(ezProcessMessage*)>;
   ezResult WaitForMessage(const ezRTTI* pMessageType, ezTime timeout, WaitForMessageCallback* pMessageCallack = nullptr);
   ezResult WaitForConnection(ezTime timeout);
-  bool IsConnected();
+  bool IsConnected() const;
 
   /// \brief Returns true if any message was processed
   bool ProcessMessages();
--- a/Code/Editor/EditorEngineProcessFramework/IPC/ProcessCommunicationChannel.h
+++ b/Code/Editor/EditorEngineProcessFramework/IPC/ProcessCommunicationChannel.h
@@ -23,6 +23,7 @@ public:
   using WaitForMessageCallback = ezDelegate<bool(ezProcessMessage*)>;
   ezResult WaitForMessage(const ezRTTI* pMessageType, ezTime timeout, WaitForMessageCallback* pMessageCallack = nullptr);
   ezResult WaitForConnection(ezTime timeout);
+  bool IsConnected();
 
   /// \brief Returns true if any message was processed
   bool ProcessMessages();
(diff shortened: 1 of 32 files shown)


Code/Editor/EditorEngineProcessFramework/IPC/ProcessCommunicationChannel.h
```diff
@@ -23,6 +23,7 @@ class EZ_EDITORENGINEPROCESSFRAMEWORK_DLL ezProcessCommunicationChannel
   using WaitForMessageCallback = ezDelegate<bool(ezProcessMessage*)>;
   ezResult WaitForMessage(const ezRTTI* pMessageType, ezTime timeout, WaitForMessageCallback* pMessageCallack = nullptr);
   ezResult WaitForConnection(ezTime timeout);
+  bool IsConnected();
```
jankrassnigg: const ?

Code/Engine/Foundation/Platform/Linux/MessageLoop_Linux.cpp
```diff
@@ -57,11 +57,11 @@ bool ezMessageLoop_linux::WaitForMessages(ezInt32 iTimeout, ezIpcChannel* pFilte
       ezUInt8 wakeupByte;
       auto readResult = read(m_wakeupPipeReadEndFd, &wakeupByte, sizeof(wakeupByte));
       m_pollInfos[0].revents = 0;
+      //ezLog::Debug("[IPC]MessageLoop_Linux::WakupProcessed");
```
jankrassnigg: remove

Code/Editor/EditorFramework/Assets/Implementation/AssetProcessor.cpp
```diff
@@ -384,114 +367,162 @@ bool ezProcessTask::GetNextAssetToProcess(ezUuid& out_guid, ezDataDirPath& out_p
...
+          {
+            m_AssetGuid = ezUuid();
+            m_AssetPath.Clear();
+            return bStartNewWork; // call again if we should be looking for new work
+          }
 
-    ezStringBuilder sTemp;
-    ezAssetCurator::GetSingleton()->GenerateTransitiveHull(ezConversionUtils::ToString(m_AssetGuid, sTemp), dependencies, true, true);
+          ezAssetInfo::TransformState state = ezAssetCurator::GetSingleton()->IsAssetUpToDate(m_AssetGuid, nullptr, nullptr, m_uiAssetHash, m_uiThumbHash);
+          EZ_ASSERT_DEV(state == ezAssetInfo::TransformState::NeedsTransform || state == ezAssetInfo::TransformState::NeedsThumbnail, "An asset was selected that is already up to date.");
```
jankrassnigg, line 415: not sure whether I would assert on those states, there are so many things that can interfere and produce race conditions ...

jankrassnigg (approved):

Sanakan8472 (approved):

Code/Editor/EditorFramework/Assets/Implementation/AssetProcessor.cpp
Ingrater: This is not new code, that was always there. Also all ezProcessTask instaces run in the same thread sequentially, so at least the ezProcessTasks can't interfere with each other.

PR #1169, opened 2024-01-19
Node Graph Improvements

PlasmaDev5:
Backport of a few style features for nodes graphs.

Background grid and Subway nodes
image

Node graph debug animation. Not currently used in any meaningful way but there ready for when debugging becomes a thing
NodeDebug

### DIFF
--- a/Code/Tools/Libs/GuiFoundation/NodeEditor/Implementation/Connection.cpp
+++ b/Code/Tools/Libs/GuiFoundation/NodeEditor/Implementation/Connection.cpp
@@ -91,6 +91,24 @@ void ezQtConnection::UpdateGeometry()
     p.moveTo(startPoint);
     p.lineTo(endPoint);
   }
+  else if (pScene->GetConnectionStyle() == ezQtNodeScene::ConnectionStyle::SubwayLines)
+  {
+    const float fPaddingFromNode = 20.0f;
+
+    QPointF startPoint = m_OutPoint;
+
+    QPointF offsetStartPoint = QPointF(startPoint.x() + fPaddingFromNode, startPoint.y());
+
+    QPointF endPoint = m_InPoint;
+    QPointF offsetEndPoint = QPointF(endPoint.x() - fPaddingFromNode, endPoint.y());
+
+    p.moveTo(startPoint);
+    p.lineTo(offsetStartPoint);
+    DrawSubwayPath(p, offsetStartPoint, offsetEndPoint);
+    p.lineTo(offsetEndPoint);
+    p.moveTo(offsetEndPoint);
+    p.lineTo(endPoint);
+  }
   else
   {
     p.moveTo(m_OutPoint);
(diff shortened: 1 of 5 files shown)


Code/Tools/Libs/GuiFoundation/NodeEditor/Implementation/Connection.cpp
```diff
@@ -168,4 +186,102 @@ void ezQtConnection::paint(QPainter* painter, const QStyleOptionGraphicsItem* op
   }
 
   painter->drawPath(path());
+
+  if (decorationFlags.IsSet(ezQtNodeScene::ConnectionDecorationFlags::DrawDebugging))
```
C-Core, line 190: There also needs to be something on the connection it self to tell whether it should show the debugging effect. Can be added later though.

Code/Tools/Libs/GuiFoundation/NodeEditor/Implementation/Connection.cpp
```diff
@@ -168,4 +186,102 @@ void ezQtConnection::paint(QPainter* painter, const QStyleOptionGraphicsItem* op
...
+    for (qreal length = 0; length < segments + 0.0005f; ++length)
+    {
+      painter->drawEllipse(path().pointAtPercent(path().percentAtLength((length + offset) * 16)), 2, 2);
+    }
+  }
+}
+
+void ezQtConnection::DrawSubwayPath(QPainterPath& path, const QPointF& startPoint, const QPointF& endPoint)
+{
+  const bool isStartAboveTarget = startPoint.y() <= endPoint.y();
```
C-Core, line 204: This looks quite involved. Can you explain what a "subway path" does?

Code/Tools/Libs/GuiFoundation/NodeEditor/NodeScene.moc.h
```diff
@@ -43,8 +43,9 @@ class EZ_GUIFOUNDATION_DLL ezQtNodeScene : public QGraphicsScene
     {
       BezierCurve,
       StraightLine,
+      SubwayLines,
 
-      Default = BezierCurve
+      Default = SubwayLines
```
C-Core: I don't think we should change the default. There should be a view option or something like that where the user can toggle between bezier and subway lines.

Code/Tools/Libs/GuiFoundation/NodeEditor/Implementation/Connection.cpp
PlasmaDev5: Yea thats the intention at the moment. I haven't added any debug logic myself yet but it made sense to include the visuals in the changes i was making

Code/Tools/Libs/GuiFoundation/NodeEditor/Implementation/Connection.cpp
PlasmaDev5: Ill add a comment for it.
It draws the same way you would see on many subway maps. this means only straight and diagonal lines to reach the end goal. It also handles connections to nodes to the left of the output 

Subway

Code/Tools/Libs/GuiFoundation/NodeEditor/NodeScene.moc.h
PlasmaDev5: I feel like ideally it should be something in the editor settings along with path width and other things like that. But i havent looked into doing that kind of thing yet. The change there was more for testing and can be reverted but i feel it gives a much more polished look for now until there is a setting for it.

C-Core: I agree that it looks nice. The reason I'm hesitant to change the default is that you will position your nodes depending on how the connections behave to untangle the mess a bit. If we simply change the default existing graphs might look weird because they have been layouted with a different connection behavior. I will play around with these changes over the weekend and test them on a few more complex graphs.

jankrassnigg:
Very cool! Thanks a lot! I'll leave the review up to @C-Core though.

Code/Tools/Libs/GuiFoundation/NodeEditor/Implementation/Connection.cpp
```diff
@@ -91,6 +91,25 @@ void ezQtConnection::UpdateGeometry()
...
+    QPointF offsetStartPoint = QPointF(startPoint.x() + fPaddingFromNode, startPoint.y());
+
+    QPointF endPoint = m_InPoint;
+    QPointF offsetEndPoint = QPointF(endPoint.x() - fPaddingFromNode, endPoint.y());
+
+    p.moveTo(startPoint);
+    p.lineTo(offsetStartPoint);
+    DrawSubwayPath(p, offsetStartPoint, offsetEndPoint);
+    p.lineTo(offsetEndPoint);
+    p.moveTo(offsetEndPoint);
```
C-Core: This additional moveTo is unnecessary and causes artifacts since the lines are not properly connected at this point.

Code/Tools/Libs/GuiFoundation/NodeEditor/Implementation/NodeView.cpp
```diff
@@ -136,9 +136,56 @@ void ezQtNodeView::resizeEvent(QResizeEvent* event)
   UpdateView();
 }
 
+void ezQtNodeView::drawBackground(QPainter* painter, const QRectF& r)
```
C-Core, line 139: While it looks nice this unfortunately costs way too much performance and makes the whole node view react very poorly while panning especially when zoomed out a bit. You can leave the code in there but put an if(false) or something like that around it so we can add an editor setting for it in the future.

Code/Tools/Libs/GuiFoundation/NodeEditor/NodeScene.moc.h
C-Core: I gave it a shot with a couple of complex graphs. It is actually not too bad, but I would still not change the default as it can be confusing in some existing graphs. I can look into adding the corresponding editor setting after this PR has been merged.

Code/Tools/Libs/GuiFoundation/NodeEditor/Implementation/NodeView.cpp
PlasmaDev5: Potentially another option in the graph settings as background grids are fairly standard these days for node tools

Code/Tools/Libs/GuiFoundation/NodeEditor/Implementation/Connection.cpp
PlasmaDev5: Fixed

C-Core (approved):

Code/Tools/Libs/GuiFoundation/NodeEditor/Implementation/NodeView.cpp
PlasmaDev5: Do we have knowledge of the zoom level. we could adapt the graph drawing based on zoom lower draw cost when navigating large graphs

PlasmaDev5: @C-Core been playing around with this a little. Not the cleanest solution but improves the performance a tone. In future i ideally id want something that has multiple resolutions scales, so at each 0.25 step the graph looks the same but scaled for that resolution if that makes sense

```
  if(m_ViewScale.manhattanLength() > 1.0)
  {
    QPen pfine(plToQtColor(plColorScheme::GetColor(plColorScheme::Black, 9)), 1.0);

    painter->setPen(pfine);
    DrawGrid(painter, 15);
  }

  if(m_ViewScale.manhattanLength() > 0.1)
  {
    double scale = m_ViewScale.manhattanLength() < 0.25 ? 150.0 : 300.0;

    QPen p(plToQtColor(plColorScheme::GetColor(plColorScheme::Gray, 1)), 1.0);

    painter->setPen(p);
    DrawGrid(painter, scale);
  }
  ```

PR #1171, opened 2024-01-19
Converted ezMemoryUtils to use "if constexpr"

jankrassnigg:
There is probably lots more code that could be simplified now, but this has already been tricky enough for me.

### DIFF
--- a/Code/Engine/Foundation/Memory/MemoryUtils.h
+++ b/Code/Engine/Foundation/Memory/MemoryUtils.h
@@ -196,11 +196,6 @@ private:
   static void CopyOrMoveConstruct(Destination* pDestination, Source&& source, IsRValueReference);
 
   template <typename T>
-  static void Copy(T* pDestination, const T* pSource, size_t uiCount, ezTypeIsPod);
-  template <typename T>
-  static void Copy(T* pDestination, const T* pSource, size_t uiCount, ezTypeIsClass);
-
-  template <typename T>
   static void CopyOverlapped(T* pDestination, const T* pSource, size_t uiCount, ezTypeIsPod);
   template <typename T>
   static void CopyOverlapped(T* pDestination, const T* pSource, size_t uiCount, ezTypeIsClass);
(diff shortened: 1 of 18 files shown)


Code/Engine/Foundation/Memory/MemoryUtils.h
```diff
@@ -27,7 +27,7 @@ class ezMemoryUtils
   ///
   /// You should use 'DefaultConstruct' instead if default construction is needed for trivial types as well.
   template <typename T>
-  static void Construct(T* pDestination, size_t uiCount = 1); // [tested]
+  static void DefaultConstructNonTrivial(T* pDestination, size_t uiCount = 1); // [tested]
```
Ingrater: The name of this function is misleading now. What it does is: It detects if `T` is a trivial type or a non-trivial type and acts accordingly. `DefaultConstruct` on the other hand does not look at `T` at all, and just always default constructs.

I would keep `Construct` for this one and call the other one `ForceDefaultConstruct`.

Ingrater (changes requested):

Code/Engine/Foundation/Memory/MemoryUtils.h
jankrassnigg: I'm not against renaming either or both functions, but I'm not happy with the suggestion. The default construct function does exactly what it is named for. It default constructs the given type. So I think that is a good name. It is the other one that has unexpected behavior. Construct does not convey that there is something different going on for different types. In open for a better name, but I think it has to be more obvious that people need to read the details.

jankrassnigg: Another option would be to turn this into one template function that takes options, then it could be called construct, and we make it a parameter whether it should you trivial types. Thoughts?

jankrassnigg: See ezConstructionMode commit. Not sure whether this is that much better, but at least it is now hard to miss that there are different ways of constructing objects.

Ingrater (approved):

PR #1177, opened 2024-01-22
More Node Graph Improvements

PlasmaDev5:
Improve performance on grid background so should be capable of being enabled by default.
Also added a create on connection drag feature shown in the gif. If you drop into an empty space it will prompt you with the create node popup. If you create a node it will try connect to the first valid input

Create on drag


jankrassnigg:
Sweet! I always wanted to have that create-on-drag feature. Does it somehow limit the content of the context menu to items that actually CAN be connected? (I'm not up to date on whether we have any filtering functionality for the menu).

PlasmaDev5:
Not currently. Unless im missing something i dont think there is a filter by input type and i dont imagine it being trivial to implement. Even then i imagine at some point there may be type casting nodes that could be use to connect unsupported types 

For now i just use OpenSearchMenu

C-Core (approved):
Performance of the grid rendering seems to be ok now.

PR #1179, opened 2024-01-23
Minor Improvements to Light Components

PlasmaDev5:
This consists of 2 major changes and a few supporting improvements.

First and smallest change is the addition of a specular multiplier control. This allows people to have diffuse only lights. Can be very useful when adding fake light to an area such as boosting the light through a window. 

Second and most important feature is the addition of kelvin temperature for lights.
This includes a new ezSliderAttribute that allows for single item sliders of any min and max. In future it would be good to skin the slider with an image in this case a gradient of the temperature.
image

Property

### DIFF
--- a/Code/Engine/Foundation/Math/Color.h
+++ b/Code/Engine/Foundation/Math/Color.h
@@ -247,6 +247,9 @@ public:
   void SetRGBA(float fLinearRed, float fLinearGreen, float fLinearBlue, float fLinearAlpha = 1.0f); // [tested]
 
   /// \brief Returns a color created from the kelvin temperature. https://wikipedia.org/wiki/Color_temperature
+  /// Originally inspired from https://tannerhelland.com/2012/09/18/convert-temperature-rgb-algorithm-code.html
+  /// But with heavy modification to better fit the mapping shown out in https://seblagarde.files.wordpress.com/2015/07/course_notes_moving_frostbite_to_pbr_v32.pdf
+  /// Physically accurate clipping points are 6580K for Red and 6560K for G and B. but approximated to 6570k for all to give a better mapping.
   static ezColor MakeFromKelvin(ezUInt32 uiKelvin);
   // *** Conversion Operators/Functions ***
 public:
--- a/Code/Engine/Foundation/Math/Color.h
+++ b/Code/Engine/Foundation/Math/Color.h
@@ -246,6 +246,7 @@ public:
   /// \brief Sets all four RGBA components.
   void SetRGBA(float fLinearRed, float fLinearGreen, float fLinearBlue, float fLinearAlpha = 1.0f); // [tested]
 
+  /// \brief Returns a color created from the kelvin temperature. https://wikipedia.org/wiki/Color_temperature
   static ezColor MakeFromKelvin(ezUInt32 uiKelvin);
   // *** Conversion Operators/Functions ***
 public:
--- a/Code/Engine/Foundation/Math/Color.h
+++ b/Code/Engine/Foundation/Math/Color.h
@@ -246,7 +246,7 @@ public:
   /// \brief Sets all four RGBA components.
   void SetRGBA(float fLinearRed, float fLinearGreen, float fLinearBlue, float fLinearAlpha = 1.0f); // [tested]
 
-  void SetKelvin(ezUInt32 uKelvin);
+  static ezColor MakeFromKelvin(ezUInt32 uiKelvin);
   // *** Conversion Operators/Functions ***
 public:
   /// \brief Sets this color from a HSV (hue, saturation, value) format.
--- a/Code/Engine/Foundation/Math/Color.h
+++ b/Code/Engine/Foundation/Math/Color.h
@@ -246,6 +246,7 @@ public:
   /// \brief Sets all four RGBA components.
   void SetRGBA(float fLinearRed, float fLinearGreen, float fLinearBlue, float fLinearAlpha = 1.0f); // [tested]
 
+  void SetKelvin(ezUInt32 uKelvin);
   // *** Conversion Operators/Functions ***
 public:
   /// \brief Sets this color from a HSV (hue, saturation, value) format.
--- a/Code/Engine/Foundation/Math/Implementation/Color_inl.h
+++ b/Code/Engine/Foundation/Math/Implementation/Color_inl.h
@@ -46,17 +46,20 @@ inline void ezColor::SetRGBA(float fLinearRed, float fLinearGreen, float fLinear
   a = fLinearAlpha;
 }
 
-inline void ezColor::SetKelvin(ezUInt32 uKelvin)
+inline ezColor ezColor::MakeFromKelvin(ezUInt32 uiKelvin)
 {
-  float kelvin = ezMath::Clamp(uKelvin, 1000u, 40000u) / 1000.0f;
+  ezColor finalColor;
+  float kelvin = ezMath::Clamp(uiKelvin, 1000u, 40000u) / 1000.0f;
   float kelvin2 = kelvin * kelvin;
 
   // Red
-  r = kelvin < 6.580f ? 1.0f : ezMath::Clamp((1.35651f + 0.216422f * kelvin + 0.000633715f * kelvin2) / (-3.24223f + 0.918711f * kelvin), 0.0f, 1.0f);
+  finalColor.r = kelvin < 6.570f ? 1.0f : ezMath::Clamp((1.35651f + 0.216422f * kelvin + 0.000633715f * kelvin2) / (-3.24223f + 0.918711f * kelvin), 0.0f, 1.0f);
   // Green
-  g = kelvin < 6.560f ? ezMath::Clamp((-399.809f + 414.271f * kelvin + 111.543f * kelvin2) / (2779.24f + 164.143f * kelvin + 84.7356f * kelvin2), 0.0f, 1.0f) : ezMath::Clamp((1370.38f + 734.616f * kelvin + 0.689955f * kelvin2) / (-4625.69f + 1699.87f * kelvin), 0.0f, 1.0f);
+  finalColor.g = kelvin < 6.570f ? ezMath::Clamp((-399.809f + 414.271f * kelvin + 111.543f * kelvin2) / (2779.24f + 164.143f * kelvin + 84.7356f * kelvin2), 0.0f, 1.0f) : ezMath::Clamp((1370.38f + 734.616f * kelvin + 0.689955f * kelvin2) / (-4625.69f + 1699.87f * kelvin), 0.0f, 1.0f);
   // Blue
-  b = kelvin > 6.580f ? 1.0f : ezMath::Clamp((348.963f - 523.53f * kelvin + 183.62f * kelvin2) / (2848.82f - 214.52f * kelvin + 78.8614f * kelvin2), 0.0f, 1.0f);
+  finalColor.b = kelvin > 6.570f ? 1.0f : ezMath::Clamp((348.963f - 523.53f * kelvin + 183.62f * kelvin2) / (2848.82f - 214.52f * kelvin + 78.8614f * kelvin2), 0.0f, 1.0f);
+
+  return finalColor;
 }
 
 // http://en.wikipedia.org/wiki/Luminance_%28relative%29
--- a/Code/Engine/Foundation/Math/Implementation/Color_inl.h
+++ b/Code/Engine/Foundation/Math/Implementation/Color_inl.h
@@ -1,3 +1,4 @@
+#include "Color.h"
 #pragma once
 
 inline ezColor::ezColor()
@@ -43,6 +44,19 @@ inline void ezColor::SetRGBA(float fLinearRed, float fLinearGreen, float fLinear
   g = fLinearGreen;
   b = fLinearBlue;
   a = fLinearAlpha;
+}
+
+inline void ezColor::SetKelvin(ezUInt32 uKelvin)
+{
+  float kelvin = ezMath::Clamp(uKelvin, 1000u, 40000u) / 1000.0f;
+  float kelvin2 = kelvin * kelvin;
+
+  // Red
+  r = kelvin < 6.580f ? 1.0f : ezMath::Clamp((1.35651f + 0.216422f * kelvin + 0.000633715f * kelvin2) / (-3.24223f + 0.918711f * kelvin), 0.0f, 1.0f);
+  // Green
+  g = kelvin < 6.560f ? ezMath::Clamp((-399.809f + 414.271f * kelvin + 111.543f * kelvin2) / (2779.24f + 164.143f * kelvin + 84.7356f * kelvin2), 0.0f, 1.0f) : ezMath::Clamp((1370.38f + 734.616f * kelvin + 0.689955f * kelvin2) / (-4625.69f + 1699.87f * kelvin), 0.0f, 1.0f);
+  // Blue
+  b = kelvin > 6.580f ? 1.0f : ezMath::Clamp((348.963f - 523.53f * kelvin + 183.62f * kelvin2) / (2848.82f - 214.52f * kelvin + 78.8614f * kelvin2), 0.0f, 1.0f);
 }
 
 // http://en.wikipedia.org/wiki/Luminance_%28relative%29
--- a/Code/Engine/Foundation/Math/Implementation/Color_inl.h
+++ b/Code/Engine/Foundation/Math/Implementation/Color_inl.h
@@ -1,4 +1,3 @@
-#include "Color.h"
 #pragma once
 
 inline ezColor::ezColor()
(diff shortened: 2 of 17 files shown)


PlasmaDev5:
Was unsure how to resolve the icon coloring for this situation

Code/Engine/Foundation/Math/Color.h
```diff
@@ -246,6 +246,7 @@ class EZ_FOUNDATION_DLL ezColor
   /// \brief Sets all four RGBA components.
   void SetRGBA(float fLinearRed, float fLinearGreen, float fLinearBlue, float fLinearAlpha = 1.0f); // [tested]
 
+  void SetKelvin(ezUInt32 uKelvin);
```
jankrassnigg: I think this should be a static make function:
static ezColor MakeFromKelvin(ezUInt32 uiKelvin);

Code/Engine/Foundation/Math/Implementation/Color_inl.h
```diff
@@ -45,6 +46,19 @@ inline void ezColor::SetRGBA(float fLinearRed, float fLinearGreen, float fLinear
   a = fLinearAlpha;
 }
 
+inline void ezColor::SetKelvin(ezUInt32 uKelvin)
```
jankrassnigg: uiKelvin, not uKelvin, otherwise our clang tidy build will complain

Code/Engine/RendererCore/Lights/Implementation/LightComponent.cpp
```diff
@@ -60,18 +75,56 @@ ezColorGammaUB ezLightComponent::GetLightColor() const
   return m_LightColor;
 }
 
+ezColorGammaUB ezLightComponent::GetFinalLightColor() const
```
jankrassnigg: If you do this inside GetLightColor, you should get the right color on the viewport icon.
Then you shouldn't need this function, at all.
If you need a function to get the 'raw' light color value, I would instead add a getter for that ("GetNonPhysicalLightColor" or something like that). But I don't really think you need that.

Code/Engine/RendererCore/Lights/Implementation/LightComponent.cpp
```diff
@@ -60,18 +75,56 @@ ezColorGammaUB ezLightComponent::GetLightColor() const
...
+}
+
 void ezLightComponent::SetIntensity(float fIntensity)
 {
   m_fIntensity = ezMath::Max(fIntensity, 0.0f);
 
   TriggerLocalBoundsUpdate();
 }
 
+void ezLightComponent::SetTemperature(ezUInt32 uTemperature)
```
jankrassnigg: uiTemperature

Code/Engine/RendererCore/Lights/Implementation/LightComponent.cpp
```diff
@@ -60,18 +75,56 @@ ezColorGammaUB ezLightComponent::GetLightColor() const
...
+}
+
 float ezLightComponent::GetIntensity() const
 {
   return m_fIntensity;
 }
 
+void ezLightComponent::SetSpecularMultiplier(float fSpecularMultiplier)
+{
+  m_fSpecularMultiplier = ezMath::Max(fSpecularMultiplier, 0.0f);
```
jankrassnigg, line 116: clamp 0 - 1 ?

Code/Engine/RendererCore/Lights/Implementation/PointLightComponent.cpp
```diff
@@ -113,8 +113,18 @@ void ezPointLightComponent::OnMsgExtractRenderData(ezMsgExtractRenderData& msg)
   auto pRenderData = ezCreateRenderDataForThisFrame<ezPointLightRenderData>(GetOwner());
 
   pRenderData->m_GlobalTransform = t;
-  pRenderData->m_LightColor = m_LightColor;
```
jankrassnigg, line 116: Can just call Get(Final)LightColor here

Code/Engine/RendererCore/Lights/Implementation/SpotLightComponent.cpp
```diff
@@ -158,8 +158,18 @@ void ezSpotLightComponent::OnMsgExtractRenderData(ezMsgExtractRenderData& msg) c
   auto pRenderData = ezCreateRenderDataForThisFrame<ezSpotLightRenderData>(GetOwner());
 
   pRenderData->m_GlobalTransform = t;
-  pRenderData->m_LightColor = m_LightColor;
```
jankrassnigg, line 161: again, call Get(Final)LightColor here

Code/Engine/RendererCore/Lights/LightComponent.h
```diff
@@ -37,13 +38,24 @@ class EZ_RENDERERCORE_DLL ezLightComponent : public ezRenderComponent
   ezLightComponent();
   ~ezLightComponent();
 
+  void SetUsingColorTemperature(bool bUseColorTemperature);
```
jankrassnigg, line 43: Can you add some brief comments to these functions, that explains roughly what this is for (for people unfamiliar with setting light colors in Kelvin).
Can also be a link to a website, like Wikipedia.

jankrassnigg (commented):
Already looking really good :)

jankrassnigg:
> Was unsure how to resolve the icon coloring for this situation

See my comment about moving the functionality in GetFinalLightColor into GetLightColor. I think that should make it work as desired.

Code/Engine/Foundation/Math/Color.h
jankrassnigg: Oh and with a nice descriptive comment, please :)

Code/Engine/RendererCore/Lights/Implementation/LightComponent.cpp
PlasmaDev5: This was intentionally left with no max so people can use it to boost specular beyond the max

Code/Engine/Foundation/Math/Color.h
PlasmaDev5: fixed

Code/Engine/RendererCore/Lights/Implementation/LightComponent.cpp
PlasmaDev5: fixed

Code/Engine/Foundation/Math/Implementation/Color_inl.h
PlasmaDev5: Fixed

Code/Engine/RendererCore/Lights/Implementation/LightComponent.cpp
PlasmaDev5: Fixed

Code/Engine/RendererCore/Lights/Implementation/PointLightComponent.cpp
PlasmaDev5: Fixed

Code/Engine/RendererCore/Lights/LightComponent.h
PlasmaDev5: Fixed

Code/Engine/RendererCore/Lights/Implementation/SpotLightComponent.cpp
PlasmaDev5: Fixed

Code/Engine/Foundation/Math/Implementation/Color_inl.h
```diff
@@ -1,3 +1,4 @@
+#include "Color.h"
```
C-Core: This include should not be here

Code/Engine/Foundation/Math/Implementation/Color_inl.h
```diff
@@ -45,6 +46,22 @@ inline void ezColor::SetRGBA(float fLinearRed, float fLinearGreen, float fLinear
   a = fLinearAlpha;
 }
 
+inline ezColor ezColor::MakeFromKelvin(ezUInt32 uiKelvin)
```
C-Core, line 48: Add a reference where this algorithm comes from

Code/Engine/Foundation/Reflection/Implementation/PropertyAttributes.cpp
```diff
@@ -144,6 +144,22 @@ EZ_BEGIN_DYNAMIC_REFLECTED_TYPE(ezDefaultValueAttribute, 1, ezRTTIDefaultAllocat
 }
 EZ_END_DYNAMIC_REFLECTED_TYPE;
 
+EZ_BEGIN_DYNAMIC_REFLECTED_TYPE(ezSliderAttribute, 1, ezRTTIDefaultAllocator<ezSliderAttribute>)
```
C-Core, line 147: We already have ezClampValueAttribute to define a min/max value, so this slider attribute does not need it as well.

Code/Engine/Foundation/Math/Implementation/Color_inl.h
jankrassnigg: remove the include, it's not needed and the Android and Linux builds don't like it anyway

Code/Engine/Foundation/Math/Color.h
```diff
@@ -246,6 +246,7 @@ class EZ_FOUNDATION_DLL ezColor
   /// \brief Sets all four RGBA components.
   void SetRGBA(float fLinearRed, float fLinearGreen, float fLinearBlue, float fLinearAlpha = 1.0f); // [tested]
 
+  static ezColor MakeFromKelvin(ezUInt32 uiKelvin);
```
jankrassnigg, line 253: Can you also add a comment to this function?

Code/Engine/RendererCore/Lights/LightComponent.h
```diff
@@ -72,8 +86,11 @@ class EZ_RENDERERCORE_DLL ezLightComponent : public ezRenderComponent
   static float CalculateScreenSpaceSize(const ezBoundingSphere& sphere, const ezCamera& camera);
 
 protected:
+  bool m_bUseColorTemperature = false;
```
C-Core: Move this bool member before or after m_bCastShadows to avoid unnecessary padding

Code/Tools/Libs/GuiFoundation/PropertyGrid/Implementation/PropertyWidget.cpp
```diff
@@ -552,7 +552,35 @@ void ezQtPropertyEditorIntSpinboxWidget::OnInit()
   auto pNoTemporaryTransactions = m_pProp->GetAttributeByType<ezNoTemporaryTransactionsAttribute>();
   m_bUseTemporaryTransaction = (pNoTemporaryTransactions == nullptr);
 
-  if (const ezClampValueAttribute* pClamp = m_pProp->GetAttributeByType<ezClampValueAttribute>())
+  if (const ezSliderAttribute* pSlider = m_pProp->GetAttributeByType <ezSliderAttribute>())
```
C-Core, line 555: There should be code already that creates a slider for int values, I think it only does it when the range is very small. We should re-use that code and force the slider code path when the sliderattribute is present. The min and max value should come from a ClampValueAttribute.

Data/Base/Shaders/Common/Lighting.h
```diff
@@ -533,6 +533,8 @@ AccumulatedLight CalculateLighting(ezMaterialData matData, ezPerClusterData clus
       #if defined(USE_MATERIAL_SUBSURFACE_COLOR)
         AccumulateLight(totalLight, SubsurfaceShading(matData, lightVector, viewVector), lightColor * (attenuation * subsurfaceShadow));        
       #endif
+
+      totalLight.specularLight *= lightData.specularMultiplier;
```
C-Core: This is just wrong. Here you multiply all accumulated specular light from all lights that have been evaluated so far by the multiplier of one light.

C-Core (changes requested):

Code/Engine/Foundation/Reflection/Implementation/PropertyAttributes.cpp
jankrassnigg: No I asked for this. The clamp attribute only does it for limited ranges automatically, it doesn't add a slider unconditionally, which is what we want here.
Also later I would like to add a completely new widget type for when this attribute exists, so that we can use dedicated images for things like the color temperature.

Code/Tools/Libs/GuiFoundation/PropertyGrid/Implementation/PropertyWidget.cpp
jankrassnigg: See my comment above, I'd prefer it like this for a dedicated widget.

Data/Base/Shaders/Common/Lighting.h
jankrassnigg: So would need to store the lightdata struct of the function that is passed into AccumulateLight and adjust the specular first, before passing it on to AccumulateLight ?

Code/Engine/Foundation/Math/Implementation/Color_inl.h
PlasmaDev5: Not sure where that came from. id guess auto includes. Fixed now.

PlasmaDev5: Added

Code/Engine/Foundation/Math/Color.h
PlasmaDev5: Added

Code/Engine/Foundation/Math/Implementation/Color_inl.h
PlasmaDev5: Resolved

Code/Engine/Foundation/Reflection/Implementation/PropertyAttributes.cpp
PlasmaDev5: This came out of a convocation with @jankrassnigg due to ezClampValueAttribute having a hard coded upper limit

Data/Base/Shaders/Common/Lighting.h
PlasmaDev5: From my limited understanding it only makes sense for DefaultShading. iv added a new AccumulateLight  that takes a specular multiplier.

Code/Engine/Foundation/Math/Implementation/Color_inl.h
PlasmaDev5: Added a comment

Code/Engine/RendererCore/Lights/LightComponent.h
PlasmaDev5: moved

PlasmaDev5:
Not sure why it reported issues from elsewhere in the shader. Could do with some advice on this one

Code/Engine/Foundation/Reflection/Implementation/PropertyAttributes.cpp
C-Core: Discussed offline with Jan we leave it like this for now, so nothing to do on your end.

Data/Base/Shaders/Common/Lighting.h
C-Core: Yep, it only makes sense for DefaultShading. AccumulateLight is a good place to do the multiplication so I'm fine with that solution.

C-Core (approved):
The shader compilation messages you see in the log are just warnings that are normal, you can ignore those. The real issue is an image comparison failure that looks like the specular contribution of the lights has changed which I don't understand since the multiplier is 1 by default so nothing should change. I can look into that.

Code/Engine/RendererCore/Lights/Implementation/LightComponent.cpp
```diff
@@ -19,12 +19,15 @@ void ezLightRenderData::FillBatchIdAndSortingKey(float fScreenSpaceSize)
...
 
 // clang-format off
-EZ_BEGIN_ABSTRACT_COMPONENT_TYPE(ezLightComponent, 4)
+EZ_BEGIN_ABSTRACT_COMPONENT_TYPE(ezLightComponent, 5)
 {
   EZ_BEGIN_PROPERTIES
   {
-    EZ_ACCESSOR_PROPERTY("LightColor", GetLightColor, SetLightColor),
+    EZ_ACCESSOR_PROPERTY("UseColorTemperature", GetUsingColorTemperature, SetUsingColorTemperature),
+    EZ_ACCESSOR_PROPERTY("LightColor", GetBaseLightColor, SetLightColor),
```
C-Core: Btw this should GetLightColor and not GetBaseLightColor, then it will also work with the icon coloring when using color temperature.

C-Core:
I've just pushed a fix for the failing particle tests (and also the right getter for the LightColor property) to your branch. Hope you don't mind.

PlasmaDev5:
I think the build machine broke. Its been going at least 5 hours at this point 😆

PR #1184, opened 2024-01-26
Sensor tag filter

x0r04rg:
* Add tag filters to SensorComponent
* SpatialSystem::QueryParams m_IncludeTags and m_ExcludeTags are now pointers to prevent unnecessary copying which could lead to memory allocations


C-Core (commented):
Looking good, thanks for the contribution!
The serialization of the include and exclude tags in the sensor component is missing though.

x0r04rg:
You're right, here it is

C-Core (approved):

PR #1185, opened 2024-01-26
Implement ez_enable_strict_warnings cmake function for clang

Ingrater:
The cmake function ez_enable_strict_warnings is currently a no-op when using the clang compiler. This commit implements ez_enable_strict_warnings for clang and fixes the resulting errors.

### DIFF
--- a/Code/Engine/Core/Scripting/ScriptCoroutine.h
+++ b/Code/Engine/Core/Scripting/ScriptCoroutine.h
@@ -104,7 +104,6 @@ private:
     static_cast<Derived*>(this)->Start(ezVariantAdapter<typename getArgument<I, Args...>::Type>(arguments[I])...);
   }
 
-protected:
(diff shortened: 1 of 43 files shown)


Code/Engine/Core/Scripting/ScriptCoroutine.h
```diff
@@ -104,7 +104,8 @@ class ezTypedScriptCoroutine : public ezScriptCoroutine
     static_cast<Derived*>(this)->Start(ezVariantAdapter<typename getArgument<I, Args...>::Type>(arguments[I])...);
   }
 
-  virtual void Start(ezArrayPtr<ezVariant> arguments) override
+protected:
```
C-Core: protected shouldn't be necessary after the rename

C-Core (approved):

PR #1186, opened 2024-01-27
Changed slider attribute to display an image to visualize the value range

jankrassnigg:
This is currently only used by the light component for the color temperature:

image

The image used is generated in code, so it looks exactly like the values that it produces.

### DIFF
--- a/Code/EditorPlugins/Scene/EditorPluginScene/EditorPluginScene.cpp
+++ b/Code/EditorPlugins/Scene/EditorPluginScene/EditorPluginScene.cpp
@@ -26,6 +26,7 @@
 #include <GuiFoundation/Action/DocumentActions.h>
 #include <GuiFoundation/Action/EditActions.h>
 #include <GuiFoundation/Action/StandardMenus.h>
+#include <GuiFoundation/PropertyGrid/Implementation/PropertyWidget.moc.h>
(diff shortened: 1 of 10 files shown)


C-Core (approved):

Code/EditorPlugins/Scene/EditorPluginScene/EditorPluginScene.cpp
```diff
@@ -89,9 +90,25 @@ void ezCameraComponent_PropertyMetaStateEventHandler(ezPropertyMetaStateEvent& e
 void ezSkyLightComponent_PropertyMetaStateEventHandler(ezPropertyMetaStateEvent& e);
 void ezGreyBoxComponent_PropertyMetaStateEventHandler(ezPropertyMetaStateEvent& e);
 void ezLightComponent_PropertyMetaStateEventHandler(ezPropertyMetaStateEvent& e);
-
 void ezSceneDocument_PropertyMetaStateEventHandler(ezPropertyMetaStateEvent& e);
 
+QImage SliderImageGenerator_LightTemperature(ezUInt32 uiWidth, ezUInt32 uiHeight)
```
C-Core: Maybe this function could also take the min and max value as input?

PR #1197, opened 2024-02-04
Added ezTypedComponentHandle

### LINKED ISSUE #1145: [FEATURE] Add support for typed component handles
See patch:

[TypedHandle.patch](https://github.com/ezEngine/ezEngine/files/13857362/TypedHandle.patch)

RiederAlex:
Added ezTypedComponentHandle and adapted the unit test to get some converage as well.

However, I'm unsure how to properly test serialization (mentioned in this [Discord discussion](https://discord.com/channels/806092968346779699/1192661940556812369)) as `ezWorldReader::ReadComponentHandle` and ` ezWorldWriter::WriteComponentHandle` exist (and they compile b/c of the implicit downcast from ezTypedComponentHandle to ezComponentHandle), but there is no obvious way to set this via the editor, as only GameObjects can be exposed as properties.

### DIFF
--- a/Code/Engine/Core/World/Declarations.h
+++ b/Code/Engine/Core/World/Declarations.h
@@ -143,6 +143,25 @@ struct ezComponentHandle
   friend class ezComponent;
 };
 
+template <typename TYPE>
+struct ezTypedComponentHandle : public ezComponentHandle
+{
+  ezTypedComponentHandle() = default;
+  explicit ezTypedComponentHandle(const ezComponentHandle& untyped)
+  {
+    m_InternalId = untyped.GetInternalID();
+  }
+
+  template <typename T, std::enable_if_t<std::is_convertible_v<T*, TYPE*>,bool> = true>
+  explicit ezTypedComponentHandle(const ezTypedComponentHandle<T>& other) : ezTypedComponentHandle(static_cast<const ezComponentHandle&>(other)) {}
+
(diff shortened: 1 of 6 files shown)


jankrassnigg:
Thanks for the PR ! I didn't expect you to open it up and thought I need to look into this, so this saves me quite a bit of effort :)
I'm not sure whether we need to test the serialization. They are indeed serialized very rarely, I couldn't find any example right now. Though if someone does it at some point and it doesn't work, we can follow up on that as needed.

Code/Engine/Core/World/Declarations.h
```diff
@@ -143,6 +143,25 @@ struct ezComponentHandle
   friend class ezComponent;
 };
 
+template <typename TYPE>
+struct ezTypedComponentHandle : public ezComponentHandle
```
jankrassnigg, line 153: Some comment would be nice, explaining what the advantage / disadvantage over an ezComponentHandle is and when you would prefer one or the other.

jankrassnigg (approved):

Code/Engine/Core/World/Declarations.h
RiederAlex: Added a comment explaining the usage in fce4414

RiederAlex:
>  I didn't expect you to open it up and thought I need to look into this

I told you I'll intent to finish it, just took me some time to get around to it ;)

PR #1204, opened 2024-02-13
Added asset filter tags to only select compatible assets

jankrassnigg:
Some assets can't be used in every situation.
* Materials aren't compatible with all meshes, in case the meshes require certain vertex streams or behavior
* CustomData assets are really only compatible if they have a common base class

These assets are shown in the asset browser UI regardless, which can be confusing and lead to errors that are difficult to reason about.

This change adds that assets can have tags, which are used to filter out incompatible assets. They still all show up in the asset browser panel, but not in dialog mode.

If a component requires a special "sub-type" of an asset, it can now do that through a second argument on the ezAssetBrowserAttribute like so:

image

In the case of CustomData assets, the asset type itself adds the class name (and all base classes) as tags to the asset when the document gets saved (meaning you need to modify and re-save all CustomData assets for the tags to be there).

For materials, it is hard to say whether these tags can be automated, for now they are editable in the UI:

image

The Kraut components require one of those tags, and thus those materials now only show up for the trees, but not for regular meshes.

The "empty tag" ("") is treated special, as the "default tag". To actually see ALL assets, a component would specifically need to ask for "*". This is so that we don't need to migrate all materials to have a new default tag, and specify that everywhere in code as well.


jankrassnigg:
@x0r04rg once this goes in, you need to modify the ezAssetBrowserAttribute's in your code, and modify / re-save all your CustomData assets. Otherwise the filtering will be all wrong. Unfortunately there is no way to automatically migrate this somehow.

Code/EditorPlugins/Assets/EditorPluginAssets/CustomDataAsset/CustomDataAsset.cpp
```diff
@@ -52,3 +52,29 @@ ezTransformStatus ezCustomDataAssetDocument::InternalTransformAsset(ezStreamWrit
...
+{
+  SUPER::UpdateAssetDocumentInfo(pInfo);
+
+  const ezDocumentObject* pObject = GetPropertyObject();
+
+  ezVariant type = pObject->GetTypeAccessor().GetValue("Type");
+  EZ_ASSERT_DEV(type.IsA<ezUuid>(), "Implementation error");
+
+  if (const ezDocumentObject* pDataObject = pObject->GetChild(type.Get<ezUuid>()))
+  {
```
Sanakan8472: Try not to use `GetTypeAccessor` on the object or `GetCild` as these functions are implementation details and require various assumptions. Better use the `ObjectAccessor`:

```suggestion
  const ezUuid typeGuid = GetObjectAccessor()->Get<ezUuid>(pObject, "Type");
  if (const ezDocumentObject* pDataObject = GetObjectAccessor()->GetObject(typeGuid))
  {
```

Code/Engine/Foundation/Reflection/Implementation/PropertyAttributes.h
```diff
@@ -462,23 +462,34 @@ class EZ_FOUNDATION_DLL ezAssetBrowserAttribute : public ezTypeWidgetAttribute
...
-  ezAssetBrowserAttribute(const char* szTypeFilter,
-    ezBitflags<ezDependencyFlags> depencyFlags = ezDependencyFlags::Thumbnail | ezDependencyFlags::Package)
+  ezAssetBrowserAttribute(const char* szTypeFilter, ezBitflags<ezDependencyFlags> depencyFlags = ezDependencyFlags::Thumbnail | ezDependencyFlags::Package, const char* szRequiredTag = "")
     : m_DependencyFlags(depencyFlags)
   {
     SetTypeFilter(szTypeFilter);
+    m_sRequiredTag = szRequiredTag;
+  }
+
+  ezAssetBrowserAttribute(const char* szTypeFilter, const char* szRequiredTag)
```
Sanakan8472: Instead of hardcoding the flags, I'd prefer to just add it at the end as a default argument and then remove `szRequiredTag` from the other ctor.
That way, you have one ctor
```
(const char* szTypeFilter, ezBitflags<ezDependencyFlags> depencyFlags = ...)
```
And another
```
(const char* szTypeFilter, const char* szRequiredTag, ezBitflags<ezDependencyFlags> depencyFlags = ...)
```
This would still not require to fix all the old code but would have the benefit that both have the flags at the end with default values. In the current state, if you want to add flags to the second ctor, you cab't add it at the end, you need to insert the values in the middle which feels a bit unituitive.
Really the only thing that bugs me about this is that two ctors have different arrangement of the same properties, don't know if I can convey this properly.

Code/UnitTests/FoundationTest/Strings/PathUtilsTest.cpp
```diff
@@ -154,12 +157,15 @@ EZ_CREATE_SIMPLE_TEST(Strings, PathUtils)
...
     EZ_TEST_BOOL(ezPathUtils::IsSubPath_NoCase("C:/DataDir", "C:/DataDir/SomeFolder"));
+    EZ_TEST_BOOL(ezPathUtils::IsSubPath_NoCase("C:/DataDir/", "C:/DataDir/SomeFolder"));
     EZ_TEST_BOOL(ezPathUtils::IsSubPath_NoCase("C:/DataDir", "C:/DataDir"));
+    EZ_TEST_BOOL(ezPathUtils::IsSubPath_NoCase("C:/DataDir/", "C:/DataDir"));
     EZ_TEST_BOOL(ezPathUtils::IsSubPath_NoCase("C:/DataDir", "C:/DataDir/"));
+    EZ_TEST_BOOL(ezPathUtils::IsSubPath_NoCase("C:/DataDir/", "C:/DataDir/"));
     EZ_TEST_BOOL(!ezPathUtils::IsSubPath_NoCase("C:/DataDir", "C:/DataDir2"));
 
     EZ_TEST_BOOL(ezPathUtils::IsSubPath_NoCase("C:\\DataDir", "C:/DataDir/SomeFolder"));
-    EZ_TEST_BOOL(ezPathUtils::IsSubPath_NoCase("C:\\DataDir", "C:/DataDir"));
```
Sanakan8472, line 162: If you removed this test case, does that mean it no longer works?

jankrassnigg: Oops, that wasn't planned.

Sanakan8472 (approved):

PR #1228, opened 2024-03-03
NavMeshObstacleComponent

x0r04rg:
Currently only static obstacles are supported.

### DIFF
--- a/Code/Engine/Core/Interfaces/PhysicsWorldModule.h
+++ b/Code/Engine/Core/Interfaces/PhysicsWorldModule.h
@@ -170,7 +170,7 @@ public:
   virtual void AddFixedJointComponent(ezGameObject* pOwner, const ezPhysicsWorldModuleInterface::FixedJointConfig& cfg) {}
 
   /// \brief Gets world space bounds of a physics object if its shape type is included in shapeTypes and its collision layer interacts with uiCollisionLayer.
-  virtual ezBoundingBoxSphere GetWorldSpaceBounds(ezGameObject* pOwner, ezUInt32 uiCollisionLayer, ezBitflags<ezPhysicsShapeType> shapeTypes, bool bRecursive) const { return ezBoundingBoxSphere::MakeInvalid(); }
+  virtual ezBoundingBoxSphere GetWorldSpaceBounds(ezGameObject* pOwner, ezUInt32 uiCollisionLayer, ezBitflags<ezPhysicsShapeType> shapeTypes, bool bIncludeChildObjects) const { return ezBoundingBoxSphere::MakeInvalid(); }
 };
 
 /// \brief Used to apply a physical impulse on the object
--- a/Code/Engine/Core/Interfaces/PhysicsWorldModule.h
+++ b/Code/Engine/Core/Interfaces/PhysicsWorldModule.h
@@ -168,6 +168,9 @@ public:
 
   /// \brief Adds a fixed joint to pOwner.
   virtual void AddFixedJointComponent(ezGameObject* pOwner, const ezPhysicsWorldModuleInterface::FixedJointConfig& cfg) {}
+
+  /// \brief Gets world space bounds of a physics object if its shape type is included in shapeTypes and its collision layer interacts with uiCollisionLayer.
+  virtual ezBoundingBoxSphere GetWorldSpaceBounds(ezGameObject* pOwner, ezUInt32 uiCollisionLayer, ezBitflags<ezPhysicsShapeType> shapeTypes, bool bRecursive) const { return ezBoundingBoxSphere::MakeInvalid(); }
 };
 
 /// \brief Used to apply a physical impulse on the object
--- a/Code/EnginePlugins/AiPlugin/Navigation/Components/NavMeshObstacleComponent.cpp
+++ b/Code/EnginePlugins/AiPlugin/Navigation/Components/NavMeshObstacleComponent.cpp
@@ -8,6 +8,12 @@
 // clang-format off
 EZ_BEGIN_COMPONENT_TYPE(ezNavMeshObstacleComponent, 1, ezComponentMode::Static)
 {
+  EZ_BEGIN_FUNCTIONS
+  {
+    EZ_SCRIPT_FUNCTION_PROPERTY(InvalidateSectors),
+  }
+  EZ_END_FUNCTIONS;
+
   EZ_BEGIN_ATTRIBUTES
   {
     new ezCategoryAttribute("AI/Navigation"),
@@ -65,7 +71,7 @@ void ezNavMeshObstacleComponent::InvalidateSectors()
     auto bounds = pPhysics->GetWorldSpaceBounds(GetOwner(), uiCollisionLayer, ezPhysicsShapeType::Static, true);
     if (bounds.IsValid())
     {
-      pNavMesh->InvalidateSector(bounds.GetBox().GetCenter().GetAsVec2(), bounds.GetBox().GetHalfExtents().GetAsVec2());
+      pNavMesh->InvalidateSector(bounds.GetBox().GetCenter().GetAsVec2(), bounds.GetBox().GetHalfExtents().GetAsVec2(), false);
     }
   }
 }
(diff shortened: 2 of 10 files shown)


Code/EnginePlugins/AiPlugin/Navigation/NavMesh.h
```diff
@@ -95,7 +95,8 @@ class EZ_AIPLUGIN_DLL ezAiNavMesh final
   /// Returns true, if all the sectors are already available, false when any of them needs to be built first.
   bool RequestSector(const ezVec2& vCenter, const ezVec2& vHalfExtents);
 
-  // void InvalidateSector(SectorID sectorID);
+  void InvalidateSector(const ezVec2& vCenter, const ezVec2& vHalfExtents);
```
jankrassnigg: Please add a comment what invalidation means, ie what effect it has.

Code/Engine/Core/Interfaces/PhysicsWorldModule.h
```diff
@@ -168,6 +168,9 @@ class EZ_CORE_DLL ezPhysicsWorldModuleInterface : public ezWorldModule
 
   /// \brief Adds a fixed joint to pOwner.
   virtual void AddFixedJointComponent(ezGameObject* pOwner, const ezPhysicsWorldModuleInterface::FixedJointConfig& cfg) {}
+
+  /// \brief Gets world space bounds of a physics object if its shape type is included in shapeTypes and its collision layer interacts with uiCollisionLayer.
+  virtual ezBoundingBoxSphere GetWorldSpaceBounds(ezGameObject* pOwner, ezUInt32 uiCollisionLayer, ezBitflags<ezPhysicsShapeType> shapeTypes, bool bRecursive) const { return ezBoundingBoxSphere::MakeInvalid(); }
```
jankrassnigg: Please add a short comment what recursive means here, ie that without it the bounds of child objects are ignored and.

Code/EnginePlugins/AiPlugin/Navigation/Implementation/NavMesh.cpp
```diff
@@ -115,20 +115,39 @@ bool ezAiNavMesh::RequestSector(const ezVec2& vCenter, const ezVec2& vHalfExtent
...
+  auto it = m_Sectors.Find(sectorID);
+  if (!it.IsValid())
+    return;
+
+  auto& sector = it.Value();
+
+  if (sector.m_FlagInvalidate == 0 && (sector.m_FlagUsable == 1 || sector.m_FlagUpdateAvailable == 1))
+  {
+    sector.m_FlagInvalidate = 1;
+    m_RequestedSectors.PushBack(sectorID);
```
jankrassnigg: This would recreate the sector right away. Another option would be to just unload it, and only recreate it again on demand when another path search comes in.
Or, a third option would be to make this an option to the Invalidate call, whether it should be recreated right away. Then one can use the Invalidate function to also unload sectors, or at least not uncur overhead for recreation, when it isn't needed at the moment.

I think I would prefer the last option, then the Invalidate function can be used for both.

Code/EnginePlugins/AiPlugin/Navigation/Components/NavMeshObstacleComponent.h
```diff
@@ -0,0 +1,23 @@
+﻿#pragma once
+
+#include <AiPlugin/AiPluginDLL.h>
+#include <Core/World/Component.h>
+#include <Core/World/ComponentManager.h>
+
+using ezNavMeshObstacleComponentManager = ezComponentManager<class ezNavMeshObstacleComponent, ezBlockStorageType::Compact>;
+
+class EZ_AIPLUGIN_DLL ezNavMeshObstacleComponent : public ezComponent
```
jankrassnigg, line 19: Please add a useful comment explaining what this component does and what a typical use case would be. Since it doesn't have any properties, it is quite mysterious and without looking at the implementation details there is no way to know what it will do.
Also it looks like you made it work specifically when it gets deactivated and reactivated. Maybe we should also just expose the InvalidateSectors() function to scripts (and make it public), so that it can be triggered at any time ?

Code/EnginePlugins/AiPlugin/Navigation/Components/NavMeshObstacleComponent.cpp
```diff
@@ -0,0 +1,71 @@
...
+void ezNavMeshObstacleComponent::OnDeactivated()
+{
+  InvalidateSectors();
+
+  SUPER::OnDeactivated();
+}
+
+void ezNavMeshObstacleComponent::InvalidateSectors()
+{
+  // TODO: dynamic obstacles not implemented yet
```
jankrassnigg, line 53: So dynamic objects "don't work" because the navmesh only considers static ones, right? Though it might be made to work at least with certain kinematic objects (assuming you trigger the navmesh update yourself after e.g. a door moved). Is that what you have in mind here?

Code/EnginePlugins/AiPlugin/Navigation/Components/NavMeshObstacleComponent.h
x0r04rg: Yea, you're right, I wanted to make it public but forgot about it

Code/EnginePlugins/AiPlugin/Navigation/Components/NavMeshObstacleComponent.cpp
x0r04rg: Yeah, and I was also considering automatically tracking if an object has moved and invalidating both previous sectors and new ones if it has. So I decided to postpone it until later and do it in a separate PR (or someone else could do it). I'll investigate it tomorrow and see if we can make navmesh consider non-static objects and then after making `ezNavMeshObstacleComponent::InvalidateSectors` public it could be useful with kinematic objects

Code/EnginePlugins/AiPlugin/Navigation/Implementation/NavMesh.cpp
x0r04rg: Good idea, I'll do as you say. We should also keep in mind that some users may want to always keep all sectors loaded and up to date and there should be an option for that

Code/Engine/Core/Interfaces/PhysicsWorldModule.h
x0r04rg: Renamed bRecursive to bIncludeChildObjects. Should I add a comment as well?

jankrassnigg: That's good enough I think.

jankrassnigg (approved):

jankrassnigg:
I'm happy with the PR, is it ready to be merged ?

x0r04rg:
Thanks! Yes, I think it's ready.

I want to bring up a few thoughts I had as well. 

First, I find `ezAiNavMesh::RequestSector` a bit confusing because it's a fire and forget operation, and it's not clear for how long those sectors should stay requested. A more intuitive approach would be to treat sectors like resources and introduce a "lock" class that keeps sectors up to date. That could also help with invalidating paths later down the road.

And regarding dynamic objects probably the best way is to add them to a list in each navmesh, so that when collecting geometry it could collect all static geometry and then add those obstacles from the list on top of that. Detour does similar thing with `dtTileCache::addObstacle`.

Well, those are just my thoughts on how this could be improved in the future

jankrassnigg:
> Thanks! Yes, I think it's ready.
> 
> I want to bring up a few thoughts I had as well.
> 
> First, I find `ezAiNavMesh::RequestSector` a bit confusing because it's a fire and forget operation, and it's not clear for how long those sectors should stay requested. A more intuitive approach would be to treat sectors like resources and introduce a "lock" class that keeps sectors up to date. That could also help with invalidating paths later down the road.
> 
> And regarding dynamic objects probably the best way is to add them to a list in each navmesh, so that when collecting geometry it could collect all static geometry and then add those obstacles from the list on top of that. Detour does similar thing with `dtTileCache::addObstacle`.
> 
> Well, those are just my thoughts on how this could be improved in the future

Agreed, there is lots of room for future improvements. This is really just the first iteration of the system and if you have concrete things you want to change, feel free to open discussions. I'm not sure about the dynamic obstacle thing, atm, because since they can't affect the navmesh itself, I don't think there is a reason to even track them there. I would rather use the spatial system to just query objects in the vicinity for dynamic avoidance. But anyway, we can improve this together over time.

PR #1230, opened 2024-03-05
MSFT contributions March 2024

mafaldag1:


jankrassnigg:
Looks good, as far as I can tell from a quick glance, thanks Mafa!

Issue #1210, opened 2024-02-17
[FEATURE] Enable EditorEngineProcess callstack printing & make it clickable


RiederAlex:
It would be very useful to always enable callstack printing for the EditorEngineProcess and be able to jump directly to the source of the problem going from ezEditor -> Code IDE by clicking on the various entries in the callstack.

I have a prototype with hardcoded paths working, which looks like this:

https://github.com/ezEngine/ezEngine/assets/14026803/98037624-fb54-4828-a616-ffdcc04144cc

Here is my proposal how I would like to implement this properly:
1. Always enable the CrashHandler in `ezEngineProcessGameApplication::AfterCoreSystemsStartup`
2. Add static global callbacks for double-clicking on LogMessages in `ezQtLogWidget`
    This is then something that can be used by each editor plugin to provide specific actions when double clicking on an log entry
3. Implement callstack-logmessage parsing somwehre in the Editor, using this new callback
4. Allow the user to configure a "Open in IDE" action via Application: General settings
    This would be a freeform text field which the user can configure to point towards their IDE, like:
      `..../clion64.exe --line {line} {file}`
    where {line} and {file} would be dynamically replaced with the information parsed from step 3).

Potentially problematic points I've identified so far/I am unsure about:

5. Log messages have no context, which means that log message parsing is plain string parsing. Potentially it would be nice to have some form of metadata attached to log messages.
  This is not scope of this feature, but just something to keep in mind. To ensure this plaintext parsing behaves well in this case, I'll add unit tests.
6. Currently this action is done by double clicking on a log message. However in the future it could happen that multiple plugins parse a log message and want to perform actions. I'm thinking of changing the design to be right click -> open context menu and have a list of plausible actions.
7. The callstack-logmessage parsing has to be implemented for each platform and I'm not sure how the callstack looks in Linux/if its even possible to open a file without additional meta information. Right now this would be windows only.

Implementing this as a first step would allow a couple of nice things in the future potentially, like jumping to a component definition (as we now have "open this file/line in IDE" action) or if its decided to add more metadata to the log messages, be able to jump directly to the gameobject/component/system which logged certain messages.

If this sounds like the right approach I would like to take over implementation on this.

jankrassnigg:
This is a really great suggestion. I'd be happy for you to implement it in the proposed way.

A couple of random thoughts.

[...]

PR #1232, opened 2024-03-07
Enable EditorEngineProcess callstack printing & make it clickable

RiederAlex:
This implements #1210 for windows

- Enabled crash handler for EditorEngineProcess
- Added ability to configure a code editor in Preferences -> Application: C++ Projects
- Double clicking a stacktrace in the engine log now launches the configured editor, showing the file/line parsed from the stacktrace.


jankrassnigg:
What would be a way to test this? When I add an assert into a component in the Monster Attack sample, I get the assert message in the log, but no callstack.

Also what editor + arguments are you using for testing? We already have the C++ IDE setting, why don't we just use that?

RiederAlex:
I tested it like this in my component's Update(): `int* bla = nullptr; *bla = 5;`
adding an assert like `EZ_ASSERT_DEV(false,"Assert!");` in the same place also triggered a callstack for me in **Dev** configuration.

However I just noticed that in **Debug** when the "Debug Assertion Failed" message pops up one has to press "Retry" for it to produce a stacktrace. I'll add support for double clicking the initial assert message as well, should be easy enough.

For testing/developing I use Clion, but from quickly checking most code editors support some kind of "open file and jump to line" CLI command, here are some examples I could test with the stuff I had installed:

- Clion: `clion64.exe {file} --line {line}`
- Sublime: `sublime_text.exe {file}:{line}`
- VSCode: `code.exe -g {file}:{line}`
- VisualStudio unfortunately is very special and would require custom code/scripts, as it does not expose "open file and jump to line" via CLI (See [this StackOverflow post](https://stackoverflow.com/questions/350323/open-a-file-in-visual-studio-at-a-specific-line-number) for possible solutions)

**Why are we not using the C++ IDE Settings?**
I've looked at this setting for my initial approach (`ezCppProject::ezIDE`) but came to the conclusion that it does not fit the needs of this feature right now, as it does not expose the editor binary directly and does not allow configuring custom editors - something which I find very important, as one cannot possibly support all of them out of the box :)

However - usability wise I agree that this would be a better choice from a user perspective, having one setting instead of two.
What I could do (if this is still in the scope of the PR):
  - Add another IDE (calling it 'Custom') so we do not loose the "I have my own esoteric code editor" usecase.
  - Figure out the path for supported editors automatically (similar to what happens in `ezQtUiServices::OpenInVsCode()`). This would mean adding extra code to support/find different VisualStudio installations and letting the user select a specific one.
  - Auto-fill m_sEditorPath/m_sEditorArgs if a supported editor is selected (making these fields read-only/hide them).
  - Let the user manually fill out these fields if IDE is set to "Custom"
  
Is this something I should do in this PR or do you want me to create a new one after this one is accepted?

jankrassnigg:
We can keep these as separate options. Because we only support certain compilers, but may want to allow users to use different text editors, also for other use cases in the future.
What would be nice, though, is a drop down with a few editors that are Auto detected, and for which we know the command lines. And one option would be custom, which then shows the two fields that we have now.

Whether you want to add this, at all, in this PR or in a follow up PR, I'll leave up to you.

RiederAlex:
I'll try to get some version of auto-detection working in this PR - especially now that I've realized special handling is needed as VS is not exposing what I would have expected via its CLI.

RiederAlex:
I've now added support for double clicking assert messages and for using VisualStudio as a code editor on Windows.

It now looks like this if VisualStudio is selected (default on Windows):
image

And it looks like this if a custom editor is configured:
image

VisualStudio support was a bit tricky to get working correctly - it now uses a script (Utilities\Scripts\open-in-msvs.vbs) in order to open up the latest installed VisualStudio and jump to the file/line. This might seem a bit odd, but it was required as VS does not expose a CLI to do that consistently. I chose this approach as it avoids doing a bunch of COM Interface stuff in C++ directly.

RiederAlex:
I've fixed the Linux compile error, but Clang-Tidy seems to have some other problem unrelated to these changes.
Also seems like ClangFormat is not re-run automatically - it still reports issues which should be fixed with a6485da6a8199d4c927af07832690e7542386a63

jankrassnigg:
I think I fixed the clang-tidy issue in the mean time.
Don't know about clang-format, it can't properly run on forked repos, that was always a problem. @Ingrater fixed some of that, but it looks like it's still not fully working.

Sanakan8472:
So due to access restrictions, clang format can't apply formatting changes on forked repos, here is the error message:
```
This PR comes from the forked repository 'RiederAlex/ezEngine'
Formatting changes can't be applied automatically.
Please download the attached patch file and apply it manually using:
git apply -p0 clangFormat.patch
You can also run ClangFormat on the necessary files (see patch) from within Visual Studio.
Error: Process completed with exit code 1.
```
So @RiederAlex, can you please download the generated [ClangFormatPatch](https://github.com/ezEngine/ezEngine/actions/runs/8333352095/artifacts/1336679132) and apply it to your branch? Thanks!

RiederAlex:
> So due to access restrictions, clang format can't apply formatting changes on forked repos, here is the error message:
> 
> ```
> This PR comes from the forked repository 'RiederAlex/ezEngine'
> Formatting changes can't be applied automatically.
> Please download the attached patch file and apply it manually using:
> git apply -p0 clangFormat.patch
> You can also run ClangFormat on the necessary files (see patch) from within Visual Studio.
> Error: Process completed with exit code 1.
> ```
> 
> So @RiederAlex, can you please download the generated [ClangFormatPatch](https://github.com/ezEngine/ezEngine/actions/runs/8333352095/artifacts/1336679132) and apply it to your branch? Thanks!

I've already did that in a force push in a6485da6a8199d4c927af07832690e7542386a63 - thats why I said that unfortunately it seems like the clang-format job does not re-run with my latest changes (maybe it does not like force pushes?).

I'm a bit lost on how I should trigger this job again from my side.

jankrassnigg:
Ah sorry, forgot about this PR. I don't think the clang-format stuff is important, I see you applied the changes, so I guess the rest is a bug in our system.
Is this good to go in now? Then I'll just submit it.

RiederAlex:
Yes - the only thing that I'm a bit unsure about is the location I've put the required .vbs script (see reasoning [here](https://github.com/ezEngine/ezEngine/pull/1232#issuecomment-2002101387)) and for the moment it looks a bit strange, as both "C++ IDE" and "Code Editor Preferences" can be configured, which might confuse users.

If you think both of these are acceptable then go ahead and submit it :)

jankrassnigg:
I don't really have time to test this in detail atm, so I'll just close the PR and then we can polish things as we notice them over time.

PR #1241, opened 2024-03-15
Sync MSFT

mafaldag1:


Code/Tools/Libs/ToolsFoundation/FileSystem/DataDirPath.h
```diff
@@ -86,9 +86,9 @@ class EZ_TOOLSFOUNDATION_DLL ezDataDirPath
 
 private:
   ezString m_sAbsolutePath;
-  mutable ezUInt16 m_uiDataDirParent = 0;
-  mutable ezUInt8 m_uiDataDirLength = 0;
-  mutable ezUInt8 m_uiDataDirIndex = 0;
+  mutable ezUInt32 m_uiDataDirParent = 0;
```
Sanakan8472: This brakes serialization. I'm curious what the intend behind this change was? What is the use case in which you use this class and break it's memory limits?
* **m_uiDataDirParent**: I doubt we will break out of the 65k character limit for the data dir path as Windows already chokes at 255 bytes.
* **m_uiDataDirLength**: This is just the length of the folder name, again, with Windows choking at 255 I doubt any folder would be 255 characters on its own.
* **m_uiDataDirIndex**: Do you have a genuine use case in which you have over 256 data directories which would require this to be wider?

Sanakan8472 (changes requested):

Code/Tools/Libs/ToolsFoundation/FileSystem/DataDirPath.h
jankrassnigg: Isn't this editor specific code anyway?

Sanakan8472: Yes it's tools only code. As we creates tens of thousands of these, size does matter in this case. I guess more strict compiler options might complain about some of the downcasts.

jankrassnigg: My point is mostly, that I am surprised for *this user* to even compile and use this code. To my knowledge that should not be the case.

Sanakan8472: I'll follow up offline tomorrow.

mafaldag1: updated with reverted changes on DataDirPath.h :)

PR #1244, opened 2024-03-27
Basic Android Support

Sanakan8472:
# Build System
* `Configure_Android.cmake`: Very similiar to the linux variant with two exceptions:
  * **Vulkan**: As we cross compile for Android on both Windows and Linux, we need to handle both versions of the Vulkan SDK which have slightly different folder structures. Also there is no shader compiler, so a dummy entry is added for `EzVulkan::DXC`.
  * **Clang**: `LINKER:--undefined-glob=*ezReferenceFunction*` This addition prevents the linker from dropping the `EZ_STATICLINK_LIBRARY` code blocks which would optimize away the entire statically linked plugin.
* `ez_requires_desktop`: This is a new macro that is added now to all tools / samples that require a desktop environment like `EZ_CMAKE_PLATFORM_WINDOWS_DESKTOP` or `EZ_CMAKE_PLATFORM_LINUX`. This is different from `ez_requires_editor` which can be toggled on or off by the user.

# Code
* Static linking of plugins: When `EZ_COMPILE_ENGINE_AS_DLL` is disabled, the `EZ_STATICLINK_LIBRARY` will register the **LibraryName** preprended with **ez** in the `ezPlugin` system. While not all libraries are plugins, this was the simplest way to figure out which plugins are actually available after static linking. Additionally, the `ezReferenceFunction_` functions are now `extern "C"` to prevent name mangling which makes preventing them from being optimized away easier (see **clang** section above).
* Turned off `EZ_SUPPORTS_PROCESSES` for Android.
* Android app startup is no longer bound to the start of the native activity. Instead, we now wait for `APP_CMD_INIT_WINDOW` to ensure that a window can be created by the app.
* Implemented `ezScreen` for Android.
* `AndroidManifest.xml`:
  * Added `uses-sdk` section with `minSdkVersion` set to 26, the first version supporting Vulkan.
  * Added `uses-permission` for `INTERNET` and `ACCESS_NETWORK_STATE` to support `ezInspector` out of the box.

# Misc
* Fixed a bug in `ezDataDirectory` when querying files in an archive with something like `AAA/../BBB` which would fail.
* `ezLogWriter::Console` now writes each line in one go. This is neccessary as Android does not support writing a log line with multiple commands. Each `printf` call will be a new line automatically in logcat.
* Logging: Added `DEBUG_STARTUP_LOGGING` which can be set in `Log.cpp` and allows to write all log messages that happen before loggers are registered to the `ezLog::Print` function. Quite useful when debugging startup systems.
* Disabled **ozz** SIMD support on Linux and Android for now. On Linux asset conversion triggers a segfault for some reason.

# Unit Tests
* `EZ_TESTFRAMEWORK_ENTRY_POINT` macros needed to be rewritten similiar to how the normal app startup works to support renderer tests in the future.
* **FoundationTest** and **CoreTest** run successfully and Android now.

# Debugging

* You can open the apk in Android Studio to debug like any other apk.
* VSCode debugging support
  * Require CodeLLDB extension.
  * `DbgAndroidLldb.ps1` starts apk app for testing and writes `lldb.setup` file.
  * `command source lldb.setup` will setup lldb to connect to the started app.
  * `DbgAndroidJdb.ps1` connects the java debugger to unblock the app start.
  * Below is an example of a `launch.json` config for VSCode:
```
{
	"name": "Android lldb",
	"type": "lldb",
	"request": "custom",
	"targetCreateCommands": [
		"shell pwsh \"${workspaceRoot}/Utilities/DbgAndroidLldb.ps1\" -PackageName \"com.ezengine.FoundationTest\" -debugTemp \"${workspaceRoot}/debugtemp\" -apk \"${workspaceRoot}/Output/Bin/AndroidNinjaClangDebugArm64/FoundationTest.apk\" -MessageBoxOnError",
		"command source ${workspaceRoot}/debugtemp/lldb.setup"
	],
	"processCreateCommands": [
		"shell pwsh \"${workspaceRoot}/Utilities/DbgAndroidJdb.ps1\" &",
		"continue"
	]
}
```


Code/BuildSystem/CMake/CMakeUtils/ezUtilsRenderer.cmake
```diff
@@ -23,9 +23,11 @@ function(ez_add_renderers TARGET_NAME)
 			RendererVulkan
 		)
 
-		add_dependencies(${TARGET_NAME}
-			ShaderCompilerDXC
-		)
+		if(EZ_CMAKE_PLATFORM_WINDOWS_DESKTOP OR EZ_CMAKE_PLATFORM_LINUX)
```
jankrassnigg: You can also write

```
if (TARGET ShaderCompilerDXC)
...
```

That checks whether the target exists. On all platforms where the project is set up, it will then add it.

Code/BuildSystem/CMake/CMakeUtils/ezUtilsRenderer.cmake
```diff
@@ -35,8 +37,10 @@ function(ez_add_renderers TARGET_NAME)
 		)
 		ez_link_target_dx11(${TARGET_NAME})
 
-		add_dependencies(${TARGET_NAME}
-			ShaderCompilerHLSL
-		)
+		if(EZ_CMAKE_PLATFORM_WINDOWS_DESKTOP OR EZ_CMAKE_PLATFORM_LINUX)
```
jankrassnigg: Same here

Code/Engine/Foundation/Platform/Android/Screen_Android.cpp
```diff
@@ -1,11 +1,22 @@
...
+#include <Foundation/System/Screen.h>
 
-#  if EZ_ENABLED(EZ_SUPPORTS_GLFW)
-#    include <Foundation/Platform/GLFW/Screen_GLFW.h>
-#  else
-#    include <Foundation/Platform/NoImpl/Screen_NoImpl.h>
-#  endif
+#if EZ_ENABLED(EZ_PLATFORM_ANDROID)
 
+ezResult ezScreen::EnumerateScreens(ezHybridArray<ezScreenInfo, 2>& out_Screens)
```
jankrassnigg: We could change this function to just take a dynamicarray.

Code/Engine/Foundation/Platform/Android/Screen_Android.cpp
```diff
@@ -1,11 +1,22 @@
...
+  {
+    ezScreenInfo& currentScreen = out_Screens.ExpandAndGetRef();
+    currentScreen.m_sDisplayName = "Current Display";
+    currentScreen.m_iOffsetX = 0;
+    currentScreen.m_iOffsetY = 0;
+    currentScreen.m_iResolutionX = ANativeWindow_getWidth(pWindow);
+    currentScreen.m_iResolutionY = ANativeWindow_getHeight(pWindow);
+    currentScreen.m_bIsPrimary = true;
+  }
+  return EZ_SUCCESS;
```
jankrassnigg: Wouldn't it be a failure, if there is no screen ?

Code/UnitTests/TestFramework/Framework/TestFramework.h
```diff
@@ -241,25 +241,100 @@ class EZ_TEST_DLL ezTestFramework
...
-      /* TODO: do something with the return value of ezAndroidMain?  */                                                    \
-      /* TODO: can we get somehow get the command line arguments to the android app? Is there even something like that? */ \
-      int iReturnCode = ezAndroidMain(0, nullptr);                                                                         \
-      __android_log_print(ANDROID_LOG_ERROR, "ezEngine", "Test framework exited with return code: '%d'", iReturnCode);     \
-    }                                                                                                                      \
-                                                                                                                           \
-    int ezAndroidMain(int argc, char** argv)                                                                               \
-    {                                                                                                                      \
-      ezTestSetup::InitTestFramework(szTestName, szNiceTestName, 0, nullptr);                                              \
+class ezAndroidTestApplication
```
jankrassnigg: I wasn't aware we have Android stuff directly in TestFramework.h.
Can we move this to a separate file ?

Code/UnitTests/CoreTest/Scripting/TypeScriptTest.cpp
```diff
@@ -92,7 +92,11 @@ EZ_CREATE_SIMPLE_TEST(Scripting, TypeScript)
 
     ezStringBuilder sTestDataDir(">sdk/", ezTestFramework::GetInstance()->GetRelTestDataPath());
     sTestDataDir.AppendPath("Scripting/TypeScript");
-    if (!EZ_TEST_RESULT(ezFileSystem::AddDataDirectory(sTestDataDir, "TypeScriptTest", "TypeScriptTest", ezFileSystem::AllowWrites)))
+    if (!EZ_TEST_RESULT(ezFileSystem::AddDataDirectory(sTestDataDir, "TypeScriptTest")))
+      return;
+
+    const ezStringBuilder sWriteDir = ezTestFramework::GetInstance()->GetAbsOutputPath();
```
jankrassnigg, line 98: Why does this test need this directory now and not before?

Code/Engine/Foundation/Configuration/Implementation/Plugin.cpp
```diff
@@ -18,6 +18,12 @@
 ezResult UnloadPluginModule(ezPluginModule& ref_pModule, ezStringView sPluginFile);
 ezResult LoadPluginModule(ezStringView sFileToLoad, ezPluginModule& ref_pModule, ezStringView sPluginFile);
 
+ezDynamicArray<ezString>& GetStaticPlugins()
+{
+  static ezDynamicArray<ezString> s_StaticPlugins;
+  return s_StaticPlugins;
+}
```
jankrassnigg, line 25: How is this being used? Do we override that function somehow somewhere?

jankrassnigg (approved):

Code/BuildSystem/CMake/CMakeUtils/ezUtilsRenderer.cmake
Sanakan8472: Oh nice, didn't know this works.

Code/Engine/Foundation/Platform/Android/Screen_Android.cpp
Sanakan8472: Will do!

Code/Engine/Foundation/Platform/Android/Screen_Android.cpp
Sanakan8472: Good point, wasn't sure what the result means, the other implementations also fail on 0 monitors so will change it.

Code/UnitTests/CoreTest/Scripting/TypeScriptTest.cpp
Sanakan8472: The old code was mapping the `TypeScriptTest` folder from the SDK as both the source dir for the test as well as the target dir for the test output. As this is part of the Data folder, it will be read-only in the apk. Thus, I need to map two directories instead, one for reading the test data and one for writing the temp files the test generates.

Code/Engine/Foundation/Configuration/Implementation/Plugin.cpp
Sanakan8472: So the process for static linking is like this: `EZ_STATICLINK_LIBRARY` calls `ezPluginRegister` which in turn adds the library to this list here (see bottom of this file).
Finally, `InitializeStaticallyLinkedPlugins` will go through the list of *registered* plugins to add them to the loaded plugin list and any subsequent `LoadPlugin` call will then be able to pick up statically linked plugins.

Code/UnitTests/TestFramework/Framework/TestFramework.h
Sanakan8472: Done! I created a `Platform` folder in the TestFramework and extracted the android code and moved the uwp files into it.

PR #1246, opened 2024-03-29
Android Renderer Support

### LINKED ISSUE #236: Add stereo support to default render pipeline
Currently, only the hololens render pipeline supports stereo rendering.

### LINKED ISSUE #411: [FEATURE] ShaderCompiler should be able to compile all shaders for a given project
Currently the shader compiler has to be given a file or folder to compile. It also already gets a project reference, but that is only used to resolve paths.

Instead just giving it a project should compile all shaders in that project. This is especially interesting for projects that have visual shaders for materials.

Once it is available, this should be used in the pipeline that builds release packages.

Sanakan8472:
# Shaders
* ⚠️The `ShaderCache` folder has moved to `>sdk/Output/ShaderCache`. This is because shaders are always the same, regardless of platform, debug/release etc. So it does not make sense to have separate shader folders for all platform flavours if the all use the same Vulkan shaders at the end. This also makes it easier to cross-compile as Android builds can't build their own shaders.
* Fixed all compile errors so that `ShaderCompiler.exe -renderer DX11 -platform DX11_SM50 -project C:\Code\ezEngine\Data\Base -shader Shaders` runs through without errors, same for `-renderer Vulkan -platform VULKAN`. Fixes #411 and fixes #236.

# Renderer / Core
* Device Capabilities: `ezGALResourceFormatSupport` cleaned up a bit and added `TextureRW` as a flag.
* Added Android window support. Android automatically creates a window for you and is very accomadating regarding the size of the swapchain: anything between 1x1 and 4kx4k is allowed on my phone at least. Thus, we enable all window modes and pretend we can handle both windowed and fullscreen windows as we have no choice in that matter anyways and always get the same window.

# Vulkan
* An Android, we require Vulkan 1.0 and `VK_KHR_maintenance1`, on all other platforms Vulkan 1.1.
* Transfer queue is now optional.
* Removed unused fence from swap chain code and added `vk::AndroidSurfaceCreateInfoKHR` support.
* Various compile fixes for pedantic Android clang.
* Static linker tools now runs over Vulkan plugin code.
* Moved required headers into `RendererVulkanDLL.h` to allow other libraries to include stuff without throwing hundreds of errors.

# Renderer Tests
* AdvancedFeatures test now checks for a proper format that can be used for the compute test.
* Added `ifdef` blocks to remove shared texture code on platforms that don't support processes.
* Textures test now checks each texture if its format is supported before testing.
* A shader compiler plugin is now optional as we won't have one on Android.
* All tests pass now on Android. However, `ShaderCompiler.exe` has to be run on the `Data\Base` and the `Assets/Data/UnitTests` before building.

### DIFF
--- a/Code/Engine/Core/System/Implementation/android/Window_android.inl
+++ b/Code/Engine/Core/System/Implementation/android/Window_android.inl
@@ -2,9 +2,9 @@
 
 #include <Core/System/Window.h>
 #include <Foundation/Basics.h>
+#include <Foundation/Basics/Platform/Android/AndroidUtils.h>
 #include <Foundation/Logging/Log.h>
 #include <Foundation/Types/UniquePtr.h>
-#include <Foundation/Basics/Platform/Android/AndroidUtils.h>
 #include <android_native_app_glue.h>
 
 struct ANativeWindow;
@@ -24,24 +24,10 @@ ezResult ezWindow::Initialize()
   }
 
   // Checking and adjustments to creation desc.
-  {
-//    if (m_CreationDescription.m_WindowMode == ezWindowMode::WindowFixedResolution)
-//    {
-//      ezLog::Warning("Only Fullscreen modes are supported on Android. Falling back to ezWindowMode::FullscreenFixedResolution.");
-//      m_CreationDescription.m_WindowMode = ezWindowMode::FullscreenFixedResolution;
-//    }
-//    else if (m_CreationDescription.m_WindowMode == ezWindowMode::WindowResizable)
-//    {
-//      ezLog::Warning("Only FullscreenFixedResolution and FullscreenBorderlessNativeResolution modes are supported on Android. Falling back to ezWindowMode::FullscreenBorderlessNativeResolution.");
-//      m_CreationDescription.m_WindowMode = ezWindowMode::FullscreenBorderlessNativeResolution;
-//    }
+  if (m_CreationDescription.AdjustWindowSizeAndPosition().Failed())
+    ezLog::Warning("Failed to adjust window size and position settings.");
 
-    if (m_CreationDescription.AdjustWindowSizeAndPosition().Failed())
-      ezLog::Warning("Failed to adjust window size and position settings.");
-
-    EZ_ASSERT_RELEASE(m_CreationDescription.m_Resolution.HasNonZeroArea(), "The client area size can't be zero sized!");
-  }
-
+  EZ_ASSERT_RELEASE(m_CreationDescription.m_Resolution.HasNonZeroArea(), "The client area size can't be zero sized!");
   EZ_ASSERT_RELEASE(s_androidWindow == nullptr, "Window already exists. Only one Android window is supported at any time!");
 
   s_androidWindow = ezAndroidUtils::GetAndroidApp()->window;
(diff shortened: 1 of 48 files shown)


Code/Engine/RendererVulkan/Utils/Implementation/ImageCopyVulkan.cpp
```diff
@@ -493,6 +493,8 @@ void ezImageCopyVulkan::RenderInternal(const ezVec3U32& sourceOffset, const vk::
         {
           write.pImageInfo = &sourceInfo;
         }
+        default:
```
jankrassnigg, line 498: should this silently continue ?

Data/Base/Shaders/Materials/SkeletonVisualization.ezShader
```diff
@@ -95,41 +95,40 @@ void main( triangle GS_IN input[3], inout TriangleStream<GS_OUT> outputStream )
 Texture2D BaseTexture;
 SamplerState BaseTexture_AutoSampler;
 
-float edgeFactor(PS_IN Input)
+float edgeFactor()
```
jankrassnigg, line 98: was this shader previously just broken ?

Code/Engine/Core/System/Implementation/android/Window_android.inl
```diff
@@ -0,0 +1,89 @@
...
+//    if (m_CreationDescription.m_WindowMode == ezWindowMode::WindowFixedResolution)
+//    {
+//      ezLog::Warning("Only Fullscreen modes are supported on Android. Falling back to ezWindowMode::FullscreenFixedResolution.");
+//      m_CreationDescription.m_WindowMode = ezWindowMode::FullscreenFixedResolution;
+//    }
+//    else if (m_CreationDescription.m_WindowMode == ezWindowMode::WindowResizable)
+//    {
+//      ezLog::Warning("Only FullscreenFixedResolution and FullscreenBorderlessNativeResolution modes are supported on Android. Falling back to ezWindowMode::FullscreenBorderlessNativeResolution.");
+//      m_CreationDescription.m_WindowMode = ezWindowMode::FullscreenBorderlessNativeResolution;
+//    }
```
jankrassnigg: remove ?

jankrassnigg (approved):

Data/Base/Shaders/Materials/SkeletonVisualization.ezShader
Sanakan8472: Not all permutations, the permutations in which the functions were not called were working. It could be that the other permutations are not used in practice but when compiling all via the shadercompiler tool it would fail. I don't know under which circumstances which permutations are used so I decided to fix the permutations instead of reducing the permutation space.

Code/Engine/Core/System/Implementation/android/Window_android.inl
Sanakan8472: Ah thanks, I first thought I can't support windowed mode but it turns out it's much easier to just pretend you do support resizeable windows than to fix all the high level code. Androdi will "do the right thing" when you select these modes so while all windows are always fullscreen on android, all the other features of a resizable window just work as expected.

Code/Engine/RendererVulkan/Utils/Implementation/ImageCopyVulkan.cpp
Sanakan8472: Good point, that was just my attempt to silence clang. I've added asserts in both unused code paths now.

PR #1247, opened 2024-03-30
Android Basic Input Support

Sanakan8472:
# Input
* `ezVirtualThumbStick` no longer worked as all input manager values are reset during the update phase. This means that no input device can read input from another inout device. To fix this, We now double-buffer the input manager's values. The downside of this is, that `ezVirtualThumbStick` inout is one frame delayed.
* Added Android implementation of the `ezStandardInputDevice`. Currently, only touch input is implemented.
	* Input is generated by the new `ezAndroidUtils::s_InputEvent` event that anything can subscribe to.
* Changed naming of input slots, `ezInputSlot_TouchPoint0` is now named "Touchpoint 0" instead of "Touchpoint 1" etc.

# Telemetry
* Added `DisconnectedFromClient` event handling to the ezInspectorPlugin to fix a memory leak on shutdown.
* Added missing dependency to ezInspectorPlugin for `ResourceManager` to prevent crashing on Android.
* `DisconnectedFromClient` was not guaranteed to fire, so I added safety checks to `ezTelemetry::CloseConnection` to make sure the event is fired on a forced shutdown.
* ezInspector can now connect to Android devices.

# ShaderExplorer
* Added two `ezVirtualThumbStick` for testing touch input.
* Added `EZ_SUPPORTS_DIRECTORY_WATCHER` ifdefs to allow building on Android.
* Added `ezTelemetry::CreateServer` and `CloseConnection` calls to enable ezInspector to connect to the running app.

### DIFF
--- a/Code/Engine/Core/Input/Implementation/InputManager.cpp
+++ b/Code/Engine/Core/Input/Implementation/InputManager.cpp
@@ -155,8 +155,9 @@ ezKeyState::Enum ezInputManager::GetInputSlotState(ezStringView sInputSlot, floa
   if (it.IsValid())
   {
     if (pValue)
-      *pValue = it.Value().m_fValue;
-
+    {
+      *pValue = s_bInputSlotResetRequired ? it.Value().m_fValue : it.Value().m_fValueOld;
+    }
     return it.Value().m_State;
   }
 
@@ -208,6 +209,7 @@ void ezInputManager::ResetInputSlotValues()
   // this is crucial for accumulating the new values and for resetting the input state later
   for (ezInputSlotsMap::Iterator it = GetInternals().s_InputSlots.GetIterator(); it.IsValid(); it.Next())
   {
+    it.Value().m_fValueOld = it.Value().m_fValue;
     it.Value().m_fValue = 0.0f;
   }
 }
--- a/Code/Engine/Core/System/Implementation/android/InputDevice_android.inl
+++ b/Code/Engine/Core/System/Implementation/android/InputDevice_android.inl
@@ -1,0 +1,193 @@
+#include <Core/System/Implementation/android/InputDevice_android.h>
+
+#include <Core/Input/InputManager.h>
+#include <Foundation/Basics/Platform/Android/AndroidUtils.h>
+#include <android/log.h>
+#include <android_native_app_glue.h>
+
+// clang-format off
+EZ_BEGIN_DYNAMIC_REFLECTED_TYPE(ezStandardInputDevice, 1, ezRTTINoAllocator)
+EZ_END_DYNAMIC_REFLECTED_TYPE;
+// clang-format on
+
+// Comment in to get verbose output on android input
+// #  define DEBUG_ANDROID_INPUT
+
+#  ifdef DEBUG_ANDROID_INPUT
+#    define DEBUG_LOG(...) ezLog::Debug(__VA_ARGS__)
+#  else
+#    define DEBUG_LOG(...)
+#  endif
+
+ezStandardInputDevice::ezStandardInputDevice(ezUInt32 uiWindowNumber)
+{
+  ezAndroidUtils::s_InputEvent.AddEventHandler(ezMakeDelegate(&ezStandardInputDevice::AndroidInputEventHandler, this));
+}
+
+ezStandardInputDevice::~ezStandardInputDevice()
+{
+  ezAndroidUtils::s_InputEvent.RemoveEventHandler(ezMakeDelegate(&ezStandardInputDevice::AndroidInputEventHandler, this));
+}
+
+void ezStandardInputDevice::SetShowMouseCursor(bool bShow) {}
+
+bool ezStandardInputDevice::GetShowMouseCursor() const
+{
+  return false;
+}
+
+void ezStandardInputDevice::SetClipMouseCursor(ezMouseCursorClipMode::Enum mode) {}
+
+ezMouseCursorClipMode::Enum ezStandardInputDevice::GetClipMouseCursor() const
+{
+  return ezMouseCursorClipMode::Default;
+}
+
+void ezStandardInputDevice::InitializeDevice()
+{
+}
+
+void ezStandardInputDevice::RegisterInputSlots()
+{
+  RegisterInputSlot(ezInputSlot_TouchPoint0, "Touchpoint 0", ezInputSlotFlags::IsTouchPoint);
+  RegisterInputSlot(ezInputSlot_TouchPoint0_PositionX, "Touchpoint 0 Position X", ezInputSlotFlags::IsTouchPosition);
+  RegisterInputSlot(ezInputSlot_TouchPoint0_PositionY, "Touchpoint 0 Position Y", ezInputSlotFlags::IsTouchPosition);
+
+  RegisterInputSlot(ezInputSlot_TouchPoint1, "Touchpoint 1", ezInputSlotFlags::IsTouchPoint);
+  RegisterInputSlot(ezInputSlot_TouchPoint1_PositionX, "Touchpoint 1 Position X", ezInputSlotFlags::IsTouchPosition);
+  RegisterInputSlot(ezInputSlot_TouchPoint1_PositionY, "Touchpoint 1 Position Y", ezInputSlotFlags::IsTouchPosition);
+
+  RegisterInputSlot(ezInputSlot_TouchPoint2, "Touchpoint 2", ezInputSlotFlags::IsTouchPoint);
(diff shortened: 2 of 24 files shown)


Code/Engine/Core/System/Implementation/android/InputDevice_android.inl
```diff
@@ -0,0 +1,193 @@
...
+  const ezUInt32 uiAction = (ezUInt32)AMotionEvent_getAction(pEvent);
+  const ezInt32 iKeyCode = AKeyEvent_getKeyCode(pEvent);
+  const ezInt32 iButtonState = AMotionEvent_getButtonState(pEvent);
+  (void)iKeyCode;
+  (void)iButtonState;
+  DEBUG_LOG("Android INPUT: iEventType: {}, iEventSource: {}, uiAction: {}, iKeyCode: {}, iButtonState: {}", iEventType,
+    iEventSource, uiAction, iKeyCode, iButtonState);
+
+  ezHybridArray<ezScreenInfo, 2> screens;
+  if (ezScreen::EnumerateScreens(screens).Failed() || screens.IsEmpty())
```
jankrassnigg: shouldn't enumerating screens only be done once ?

Code/Engine/Core/System/Implementation/android/InputDevice_android.inl
```diff
@@ -0,0 +1,193 @@
...
+
+ezInt32 ezStandardInputDevice::AndroidHandleInput(AInputEvent* pEvent)
+{
+  // #TODO_ANDROID Only touchscreen input is implemented right now.
+  const ezInt32 iEventType = AInputEvent_getType(pEvent);
+  const ezInt32 iEventSource = AInputEvent_getSource(pEvent);
+  const ezUInt32 uiAction = (ezUInt32)AMotionEvent_getAction(pEvent);
+  const ezInt32 iKeyCode = AKeyEvent_getKeyCode(pEvent);
+  const ezInt32 iButtonState = AMotionEvent_getButtonState(pEvent);
+  (void)iKeyCode;
```
jankrassnigg: EZ_IGNORE_UNUSED ?

Code/Engine/Core/System/Implementation/android/InputDevice_android.inl
```diff
@@ -0,0 +1,193 @@
...
+        // Finger moved (we always update that at the top).
+        return 1;
+      case AMOTION_EVENT_ACTION_UP:
+      case AMOTION_EVENT_ACTION_POINTER_UP:
+      case AMOTION_EVENT_ACTION_CANCEL:
+      case AMOTION_EVENT_ACTION_OUTSIDE:
+        m_InputSlotValues[ezInputManager::GetInputSlotTouchPoint(id)] = 0;
+        DEBUG_LOG("Finger UP: {}", id);
+        return 1;
+      case AMOTION_EVENT_ACTION_SCROLL:
```
jankrassnigg, line 207: We have a mouse wheel event, that could be used for the scrolling.

Code/Engine/Foundation/Basics/Platform/Android/AndroidUtils.h
```diff
@@ -11,6 +12,15 @@ struct _JNIEnv;
 using JNIEnv = _JNIEnv;
 class _jobject;
 using jobject = _jobject*;
+struct AInputEvent;
+
+/// \brief Event fired by ezAndroidUtils::s_InputEvent.
+/// Event listeners should inspect m_pEvent and set m_iReturn to 1 if they handled the event.
```
jankrassnigg: Does m_iReturn have any other meaning? If it's just about handled vs. not-handled, then it should be a bool.

Code/Samples/ShaderExplorer/ShaderExplorer.cpp
```diff
@@ -361,8 +394,13 @@ void ezShaderExplorerApp::AfterCoreSystemsStartup()
 
 void ezShaderExplorerApp::BeforeHighLevelSystemsShutdown()
 {
+  ezTelemetry::CloseConnection();
```
jankrassnigg, line 398: why is it necessary to call CloseConnection here manually? this should automatically be taken care of elsewhere

does this function call it's base implementation (BeforeHighLevelSystemsShutdown()) ?

Code/Samples/ShaderExplorer/ShaderExplorer.cpp
```diff
@@ -244,10 +249,28 @@ void ezShaderExplorerApp::AfterCoreSystemsStartup()
   ezGALDeviceFactory::GetShaderModelAndCompiler(sRendererName, szShaderModel, szShaderCompiler);
 
   ezShaderManager::Configure(szShaderModel, true);
-  EZ_VERIFY(ezPlugin::LoadPlugin(szShaderCompiler).Succeeded(), "Shader compiler '{}' plugin not found", szShaderCompiler);
+  ezPlugin::LoadPlugin(szShaderCompiler).IgnoreResult();
```
jankrassnigg, line 253: LoadOptionalPlugin ?

Code/Samples/ShaderExplorer/ShaderExplorer.cpp
```diff
@@ -230,6 +234,7 @@ void ezShaderExplorerApp::AfterCoreSystemsStartup()
   ezGlobalLog::AddLogWriter(ezLogWriter::Console::LogMessageHandler);
   ezGlobalLog::AddLogWriter(ezLogWriter::VisualStudio::LogMessageHandler);
 
+  ezTelemetry::CreateServer();
```
jankrassnigg, line 238: This should already be done by the application base class somewhere.

Code/Engine/Core/Input/Implementation/InputManager.cpp
```diff
@@ -208,6 +209,7 @@ void ezInputManager::ResetInputSlotValues()
   // this is crucial for accumulating the new values and for resetting the input state later
   for (ezInputSlotsMap::Iterator it = GetInternals().s_InputSlots.GetIterator(); it.IsValid(); it.Next())
   {
+    it.Value().m_fValueOld = it.Value().m_fValue;
```
jankrassnigg, line 212: can't think of a better solution atm, at least not an easy one

jankrassnigg (approved):

Code/Engine/Foundation/Basics/Platform/Android/AndroidUtils.h
Sanakan8472: Looked at the docs "Return 1 if you have handled the event, 0 for any default dispatching.". So yes, should be fine to change it to just a bool.

Code/Samples/ShaderExplorer/ShaderExplorer.cpp
Sanakan8472: So the `ezShaderExplorerApp` derives from the `ezApplication` class and not `ezGameApplicationBase` which would have called the function. The sample is pretty minimalistic which is great for porting stuff. Deriving from `ezGameApplication` would require solving many other problems first before you have a working app. I doubt Android is currently able to manage that.

Code/Engine/Core/System/Implementation/android/InputDevice_android.inl
Sanakan8472: Ah I was just too lazy to implement onResize logic to detect when the screen size changes, adding that now.

Code/Engine/Core/System/Implementation/android/InputDevice_android.inl
Sanakan8472: Well, I implemented it now, but it looks like it's actually only used for an actual wheel on a physical mouse, not finger scrolling. But that's half a mouse already implemented right?🙂

Code/Samples/ShaderExplorer/ShaderExplorer.cpp
Sanakan8472: Changed it to `ezPlugin::LoadPlugin("ezInspectorPlugin", ezPluginLoadFlags::PluginIsOptional).IgnoreResult();`
Still need to discard the result thought.

Code/Engine/Core/Input/Implementation/InputManager.cpp
Sanakan8472: Ok let's leave it like that for now.

Code/Samples/ShaderExplorer/ShaderExplorer.cpp
Sanakan8472: See comment on the other ezTelemetry function.

PR #1249, opened 2024-04-01
Adjusted modifiers for gizmos to prevent unwanted input

jankrassnigg:
With the current way the gizmos use modifiers (shift, alt, ctrl), there is often unwanted input when you accidentally click next to the gizmo instead of on it.

Pressing ALT+LMB will orbit around the currently framed object. ALT+drag on a gizmo disables snapping. Thus ALT+ accidentally not hitting the gizmo will move the camera to a completely different position, which is really annoying.

Similarly SHIFT+drag on the gizmos duplicates objects, and SHIFT+LMB drag moves the camera really fast (SHIFT makes it move faster). Thus accidentally not hitting the gizmo again makes the camera move really fast away.

This PR changes the behavior to reduce this problem.

Instead of ALT to toggle snapping, it is now using SHIFT. This is actually inspired by Unity, and thus more intuitive for many people. Since it is commonly used, using a different modifier than ALT prevents frequent collision with the orbit camera behavior.

Camera movement can only be triggered via LMB+drag, not when SHIFT is already down. Thus SHIFT+LMB-drag while accidentally not hitting a gizmo simply does nothing.

Instead of SHIFT+drag on gizmo to duplicate, use CTRL+drag. Accidentally not hitting the gizmo then has either no effect, or only toggles the selection, which is quickly restorable with a second click.

Also fixed that using the arrow keys to navigate was very jarring, due to a large timestep.

We also had UE's move-camera-with-gizmo feature implemented (CTRL+drag on gizmo), but I don't think this was really used, so I've deactivated those code paths.

### DIFF
--- a/Code/Editor/EditorFramework/EditTools/Implementation/StandardGizmoEditTools.cpp
+++ b/Code/Editor/EditorFramework/EditTools/Implementation/StandardGizmoEditTools.cpp
@@ -63,19 +63,21 @@ void ezTranslateGizmoEditTool::TransformationGizmoEventHandlerImpl(const ezGizmo
     case ezGizmoEvent::Type::BeginInteractions:
     {
       const bool bDuplicate =
-        QApplication::keyboardModifiers().testFlag(Qt::KeyboardModifier::ShiftModifier) && GetGizmoInterface()->CanDuplicateSelection();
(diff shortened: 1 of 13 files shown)


jankrassnigg:
@Sanakan8472 @C-Core maybe you could give this a try and let me know, whether you disagree with these changes.

PR #1253, opened 2024-04-04
Moved State Machine Asset transform to engine process

### LINKED ISSUE #1200: [BUG] StateMachineStates can't be added through game plugins
If you try to use a StateMachineState that is defined in a game plugin, the editor tries to create an instance of it, to serialize the data. This doesn't work, because the code is not loaded into the editor.

A different solution has to be used for the state serialization.

C-Core:
* The connection information is now stored in the connection object itself instead of in metadata. 
* This allowed removal of all special handling for the render pipeline transform.
* Moved state machine asset transform to engine process so it can instantiate states from custom plugins. Fixes #1200 
* Re-saved all assets to reflect recent format changes.


Code/EditorPlugins/Assets/EditorPluginAssets/StateMachineAsset/StateMachineGraph.cpp
```diff
@@ -175,13 +124,28 @@ const ezRTTI* ezStateMachineNodeManager::GetConnectionType() const
...
+  if (pCommandHistory == nullptr)
+    return;
+
+  if (e.m_EventType == ezDocumentObjectStructureEvent::Type::AfterObjectAdded)
   {
-    if (m_pInitialStateObject == nullptr && IsAnyState(e.m_pObject) == false)
+    if (GetInitialState() == nullptr && IsAnyState(e.m_pObject) == false)
     {
-      SetInitialState(e.m_pObject);
+      if (pCommandHistory->IsInTransaction() && e.m_pObject->GetTypeAccessor().GetValue(s_szIsInitialState) == false)
```
Sanakan8472: Nit: As `GetInitialState()` will itterate over all root nodes, it might make sense to test `pCommandHistory->IsInTransaction()` before that to skip the function call in the general case.

Code/Tools/Libs/ToolsFoundation/NodeObject/Implementation/DocumentNodeManager.cpp
```diff
@@ -427,67 +436,90 @@ void ezDocumentNodeManager::RestoreMetaDataAfterLoading(const ezAbstractObjectGr
...
       {
-        ezLog::Error("Unknown input pin '{}' on '{}'. The connection has been removed.", connectionMetaData.m_TargetPin, pTarget->GetType()->GetTypeName());
         RemoveObject(pObject);
         DestroyObject(pObject);
-        continue;
       }
+    }
+    else
+    {
+      DocumentNodeManager_ConnectionMetaData connectionMetaData;
```
Sanakan8472, line 500: When is this case called? If it's not a node and not a connection, is it a sub-object or a legacy connection? What will happen to sub-object in this function anyways?

Code/EditorPlugins/Assets/EnginePluginAssets/StateMachineAsset/StateMachineContext.cpp
```diff
@@ -0,0 +1,139 @@
...
+  auto AddState = [&](const ezStateMachineNode* pNode, const ezUuid& uuid)
+  {
+    const ezString& name = pNode->m_sName;
+    if (stateNames.Contains(name))
+    {
+      return ezStatus(ezFmt("A state named '{}' already exists. State names have to be unique.", name));
+    }
+    stateNames.Insert(name);
+
+    ezUniquePtr<ezStateMachineState> pState = ezUniquePtr<ezStateMachineState>(pNode->m_pType, nullptr);
```
Sanakan8472, line 63: Is this effectively creating an `ezUniquePtr` without ownership?

Sanakan8472 (approved):

Code/Tools/Libs/ToolsFoundation/NodeObject/Implementation/DocumentNodeManager.cpp
C-Core: I introduced a new class for connections so the IsConnection check above won't work. In this case we read the metadata and create a new object of the new class and set the properties from metadata. 
A sub-object is not a node, nor a connection nor will it have a valid DocumentNodeManager_ConnectionMetaData, so nothing happens in this case.

Code/EditorPlugins/Assets/EnginePluginAssets/StateMachineAsset/StateMachineContext.cpp
C-Core: Yes, in this case the unique ptr acts like a raw pointer. I'm not super happy with this workaround but the state machine desc needs a unique ptr, which makes sense for the runtime usecase. So either I would need a different desc just for the export code or I would need to clone the state machine state/transition objects here just to get an unique ptr to them.

Code/EditorPlugins/Assets/EditorPluginAssets/StateMachineAsset/StateMachineGraph.cpp
C-Core: Done

PR #1256, opened 2024-04-08
Add WDStudios Implementation of Live++

JailbreakPapa:
This PR adds WD Studios Modification to add [Live++](https://liveplusplus.tech) to ezEngine.

NOTES:
 - Live++ Only works on Windows, Xbox & PS5. this commit only supports Windows.
 - Live++ **IS Paid**, but a 30-Day Trial can be used within the Broker Application.


JailbreakPapa:
Im currently going to put this PR as a Draft due to a bug found in Live++ that blocks Hot-Reloading with the Editor launched.

jankrassnigg:
Hey @JailbreakPapa thanks for the contribution, this is a very interesting addition.
I didn't have much time yet, so only took a quick glance, the CMake and C++ code looks fine.

However, I don't think we can check in those binaries that way. First of all, it's quite a big amount of data. We are very contiuous about keeping the overall repository size as small as possible (granted, not as successfully as we'd like), mainly because GitHub has no defined guidelines around allowed repo sizes, and we really don't want to get a letter from them that we have exceeded their quota. And with binary files you always have the problem, that you need to update them every once in a while, which potentially quickly adds up in size. Also that would require us to maintain it, which we can't, because we don't even have a license.

So for a feature that is not essential (like PhysX was or Fmod is), I would not want to check those files in.

However, I assume it isn't hard to get those (up-to-date) binaries. So what we can do is to check in the code changes and, [in the documentation](https://ezengine.net/pages/docs/appendix/third-party-code.html), add something to tell people where to get those files and where to put them, etc. For people who do have a license, or want to give it a try, I don't think that's a big issue.

If you could write up the rough steps, I can help out with setting up the documentation page.
How does that sound to you?

JailbreakPapa:
> Hey @JailbreakPapa thanks for the contribution, this is a very interesting addition. I didn't have much time yet, so only took a quick glance, the CMake and C++ code looks fine.
> 
> However, I don't think we can check in those binaries that way. First of all, it's quite a big amount of data. We are very contiuous about keeping the overall repository size as small as possible (granted, not as successfully as we'd like), mainly because GitHub has no defined guidelines around allowed repo sizes, and we really don't want to get a letter from them that we have exceeded their quota. And with binary files you always have the problem, that you need to update them every once in a while, which potentially quickly adds up in size. Also that would require us to maintain it, which we can't, because we don't even have a license.
> 
> So for a feature that is not essential (like PhysX was or Fmod is), I would not want to check those files in.
> 
> However, I assume it isn't hard to get those (up-to-date) binaries. So what we can do is to check in the code changes and, [in the documentation](https://ezengine.net/pages/docs/appendix/third-party-code.html), add something to tell people where to get those files and where to put them, etc. For people who do have a license, or want to give it a try, I don't think that's a big issue.
> 
> If you could write up the rough steps, I can help out with setting up the documentation page. How does that sound to you?

Hey @jankrassnigg,

Thanks for checking out the PR. Yeah, we can let users install it directly from the [Live++ Website](https://liveplusplus.tech/) and that solves the Storage Issue. 

I'm also thinking that It can just be installed like Qt6 Binaries are installed, and it's through a zip file. 

For the documentation, I don't mind doing a PR to the docs repo as well, but heres a basic overview:

# Basic Explantation of Live++

Live++ is a Hot-Reloading Middleware for C++ Applications. Whether you're working on new features, trying out gameplay mechanics, optimizing code, or fixing bugs, Live++ will massively cut down your iteration times.

Live++ only currently works with Windows, but Linux Support is coming.

## How to use Live++

Head to https://liveplusplus.tech/ and Scroll down to the free-trial section if you already dont have a license. 

You should get a ZIP File that includes the Live++ API & Tools. Take the ZIP File, and extract it to: Code/ThirdParty/LivePP. 

Now Configure & Generate ezEngine and Build.

When the engine opens up you should see the Live++ Broker pop up:

image

Verify in the **Processes** Section that ezEditor.exe, and ezEditorEngineProcess.exe are shown in the Live++ Broker.

Edit a Source file, Save it, and Press Ctrl + Alt + F11 at the same time.

Wait for a second and you should see this modal pop-up:

image

After that modal dissapears, another one should also pop-up, stating that Hot-Reloading was Successful:

image

PlasmaDev5:
Out of curiosity how does it stackup/compare to the existing "hot reload" behavior used for game plugins

JailbreakPapa:
> Out of curiosity how does it stackup/compare to the existing "hot reload" behavior used for game plugins

Hmm, from testing it does the same behavior even within the Player application, i guess the only thing that you dont have to do is restart the engine editor process. but they should work fine without conflicting with each-other.

JailbreakPapa:
> > Out of curiosity how does it stackup/compare to the existing "hot reload" behavior used for game plugins
> 
> Hmm, from testing it does the same behavior even within the Player application, I guess the only thing that you don't have to do is restart the engine editor process. but they should work fine without conflicting with each-other.

We originally added it for working with the core engine without considering GamePlugins, so they were a afterthought.

Code/ThirdParty/LivePP/init.cmake
```diff
@@ -0,0 +1,31 @@
+# LIVE++ Utils
+# NOTES: To not use github storage, you can download the Live++ SDK from the official website and put it in the same directory as this file.
+
+if(EZ_CMAKE_PLATFORM_WINDOWS)
+    set(EZ_SUPPORT_LIVEPP ON CACHE BOOL "Enable Live++ support for Ez")
```
jankrassnigg: Should be named EZ_3RDPARTY_LIVEPP_SUPPORT to be in line with other 3rd party integrations.

Also should be off by default on all platforms, but can be automatically set to ON, when the necessary binaries are copied here by the user.

jankrassnigg (changes requested):
Note that the build has to run without the binaries checked in, which I suspect it won't do currently.

JailbreakPapa:
> Note that the build has to run without the binaries checked in, which I suspect it won't do currently.

~~Yeah, that means the AZ Pipelines will eventually fail. I'm not at my PC right now, but once I'm able to, I'll resolve some of these issues~~

Disregard that, it should work as long as the header files are there.

JailbreakPapa:
> > Note that the build has to run without the binaries checked in, which I suspect it won't do currently.
> 
> ~~Yeah, that means the AZ Pipelines will eventually fail. I'm not at my PC right now, but once I'm able to, I'll resolve some of these issues~~
> 
> Disregard that, it should work as long as the header files are there.

Aaaandddd it fails. I'll get to fixing these in a few hours.

JailbreakPapa:
Added the Live++ Headers back, and CMake should throw an error if the user has not placed the Live++ SDK in the directory.

This should fix the AZ Pipeline Failures (In-terms of building)

JailbreakPapa:
> Added the Live++ Headers back, and CMake should throw an error if the user has not placed the Live++ SDK in the directory.
> 
> This should fix the AZ Pipeline Failures (In-terms of building)

Actually, this still doesn't solve the problem of "Maintaining" updates with Live++. Their API Hardly ***EVER*** changes, but this still applies.

One way this can simply be solved, is if the user just enables the build flag, and adds the needed exe's & API, then builds.

JailbreakPapa:
Nice, seems that builds & tests are good. The change request should be resolved as it is fixed in the latest commit.

Code/ThirdParty/LivePP/init.cmake
JailbreakPapa: @jankrassnigg Can this be marked as resolved on your end since it is fixed in newer commits to the PR

Code/ThirdParty/LivePP/init.cmake
```diff
@@ -0,0 +1,35 @@
+# LIVE++ Utils
+# NOTES: To not use github storage, you can download the Live++ SDK from the official website and put it in the same directory as this file.
+
+if(EZ_CMAKE_PLATFORM_WINDOWS)
+    set(EZ_3RDPARTY_LIVEPP_SUPPORT OFF CACHE BOOL "Enable Live++ support for Ez")
```
jankrassnigg: in this code path it would be nice to check with CMake whether the Live++ binaries exist, which means that the user put them there themselves, and then enable EZ_3RDPARTY_LIVEPP_SUPPORT.

This can easily be done with CMake just with

```
if (EXISTS "${CURRENT_SOURCE_DIR}/something.lib")
```

You would probably need to set the CACHE variable with FORCE, to ensure it gets overridden, so that users who copy the libs in later still get the feature enabled.

JailbreakPapa: The preview of this review doesn't show it, but in the latest commit, I added a check for if the Live++ broker was in the directory. 

Would you want me to check for the DLL files?

JailbreakPapa: The preview of this review doesn't show it, but in the latest commit, I added a check for if the Live++ broker was in the directory. 

Would you want me to check for the DLL files?

jankrassnigg: Oh I did not see that. But then you need to change this line to ON, because currently it is always switched to OFF.

jankrassnigg: Also I would test for EZ_CMAKE_PLATFORM_WINDOWS_DESKTOP, to exclude UWP.

Code/ThirdParty/LivePP/init.cmake
```diff
@@ -2,21 +2,25 @@
...
 else()
-    set(EZ_SUPPORT_LIVEPP OFF CACHE BOOL "Enable Live++ support for Ez")
+    set(EZ_3RDPARTY_LIVEPP_SUPPORT OFF CACHE BOOL "Enable Live++ support for Ez")
 endif()
 
 # #####################################
 # ## ez_requires_livepp()
 # #####################################
 macro(ez_requires_livepp)
     ez_requires_one_of(EZ_CMAKE_PLATFORM_WINDOWS)
```
jankrassnigg: ez_requires_one_of(EZ_CMAKE_PLATFORM_WINDOWS)

This should be redundant, because EZ_3RDPARTY_LIVEPP_SUPPORT is already tied to the platform.

Code/ThirdParty/LivePP/init.cmake
JailbreakPapa: Resolved.

Code/ThirdParty/LivePP/init.cmake
JailbreakPapa: Fixed in latest commit.

JailbreakPapa:
Seems to have failed due to the asset processor timing out.

JailbreakPapa:
Nice, everything passed. since the change requested is outdated & resolved, I think this PR is done 🤞

jankrassnigg:
> Nice, everything passed. since the change requested is outdated & resolved, I think this PR is done 🤞

I haven't had time to give this a try yet, and I'd like the rest of the team to also get an opportunity to try it out before merging it in. So let's keep this PR open at least until the end of the weekend.

JailbreakPapa:
> > Nice, everything passed. since the change requested is outdated & resolved, I think this PR is done 🤞
> 
> 
> 
> I haven't had time to give this a try yet, and I'd like the rest of the team to also get an opportunity to try it out before merging it in. So let's keep this PR open at least until the end of the weekend.

No problem.

JailbreakPapa:
Actually, there are a few ways that I can improve working with Live++ [Hot-Restart](https://liveplusplus.tech/docs/documentation.html#tools_hot_restart)

Especially with when the Editor restarts, patches could most likely not get applied in situations like the editor getting closed/restarted by Visual Studio, a Plugin needs to be enabled/disabled, etc...

jankrassnigg (approved):

jankrassnigg:
@JailbreakPapa unless you still want to change something, I would check this in now, even though so far I had no time to give it a try.

Also, would be great if could add a page to the docs repository about this feature.

JailbreakPapa:
> @JailbreakPapa unless you still want to change something, I would check this in now, even though so far I had no time to give it a try.
> 
> 
> 
> Also, would be great if could add a page to the docs repository about this feature.

This can be merged. I can also set up another PR for the docs on that repo

JailbreakPapa:
Seemed to fail on dev due to some storage error.

PR #1257, opened 2024-04-08
Add WDStudios Tracy intergration to ezEngine

JailbreakPapa:
This PR Adds [Tracy](https://github.com/wolfpld/tracy) Support to ezEngine.

image


Code/Engine/RendererCore/RenderWorld/Implementation/RenderWorld.cpp
```diff
@@ -580,10 +580,12 @@ void ezRenderWorld::Render(ezRenderContext* pRenderContext)
   }
 
   filteredRenderPipelines.Clear();
-
-  renderEvent.m_Type = ezRenderWorldRenderEvent::Type::EndRender;
-  EZ_PROFILE_SCOPE("EndRender");
-  s_RenderEvent.Broadcast(renderEvent);
+  /// NOTE: (Only Applies When Tracy is Enabled.)Tracy Seems to declare Timers in the same scope, so dual profile macros can throw: '__tracy_scoped_zone' : redefinition; multitple initalization, so we must scope the two events.
```
JailbreakPapa, line 583: Im unsure if we should scope this code here, but Tracy declares Timers in the same scope, and in turn, gives us a redefinition error.

Code/EnginePlugins/OpenXRPlugin/OpenXRSingleton.cpp
```diff
@@ -941,35 +940,36 @@ void ezOpenXR::EndFrame()
 
   if (!m_bRenderInProgress || !pSwapChain)
     return;
-
-  EZ_PROFILE_SCOPE("OpenXrEndFrame");
-  for (uint32_t i = 0; i < 2; i++)
+  /// NOTE: (Only Applies When Tracy is Enabled.)Tracy Seems to declare Timers in the same scope, so dual profile macros can throw: '__tracy_scoped_zone' : redefinition; multitple initalization, so we must scope the two events.
```
JailbreakPapa, line 943: See: https://github.com/ezEngine/ezEngine/pull/1257/files#r1556459342

jankrassnigg:
Thank you for that contribution. I use Tracy at work, where it has proven to be a really great tool, so happy to see this integrated into ez. I'm currently a bit short on time, but will try to review the PR soon.

Code/ThirdParty/tracy/LICENSE
```diff
@@ -0,0 +1,27 @@
+Tracy Profiler (https://github.com/wolfpld/tracy) is licensed under the
```
C-Core, line 1: Can you also add a "cgmanifest.json" file to specify which version and which commit hash of tracy this is exactly. For example look at Code/ThirdParty/Imgui. This makes it easier to update tracy in the future or investigate potential bugs etc.

Code/Engine/Foundation/Profiling/Profiling.h
```diff
@@ -164,25 +164,25 @@ class EZ_FOUNDATION_DLL ezProfilingSystem
...
+#  ifndef TRACY_ENABLE
 /// \brief Profiles the current scope using the given name.
 ///
 /// It is allowed to nest EZ_PROFILE_SCOPE, also with EZ_PROFILE_LIST_SCOPE. However EZ_PROFILE_SCOPE should start and end within the same list scope
 /// section.
 ///
 /// \note The name string must not be destroyed before the current scope ends.
 ///
-/// \sa ezProfilingScope
+/// \sa nsProfilingScope
```
C-Core: What is that ns prefix doing here?

Code/Engine/Foundation/Profiling/Profiling.h
```diff
@@ -193,17 +193,64 @@ class EZ_FOUNDATION_DLL ezProfilingSystem
...
+/// \brief Profiles the current scope using the given name.
+///
+/// It is allowed to nest EZ_PROFILE_SCOPE, also with EZ_PROFILE_LIST_SCOPE. However EZ_PROFILE_SCOPE should start and end within the same list scope
+/// section.
+///
+/// \note The name string must not be destroyed before the current scope ends.
+///
+/// \sa nsProfilingScope
+/// \sa EZ_PROFILE_LIST_SCOPE
+#    define EZ_PROFILE_SCOPE(szScopeName) \
```
C-Core: Is there no way to make use of the user given name?

ezCMakeConfig.cmake
```diff
@@ -19,3 +19,4 @@ set_property(GLOBAL PROPERTY EZ_CONFIG_QT_WINX64_URL "https://github.com/ezEngin
 set_property(GLOBAL PROPERTY EZ_CONFIG_VULKAN_SDK_LINUXX64_VERSION "1.3.275.0")
 set_property(GLOBAL PROPERTY EZ_CONFIG_VULKAN_SDK_LINUXX64_URL "https://sdk.lunarg.com/sdk/download/1.3.275.0/linux/vulkansdk-linux-x86_64-1.3.275.0.tar.xz")
 
+set(EZ_SUPPORT_TRACY ON CACHE BOOL "Support Tracy Profiler")
```
C-Core: Since tracy replaces the builtin profiling I don't think it should be on by default.

Code/Engine/Foundation/Profiling/Profiling.h
JailbreakPapa: Hmm, left over from porting it over to the Ez clone. That's my bad. I'm not at my pc currently, but I can tackle these issues soon.

Code/Engine/Foundation/Profiling/Profiling.h
JailbreakPapa: There is a way, using Tracy's ZoneScopedN macro, but this would force some profiling macros to fail, as it wouldn't be able to convert the types for Tracy.

JailbreakPapa: But I definitely agree this should be changed.

Code/Engine/RendererCore/RenderWorld/Implementation/RenderWorld.cpp
C-Core: Just looked a bit through the documentation and found this macro: ZoneNamedN(variableName, "Zone name"). This would allow us to use a custom variable name to prevent these scoping issues, like e.g. concat a fixed name with the line number like we do with the ez profiling and also pass it a custom name.

JailbreakPapa:
AZ Run failed due to some left over ns Prefixes from our fork of the engine I forgot to change, my apologies.

TheophilusE:
When I integrated Tracy, about a year ago. This was what I did for the API usage, not sure if things have changed now.
```cpp
#if XII_ENABLED(XII_USE_PROFILING) || defined(XII_DOCS)

#  if BUILDSYSTEM_ENABLE_TRACY_SUPPORT

XII_ALWAYS_INLINE xiiUInt32 ___tracyGetStringLength(const char* szString)
{
  return xiiStringUtils::GetStringElementCount(szString);
}

XII_ALWAYS_INLINE xiiUInt32 ___tracyGetStringLength(xiiStringView szString)
{
  return szString.GetElementCount();
}

XII_ALWAYS_INLINE xiiUInt32 ___tracyGetStringLength(const xiiStringBuilder& szString)
{
  return szString.GetElementCount();
}

#    include <Tracy/tracy/Tracy.h>
#    define TRACY_PROFILE_SCOPE_DYNAMIC(szScopeName) \
      ZoneScoped;                                    \
      ZoneName(szScopeName, ___tracyGetStringLength(szScopeName))
#  else
#    define TRACY_PROFILE_SCOPE_DYNAMIC(szScopeName)
#  endif

/// \brief Profiles the current scope using the given name.
///
/// It is allowed to nest XII_PROFILE_SCOPE, also with XII_PROFILE_LIST_SCOPE. However XII_PROFILE_SCOPE should start and end within the same list scope
/// section.
///
/// \note The name string must not be destroyed before the current scope ends.
///
/// \sa xiiProfilingScope
/// \sa XII_PROFILE_LIST_SCOPE
#  define XII_PROFILE_SCOPE(szScopeName)                                                                                  \
    xiiProfilingScope XII_CONCAT(_xiiProfilingScope, XII_SOURCE_LINE)(szScopeName, XII_SOURCE_FUNCTION, xiiTime::Zero()); \
    TRACY_PROFILE_SCOPE_DYNAMIC(szScopeName)
```

TheophilusE:
In addition, the CMakeLists is quite bare, there are things that might want to be disabled, if necessary. In any case, the source CMake configuration may be modified like so.

```cmake
xii_cmake_init()

xii_requires(XII_BUILD_TRACY)

# Get the name of this folder as the project name
get_filename_component(PROJECT_NAME ${CMAKE_CURRENT_SOURCE_DIR} NAME_WE)

find_package(Threads REQUIRED)

xii_create_target(LIBRARY TracyClient NO_XII_PREFIX NO_UNITY MANUAL_SOURCE_FILES TracyClient.cpp)

target_compile_features(TracyClient PUBLIC cxx_std_11)

target_include_directories(TracyClient SYSTEM PUBLIC
    $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}>
    $<INSTALL_INTERFACE:include>)

target_link_libraries(TracyClient
    PUBLIC
    Threads::Threads
    ${CMAKE_DL_LIBS}
)

# Public dependency on some libraries required when using MinGw
if(WIN32 AND ${CMAKE_CXX_COMPILER_ID} MATCHES "GNU")
    target_link_libraries(TracyClient PUBLIC ws2_32 dbghelp)
endif()

if(CMAKE_SYSTEM_NAME MATCHES "FreeBSD")
    find_library(EXECINFO_LIBRARY NAMES execinfo REQUIRED)
    target_link_libraries(TracyClient PUBLIC ${EXECINFO_LIBRARY})
endif()

add_library(Tracy::TracyClient ALIAS TracyClient)

macro(set_option option help value)
    option(${option} ${help} ${value})
    mark_as_advanced(${option})
    if(${option})
        message(STATUS "${option}: ON")
        target_compile_definitions(TracyClient PUBLIC ${option})
    else()
        message(STATUS "${option}: OFF")
    endif()
endmacro()

set_option(TRACY_ENABLE "Enable profiling" ON)
set_option(TRACY_ON_DEMAND "On-demand profiling" ON)
set_option(TRACY_CALLSTACK "Enfore callstack collection for tracy regions" OFF)
set_option(TRACY_NO_CALLSTACK "Disable all callstack related functionality" OFF)
set_option(TRACY_NO_CALLSTACK_INLINES "Disables the inline functions in callstacks" OFF)
set_option(TRACY_ONLY_LOCALHOST "Only listen on the localhost interface" OFF)
set_option(TRACY_NO_BROADCAST "Disable client discovery by broadcast to local network" OFF)
set_option(TRACY_ONLY_IPV4 "Tracy will only accept connections on IPv4 addresses (disable IPv6)" OFF)
set_option(TRACY_NO_CODE_TRANSFER "Disable collection of source code" OFF)
set_option(TRACY_NO_CONTEXT_SWITCH "Disable capture of context switches" OFF)
set_option(TRACY_NO_EXIT "Client executable does not exit until all profile data is sent to server" OFF)
set_option(TRACY_NO_SAMPLING "Disable call stack sampling" OFF)
set_option(TRACY_NO_VERIFY "Disable zone validation for C API" OFF)
set_option(TRACY_NO_VSYNC_CAPTURE "Disable capture of hardware Vsync events" OFF)
set_option(TRACY_NO_FRAME_IMAGE  "Disable the frame image support and its thread" OFF)
set_option(TRACY_NO_SYSTEM_TRACING  "Disable systrace sampling" OFF)
set_option(TRACY_DELAYED_INIT "Enable delayed initialization of the library (init on first call)" OFF)
set_option(TRACY_MANUAL_LIFETIME "Enable the manual lifetime management of the profile" OFF)
set_option(TRACY_FIBERS "Enable fibers support" OFF)
set_option(TRACY_NO_CRASH_HANDLER "Disable crash handling" OFF)
set_option(TRACY_TIMER_FALLBACK "Use lower resolution timers" OFF)

if(XII_COMPILE_ENGINE_AS_DLL)
    target_compile_definitions(TracyClient PRIVATE TRACY_EXPORTS)
    target_compile_definitions(TracyClient PUBLIC TRACY_IMPORTS)
endif()
```

JailbreakPapa:
> When I integrated Tracy, about a year ago. This was what I did for the API usage, not sure if things have changed now.
> 
> ```cpp
> 
> #if XII_ENABLED(XII_USE_PROFILING) || defined(XII_DOCS)
> 
> 
> 
> #  if BUILDSYSTEM_ENABLE_TRACY_SUPPORT
> 
> 
> 
> XII_ALWAYS_INLINE xiiUInt32 ___tracyGetStringLength(const char* szString)
> 
> {
> 
>   return xiiStringUtils::GetStringElementCount(szString);
> 
> }
> 
> 
> 
> XII_ALWAYS_INLINE xiiUInt32 ___tracyGetStringLength(xiiStringView szString)
> 
> {
> 
>   return szString.GetElementCount();
> 
> }
> 
> 
> 
> XII_ALWAYS_INLINE xiiUInt32 ___tracyGetStringLength(const xiiStringBuilder& szString)
> 
> {
> 
>   return szString.GetElementCount();
> 
> }
> 
> 
> 
> #    include <Tracy/tracy/Tracy.h>
> 
> #    define TRACY_PROFILE_SCOPE_DYNAMIC(szScopeName) \
> 
>       ZoneScoped;                                    \
> 
>       ZoneName(szScopeName, ___tracyGetStringLength(szScopeName))
> 
> #  else
> 
> #    define TRACY_PROFILE_SCOPE_DYNAMIC(szScopeName)
> 
> #  endif
> 
> 
> 
> /// \brief Profiles the current scope using the given name.
> 
> ///
> 
> /// It is allowed to nest XII_PROFILE_SCOPE, also with XII_PROFILE_LIST_SCOPE. However XII_PROFILE_SCOPE should start and end within the same list scope
> 
> /// section.
> 
> ///
> 
> /// \note The name string must not be destroyed before the current scope ends.
> 
> ///
> 
> /// \sa xiiProfilingScope
> 
> /// \sa XII_PROFILE_LIST_SCOPE
> 
> #  define XII_PROFILE_SCOPE(szScopeName)                                                                                  \
> 
>     xiiProfilingScope XII_CONCAT(_xiiProfilingScope, XII_SOURCE_LINE)(szScopeName, XII_SOURCE_FUNCTION, xiiTime::Zero()); \
> 
>     TRACY_PROFILE_SCOPE_DYNAMIC(szScopeName)
> 
> ```

Nice. This solves the szScopeName not being used.

Code/ThirdParty/tracy/LICENSE
JailbreakPapa: Added.

Code/Engine/Foundation/Profiling/Profiling.h
JailbreakPapa: Fixed

JailbreakPapa:
@jankrassnigg Is it possible if the AZ Pipelines can be ran again?

JailbreakPapa:
> @jankrassnigg Is it possible if the AZ Pipelines can be ran again? 

Hmm looks like I messed up something with the Tracy defines in Profiling that made the build fail. Going to fix.

JailbreakPapa:
Hopefully, this will be the [final fix](https://github.com/ezEngine/ezEngine/pull/1257/commits/847533f437759805f06e24416a7cfc53be6367ed) so Pipelines can be built properly.

JailbreakPapa:
Well, this is embarrassing.... its getting really late here, so ill continue fixing this tomorrow. Apologies for this.

jankrassnigg:
> Well, this is embarrassing.... its getting really late here, so ill continue fixing this tomorrow. Apologies for this.

And here it's pretty early :D No worries, this isn't urgent.

JailbreakPapa:
Just one more fix.....

JailbreakPapa:
Alright, last last one that should fix, seems to run fine on our Jenkins pipelines. If this fails I'll check back later

JailbreakPapa:
Alright, at least build the define issues are done. 

Give me a second to fix the header check issues, and the Threads issue.

JailbreakPapa:
Pushed the HeaderCheck & Editor Test fixes, but linux might still fail due to a strange error with tracy: 
image

JailbreakPapa:
Really not sure why Linux is failing.

jankrassnigg:
Maybe fix the Clang, Mac and UWP build first? Might fix the Linux build as well, or change the build error to something easier to understand.

jankrassnigg:
It's possible that this is an error in the compiler / STL of the version that we use on Linux. It looks like the internal libraries have a syntax error. In that case it may go away with a compiler upgrade, but we probably won't do that soon. So you can also just disable Tracy on Linux in general.

JailbreakPapa:
> It's possible that this is an error in the compiler / STL of the version that we use on Linux. It looks like the internal libraries have a syntax error. In that case it may go away with a compiler upgrade, but we probably won't do that soon. So you can also just disable Tracy on Linux in general.

alright then, I'll fix the define issue, and see if that was the underlying error, if it still fails, ill disable Linux for Tracy.

Code/BuildSystem/CMake/CMakeUtils/ezUtilsTarget.cmake
```diff
@@ -155,4 +155,20 @@ macro(ez_create_target TYPE TARGET_NAME)
...
+				TRACY_NO_INVARIANT_CHECK
+				TRACY_NO_FRAME_IMAGE
+			)
+
+			if(NOT ${TARGET_NAME} STREQUAL "TracyClient" OR "Tracy::TracyClient")
+				target_link_libraries(${TARGET_NAME} PRIVATE TracyClient)
+			endif()
+		endif()
+	endif()
+
```
jankrassnigg: This shouldn't be here.

The compile definitions should be set on the Tracy library as PUBLIC, then they will automatically propagate to linked targets.

ezFoundation should be the only library that links against tracy, and it should link PUBLIC against it, so that the PUBLIC definitions of tracy propagate to its users.

The EZ_SUPPORT_TRACY option should be set up in an init.cmake file in the tracy lib.

jankrassnigg (changes requested):

Code/Engine/Foundation/Application/Implementation/MainLoop.cpp
```diff
@@ -26,6 +26,7 @@ void ezRun_MainLoop(ezApplication* pApplicationInstance)
 {
   while (pApplicationInstance->Run() == ezApplication::Execution::Continue)
   {
+    EZ_PROFILER_END_FRAME;
```
jankrassnigg: @C-Core don't we already have some end frame marker? I thought we need this for the GPU stuff.

Code/Engine/Core/World/Implementation/World.cpp
```diff
@@ -1070,7 +1070,7 @@ void ezWorld::UpdateSynchronous(const ezArrayPtr<ezInternal::WorldData::Register
       continue;
 
     {
-      EZ_PROFILE_SCOPE(updateFunction.m_sFunctionName);
+      EZ_PROFILE_SCOPE(updateFunction.m_sFunctionName.GetData());
```
jankrassnigg: There should be no need to explicitly write the GetData() call here. If it doesn't compile, the tracy macros need to be adjusted.

Code/Editor/EditorFramework/IPC/EngineProcessConnection.cpp
```diff
@@ -288,7 +288,10 @@ bool ezEditorEngineProcessConnection::SendMessage(ezProcessMessage* pMessage)
 
 ezResult ezEditorEngineProcessConnection::WaitForMessage(const ezRTTI* pMessageType, ezTime timeout, ezProcessCommunicationChannel::WaitForMessageCallback* pCallback)
 {
-  EZ_PROFILE_SCOPE(pMessageType->GetTypeName());
+  // NOTE: When Tracy is enabled, there is no conversion for ezStringView to const char*, so this is a workaround.
+  ezStringBuilder sb;
+  pMessageType->GetTypeName().GetData(sb);
+  EZ_PROFILE_SCOPE(sb);
```
jankrassnigg: This has to work without modification here, otherwise it is inconvenient and wastes performance for unnecessary string copies.

I'll mark up all cases that I see that need to be fixed.

Code/Engine/GameEngine/GameApplication/Implementation/GameApplication.cpp
```diff
@@ -141,7 +141,10 @@ void ezGameApplication::Run_Present()
 
   for (ezActor* pActor : allActors)
   {
-    EZ_PROFILE_SCOPE(pActor->GetName());
+    // NOTE: When Tracy is enabled, there is no conversion for ezStringView to const char*, so this is a workaround.
```
jankrassnigg: revert this

Code/EnginePlugins/TypeScriptPlugin/Components/TypeScriptComponent.cpp
```diff
@@ -294,7 +294,9 @@ void ezTypeScriptComponent::Update(ezTypeScriptBinding& binding)
   if (m_LastUpdate + m_UpdateInterval > tNow)
     return;
 
-  EZ_PROFILE_SCOPE(GetOwner()->GetName());
+  ezStringBuilder sb;
```
jankrassnigg: revert this

Code/ThirdParty/tracy/CMakeLists.txt
```diff
@@ -0,0 +1,82 @@
+ez_cmake_init()
+
+ez_requires(EZ_SUPPORT_TRACY)
```
jankrassnigg: The variable should be called EZ_3RDPARTY_TRACY_SUPPORT

Code/BuildSystem/CMake/CMakeUtils/ezUtilsTarget.cmake
JailbreakPapa: Resolved in latest commit.

ezCMakeConfig.cmake
```diff
@@ -19,3 +19,4 @@ set_property(GLOBAL PROPERTY EZ_CONFIG_QT_WINX64_URL "https://github.com/ezEngin
 set_property(GLOBAL PROPERTY EZ_CONFIG_VULKAN_SDK_LINUXX64_VERSION "1.3.275.0")
 set_property(GLOBAL PROPERTY EZ_CONFIG_VULKAN_SDK_LINUXX64_URL "https://sdk.lunarg.com/sdk/download/1.3.275.0/linux/vulkansdk-linux-x86_64-1.3.275.0.tar.xz")
 
+set(EZ_SUPPORT_TRACY OFF CACHE BOOL "Support Tracy Profiler")
```
jankrassnigg: This should be in an init.cmake file in the tracy lib folder.

Code/Tools/Libs/ToolsFoundation/Document/Implementation/DocumentManager.cpp
```diff
@@ -237,7 +237,9 @@ ezStatus ezDocumentManager::CreateOrOpenDocument(bool bCreate, ezStringView sDoc
       EZ_ASSERT_DEV(DocumentTypes[i]->m_bCanCreate, "This document manager cannot create the document type '{0}'", sDocumentTypeName);
 
       {
-        EZ_PROFILE_SCOPE(sDocumentTypeName);
+        ezStringBuilder szDocumentTypeName;
```
jankrassnigg: revert this

jankrassnigg (changes requested):

Code/Editor/EditorFramework/IPC/EngineProcessConnection.cpp
JailbreakPapa: Ill make a converter for ez string types within Profiler_Tracy

Code/Engine/RendererCore/RenderWorld/Implementation/RenderWorld.cpp
JailbreakPapa: Unresolving to add conversions for ez string types

jankrassnigg:
I've added memory tracking. It doesn't record callstacks though, not sure why. But not important at the moment.

Code/Engine/Foundation/Application/Implementation/MainLoop.cpp
C-Core: We have end frame in GALDevice that is used for the GPU profiling scopes, but it wouldn't make sense to add the tracy end frame marker there. We could use the ezProfilingSystem::StartNewFrame function though and add the tracy macro in there. The downside would be that tracy would only work in game applications, this way it works with all ez applications.

Code/Engine/Foundation/Profiling/Profiling_Tracy.h
```diff
@@ -0,0 +1,98 @@
...
+{
+  return szString.GetData();
+}
+EZ_ALWAYS_INLINE const char* __convertezStringToConstChar(const ezHashedString& szString)
+{
+  return szString.GetData();
+}
+EZ_ALWAYS_INLINE const char* __convertezStringToConstChar(const ezStringView& szString)
+{
+  ezStringBuilder temp_str;
```
jankrassnigg: This is a typical use after free, you can't return the pointer to the stringbuilder data, when that stringbuilder is local to the function.

Is there no way to give tracy a string view?
Alternatively you would need something else to copy the strings to for a longer time.

@C-Core do you have any idea? Maybe use the frame allocator?

Code/Engine/Foundation/Profiling/Profiling_Tracy.h
```diff
@@ -21,11 +22,36 @@ EZ_ALWAYS_INLINE ezUInt32 ___tracyGetStringLength(const ezStringBuilder& szStrin
...
+  szString.GetData(temp_str);
+  return temp_str.GetData();
```
jankrassnigg: This is a typical use after free, you can't return the pointer to the stringbuilder data, when that stringbuilder is local to the function.

Is there no way to give tracy a string view?
Alternatively you would need something else to copy the strings to for a longer time

jankrassnigg (changes requested):

Code/Engine/Foundation/Profiling/Profiling_Tracy.h
jankrassnigg: Just saw how this was used and fixed it, because Tracy already takes strings as startptr+size.

Code/Engine/Foundation/Profiling/Profiling_Tracy.h
```diff
@@ -0,0 +1,98 @@
...
+///
+/// It is allowed to nest EZ_PROFILE_SCOPE, also with EZ_PROFILE_LIST_SCOPE. However EZ_PROFILE_SCOPE should start and end within the same list scope
+/// section.
+///
+/// \note The name string must not be destroyed before the current scope ends.
+///
+/// \sa ezProfilingScope
+/// \sa EZ_PROFILE_LIST_SCOPE
+#    define EZ_PROFILE_SCOPE(szScopeName)                                                                                 \
+      ezProfilingScope EZ_CONCAT(_ezProfilingScope, EZ_SOURCE_LINE)(szScopeName, EZ_SOURCE_FUNCTION, ezTime::MakeZero()); \
```
jankrassnigg: This is a bit inconsistent.

For EZ_PROFILE_SCOPE we now call BOTH the ez profiling system AND tracy.
For several other macros below, the implementation is disabled.
Also Tracy is currently globally enabled.

The question is, what do we want. Is Tracy supposed to fully replace the ez system, or should it work in addition to it.

If it is supposed to replace it, then this macro shouldn't call the ez stuff, but then the default should also be to disable tracy in CMake.

If it should be in addition to ez, then all macros here should forward to the ez scopes.
In that case it could stay enabled by default in cmake.

jankrassnigg: I've now changed it so that Tracy overrides the EZ macros, but keeps all EZ profiling functionality as well.
In theory one can now also use EZ_TRACY_PROFILE_SCOPE directly, in case one wants to explicitly ignore the EZ profiling system.
I've also removed the duplicate code comments on the macros, they should not be needed.

@C-Core what's your opinion about this?

JailbreakPapa:
Seems to have failed due to the Build Machine not supporting some cpu instruction for Tracy I believe.

jankrassnigg:
> Seems to have failed due to the Build Machine not supporting some cpu instruction for Tracy I believe.

Ugh, maybe it ran on some older hardware. Don't think this is important to fix in any way, so we could just set the option that disables the check.

jankrassnigg (approved):

PR #1260, opened 2024-04-13
Fixed ezTypedComponentHandle type safety with TryGetComponent

### LINKED ISSUE #1259: [BUG] ezTypedComponentHandle isn't typesafe
Consider this code:
```cpp
ezTypedComponentHandle<ezMeshComponent> hMeshComp;

ezLocalBlackboardComponent* pBlackboardComp = nullptr;
if (GetWorld()->TryGetComponent(hMeshComp, pBlackboardComp))
{
  // ...
}
```

The purpose of ezTypedComponentHandle is to catch bugs like these at compile time. But this code compiles successfully because ezTypedComponentHandle decays into ezComponentHandle and the untyped overload of TryGetComponent is called.

x0r04rg:
Fixes https://github.com/ezEngine/ezEngine/issues/1259

If handle and pointer types are not related they will match a deleted TryGetComponent signature and therefore throw a compile time error. Otherwise, the untyped overload will be called.

By "related" I mean if handle type is convertible to the pointer type or if a handle type is a base of a pointer type (so querying a derived type from a handle to a base type is allowed).

~~Also removed a test that does not compile anymore and is now irrelevant. Tell me if you want me to replace it with some other test maybe?~~ Nevermind, the test is still useful.

### DIFF
--- a/Code/Engine/Core/World/World.h
+++ b/Code/Engine/Core/World/World.h
@@ -178,11 +178,11 @@ public:
   [[nodiscard]] bool TryGetComponent(const ezComponentHandle& hComponent, const ComponentType*& out_pComponent) const;
 
   /// \brief Explicitly delete TryGetComponent overload when handle type is not related to a pointer type given by out_pComponent.
-  template <typename T, typename U, std::enable_if_t<!std::disjunction_v<std::is_convertible<T*, U*>, std::is_base_of<T, U>>, bool> = true>
+  template <typename T, typename U, std::enable_if_t<!std::disjunction_v<std::is_base_of<U, T>, std::is_base_of<T, U>>, bool> = true>
   [[nodiscard]] bool TryGetComponent(const ezTypedComponentHandle<T>& hComponent, U*& out_pComponent) = delete;
 
   /// \brief Explicitly delete TryGetComponent overload when handle type is not related to a pointer type given by out_pComponent.
-  template <typename T, typename U, std::enable_if_t<!std::disjunction_v<std::is_convertible<T*, U*>, std::is_base_of<T, U>>, bool> = true>
+  template <typename T, typename U, std::enable_if_t<!std::disjunction_v<std::is_base_of<U, T>, std::is_base_of<T, U>>, bool> = true>
   [[nodiscard]] bool TryGetComponent(const ezTypedComponentHandle<T>& hComponent, const U*& out_pComponent) const = delete;
 
   /// \brief Creates a new component init batch.
--- a/Code/Engine/Core/World/World.h
+++ b/Code/Engine/Core/World/World.h
@@ -177,17 +177,13 @@ public:
   template <typename ComponentType>
   [[nodiscard]] bool TryGetComponent(const ezComponentHandle& hComponent, const ComponentType*& out_pComponent) const;
 
-  template <typename ComponentType>
-  [[nodiscard]] bool TryGetComponent(const ezTypedComponentHandle<ComponentType>& hComponent, ComponentType*& out_pComponent)
-  {
-    return TryGetComponent<ComponentType>(static_cast<const ezComponentHandle&>(hComponent), out_pComponent);
-  }
+  /// \brief Explicitly delete TryGetComponent overload when handle type is not related to a pointer type given by out_pComponent.
+  template <typename T, typename U, std::enable_if_t<!std::disjunction_v<std::is_convertible<T*, U*>, std::is_base_of<T, U>>, bool> = true>
+  [[nodiscard]] bool TryGetComponent(const ezTypedComponentHandle<T>& hComponent, U*& out_pComponent) = delete;
 
-  template <typename ComponentType>
-  [[nodiscard]] bool TryGetComponent(const ezTypedComponentHandle<ComponentType>& hComponent, const ComponentType*& out_pComponent) const
-  {
-    return TryGetComponent<ComponentType>(static_cast<const ezComponentHandle&>(hComponent), out_pComponent);
-  }
+  /// \brief Explicitly delete TryGetComponent overload when handle type is not related to a pointer type given by out_pComponent.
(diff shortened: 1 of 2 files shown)


jankrassnigg (commented):
Looks like a good solution, but I need a quiet moment to look up how this exactly works 🤣

x0r04rg:
The usual logic is kind of inverted here: instead of providing an implementation for valid template arguments (when handle type is related to the pointer type), i'm doing the opposite - i declare a method for invalid arguments and mark it as deleted. So when the method is called with invalid arguments it will match the deleted signature and this will be a compile time error. If the arguments are valid, it will not match the deleted signature and will fall back to the untyped ezComponentHandle overload. This is basically a trap, lol

Code/Engine/Core/World/World.h
```diff
@@ -177,17 +177,13 @@ class EZ_CORE_DLL ezWorld final
...
   [[nodiscard]] bool TryGetComponent(const ezComponentHandle& hComponent, const ComponentType*& out_pComponent) const;
 
-  template <typename ComponentType>
-  [[nodiscard]] bool TryGetComponent(const ezTypedComponentHandle<ComponentType>& hComponent, ComponentType*& out_pComponent)
-  {
-    return TryGetComponent<ComponentType>(static_cast<const ezComponentHandle&>(hComponent), out_pComponent);
-  }
+  /// \brief Explicitly delete TryGetComponent overload when handle type is not related to a pointer type given by out_pComponent.
+  template <typename T, typename U, std::enable_if_t<!std::disjunction_v<std::is_convertible<T*, U*>, std::is_base_of<T, U>>, bool> = true>
+  [[nodiscard]] bool TryGetComponent(const ezTypedComponentHandle<T>& hComponent, U*& out_pComponent) = delete;
```
jankrassnigg: So I understand what the idea is, but I'm unsure about the implementation.

Why the is_convertible check, at all? Wouldn't is_base_of be sufficient?

Also not sure about the argument order of the two. is_convertible<T,U> basically checks whether U is a base class of T, which makes sense, the handle should be to a class of type U or derived.

But then is_base_of<T,U> checks whether T is a base class of U (so the other way round), which doesn't make sense, because the handle has to be to a stricter (derived) type. Did you just accidentally swap those?

And finally, why the "= true" at the end? I guess that's some C++ enable_if detail, but if you have an explanation for my understanding, that would be great.

x0r04rg: Yeah, the is_convertible<T*, U*> check could be replaced with is_base_of<U, T>, I can replace it if you want as i don't have an opinion on which is better.

The second is_base_of<T, U> check is intentionally allowing getting derived types from a handle to a base type because this is convenient and intuitive and user still has to check the result of TryGetComponent so it's not like it will break anything. Also I had a bunch of these in my game project and it felt weird that i have to convert a typed handle to the base class to the untyped one before trying to acquire a pointer to derived class. If you don't like it i can remove it of course.

Regarding "= true", I have no idea either! I copied this from ezTypedComponentHandle copy ctor, and from a brief googling session i was convinced this is "modern" c++ SFINAE form, but i didn't really dug into details (i should do it sometime though).

jankrassnigg: I'd remove the is_convertible check, as I just don't see how that kind of check is relevant in this context.

I am a bit surprised by your argument about the second check.

So you are telling me, you have something like TypedHandle<JoltActor> but then in the TryGet you pass in a JoltDynamicActor* as the target?

Because that way around, it is NOT typesafe anymore, since you are casting from a handle to a base class, to a pointer of a derived type. I would allow the other way round, obviously, but if we allow this way, then the whole idea of typed handles is again a bit moot, no?

I am willing to be a bit more lenient here, for the sake of usability, but it would be helpful to make up my mind, if you could give me a few more concrete examples how you use that and why you feel that this is not actually a problem.

x0r04rg: Let's say there is an abstract `ResourceComponent`, for example, and a few derived ones like `WoodResourceComponent`, `StoneResourceComponent`, etc. And let's say there is a `HarvestResourceBehavior` that has a member `ezTypedComponentHandle<ResourceComponent> m_hResource`. 

And then somewhere in it's logic code it wants to know if m_hResource is actually a WoodResourceComponent, and do something with its pointer. So it feels natural that you could do something like this:
```cpp
WoodResourceComponent* pWoodResource = nullptr;
if (GetWorld()->TryGetComponent(m_hResource, pWoodResource)) // m_hResource is a ezTypedComponentHandle<ResourceComponent>
{
  // ...
}
```

If that wouldn't be possible you would have to do this instead:
```cpp
WoodResourceComponent* pWoodResource = nullptr;
if (GetWorld()->TryGetComponent(ezComponentHandle(m_hResource), pWoodResource)) // <-- m_hResource has to be explicitly converted to the untyped handle
{
  // ...
}
```
Which feels kinda eh?

Still, if you try something that is obviously wrong, it's a compile time error:
```cpp
ezJoltActorComponent* pPhysicsActor = nullptr;
if (GetWorld()->TryGetComponent(m_hResource, pPhysicsActor)) // <-- won't compile
{
  // ...
}
```

So bottom line is: getting a pointer to derived from a handle to base isn't always an error (i would say it is not an error but intended behavior most of the time) but it could fail and that's ok, since the API forces you to check the result. On the other hand, if the handle type and a pointer are completely unrelated (for example ezMeshComponent and ezSpawnComponent) then it's always an error and should be caught at compile time.

jankrassnigg: Oh right, inside TryGetComponent we actually do a ezDynamicCast which ensures that the cast to the target type is possible.

I guess in that case I'm fine with allowing that.

Could you make the remaining adjustment as discussed above, then we can merge it.

x0r04rg: done

PR #1263, opened 2024-04-16
ezPlayer prefix fix

kulaev:
Fixes project export errors. I'm still unable to launch exported project due to missing TracyClient.dll but i'm not sure how to fix it


jankrassnigg:
TracyClient.dll should be built in your own Visual Studio solution under Thirdparty:

image

If it's not there, run `.\GenerateWin64vs2022.bat` again to regenerate the VS solution, then it should show up.

jankrassnigg:
Thanks for the contribution, looks good!

Issue #1262, opened 2024-04-16
[BUG] ezMakeDelegate automatic conversion is error-prone when used with events


x0r04rg:
The `ezMakeDelegate` is able to automatically convert between compatible method signatures. In the context of events this means you can subscribe to an event with a method which signature is not exactly matching.

For example:
```cpp
class MyClass
{
	void HandleByValue(MyEventData data);
	void HandleByRef(MyEventData& data);
	void HandleByConstRef(const MyEventData& data);
};

MyClass* myObj = new MyClass;
ezEvent<MyEventData> myEvent;

myEvent.AddEventHandler(ezMakeDelegate(&MyClass::HandleByValue, myObj)); // this works
myEvent.AddEventHandler(ezMakeDelegate(&MyClass::HandleByRef, myObj)); // this also works
myEvent.AddEventHandler(ezMakeDelegate(&MyClass::HandleByConstRef, myObj)); // and also this
```

However, when unsubscribing from events, things are not so nice anymore:
```cpp
myEvent.RemoveEventHandler(ezMakeDelegate(&MyClass::HandleByValue, myObj)); // this works
myEvent.RemoveEventHandler(ezMakeDelegate(&MyClass::HandleByRef, myObj)); // this will crash (failed assert)
myEvent.RemoveEventHandler(ezMakeDelegate(&MyClass::HandleByConstRef, myObj)); // this will also crash (same as above)
```

I think this is too error-prone, the automatic conversion should be opt-in, and by default it should be a compile-time error when handler's signature is not exactly matching.

In fact, there is a problem like this already in the engine, which leads to a crash (i will make a PR with a fix later today).

jankrassnigg:
@Ingrater might have an opinion about this.

PR #1265, opened 2024-04-16
Fix LocalBlackboardComponent crash

x0r04rg:
Steps to reproduce:
1. Tick "Send Entry Changed Message" on a LocalBlackboardComponent in the editor
2. Untick it
3. Observe the engine process crash

This is the issue i've mentioned here https://github.com/ezEngine/ezEngine/issues/1262

### DIFF
--- a/Code/Engine/Core/Utils/Blackboard.h
+++ b/Code/Engine/Core/Utils/Blackboard.h
@@ -170,7 +170,7 @@ public:
   const ezHashTable<ezHashedString, Entry>& GetAllEntries() const { return m_Entries; }
 
   /// \brief Allows you to register to the OnEntryEvent. This is broadcast whenever an entry is modified that has the flag ezBlackboardEntryFlags::OnChangeEvent.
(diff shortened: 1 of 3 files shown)


x0r04rg:
Btw, if subscription id approach is ok, then there is no use for `BCFlags::SendEntryChangedMessage`, should i remove it?

jankrassnigg:
The approach is okay, but I think this line is already wrong:

```cpp
ezEvent<EntryEvent> m_EntryEvents;
```

That should be

```cpp
ezEvent<const EntryEvent&> m_EntryEvents;
```

x0r04rg:
So should i change it to that instead? And then i could remove the subscription ID stuff and leave BCFlags::SendEntryChangedMessage as it is

jankrassnigg:
Yeah, that would be better, I think. The issue that you raised in the ticket still stands, of course, but that needs to be investigated separately. I agree that that is problematic.

Either way, I don't think that `BCFlags::SendEntryChangedMessage` could be removed, but I'm not sure about that.

x0r04rg:
done

PR #1275, opened 2024-04-29
Disable Tracy and fix some other issues

jankrassnigg:
Tracy always loads the process symbols. This locks the PDB files, which breaks the EZ engine process hot reload of the editor.

Since this is a core feature of the editor, this is unacceptable.

I've attempted to disable all Tracy callstack functionality, for which it even has options, but that doesn't prevent Tracy from locking the PDBs. Consequently, I'm disabling Tracy by default now. If somebody wants Tracy support, they can enable it manually through CMake.

Feel free to attempt to fix this, so that Tracy can be enabled by default.


JailbreakPapa:
Going to take a look at this. some of our engineers are now also experiencing this. possibly TRACY_DELAYED_INIT could help?

Despite that, ill create a PR once I have a fix ready.

JailbreakPapa:
Seems to also cause a deadlock in the tracy threads

JailbreakPapa:
a **possible** fix for this is to just directly manage the lifetime of tracy TRACY_MANUAL_LIFETIME, and whenever the editorengine process restarts, we can just end the tracy process there, and recreate it.

PR #1276, opened 2024-05-02
Particle lighting

### LINKED ISSUE #1196: [FEATURE] Particle Shading
Currently sustained particles look massively out place due to lack of shading within the scene.
When place in shadowed area they look to glow in relation to the scene. 
For effects such as smoke there is no directionality to the lighting making it look out of place in the scene.

Example from unreal:
image

C-Core:
Added optional per-vertex lighting to particle effects:
image

This includes lighting from all light types: directional, point, spot all including shadows as well as ambient/sky light. Fixes #1196

Implemented some debugging modes for particles like "diffuse lit only", "texcoords", "diffuse color" and "pixel/vertex normal".

Also this PR should fix a weird shadow issue that shadows would fade out if they are very close to the camera.

Note that I optimized some of the particle shader code and particle rotation now behaves different than before. All cases including local space simulation etc. still work correctly but the visual result is just different as you can see in some of the test reference images.


Code/EditorPlugins/Scene/EnginePluginScene/SceneView/SceneView.cpp
```diff
@@ -183,6 +183,17 @@ void ezSceneViewContext::SetCamera(const ezViewRedrawMsgToEngine* pMsg)
...
+        auto cameraMode = pCameraComponent->GetCameraMode();
+        float fFovOrDim = pCameraComponent->GetFieldOfView();
+        if (cameraMode == ezCameraMode::OrthoFixedWidth || cameraMode == ezCameraMode::OrthoFixedHeight)
+        {
+          fFovOrDim = pCameraComponent->GetOrthoDimension();
+        }
+
+        const float fNearPlane = pCameraComponent->GetNearPlane();
+        const float fFarPlane = pCameraComponent->GetFarPlane();
+        m_CullingCamera.SetCameraMode(cameraMode, fFovOrDim, fNearPlane, ezMath::Max(fNearPlane + 0.00001f, fFarPlane));
```
jankrassnigg, line 196: what is this change for ?

Code/Engine/RendererCore/Lights/Implementation/ShadowPool.cpp
```diff
@@ -548,29 +560,33 @@ ezUInt32 ezShadowPool::AddDirectionalLight(const ezDirectionalLightComponent* pD
...
+      const ezMat4 worldToLightMatrix = pView->GetViewMatrix(ezCameraEye::Left);
+      const float texelInWorld = (2.0f * radius) / cvar_RenderingShadowsMaxShadowMapSize;
       ezVec3 offset = worldToLightMatrix.TransformPosition(ezVec3::MakeZero());
-      float texelInWorld = (2.0f * radius) / cvar_RenderingShadowsMaxShadowMapSize;
       offset.x -= ezMath::Floor(offset.x / texelInWorld) * texelInWorld;
       offset.y -= ezMath::Floor(offset.y / texelInWorld) * texelInWorld;
 
       camera.MoveLocally(0.0f, offset.x, offset.y);
+
+#if EZ_ENABLED(EZ_COMPILE_FOR_DEVELOPMENT)
```
jankrassnigg, line 582: Do we want to compile these debug rendering modes out? We usually don't do this. 
Also then you should ifdef the cvar as well.

Data/Base/Shaders/Common/Lighting.h
```diff
@@ -171,8 +171,11 @@ float SampleShadow(float3 shadowPosition, float2x2 randomRotation, float penumbr
 #endif
 }
 
-float CalculateShadowTerm(ezMaterialData matData, float3 lightVector, float distanceToLight, uint type,
-  uint shadowDataOffset, float noise, float2x2 randomRotation, out float subsurfaceShadow, out float3 debugColor)
+// Enable to only sample the last cascade for directional lights (which will always enclose the whole shadow range)
+// #define SHADOW_FORCE_LAST_CASCADE
```
jankrassnigg, line 175: Could instead use if constexpr, then the code always gets checked for syntax errors.

jankrassnigg (approved):

Code/EditorPlugins/Scene/EnginePluginScene/SceneView/SceneView.cpp
C-Core: This is for culling debugging so that it takes the given fov, near and far plane from the camera component for culling, not the values from the editor camera.

Data/Base/Shaders/Common/Lighting.h
C-Core: There is no if constexpr in shader code

Code/Engine/RendererCore/Lights/Implementation/ShadowPool.cpp
C-Core: The cvar is already ifdef'ed and we do that in a couple of other places as well. E.g. Extractor.cpp has all its debug code ifdef'ed.

PR #1286, opened 2024-05-12
fix for building without qt support

brightening-eyes:
hello.
this is the fix for cmake that I made in order to disable qt support


brightening-eyes:
As editor requires qt, maybe ez_requires_editor() should check for it?

jankrassnigg:
> As editor requires qt, maybe ez_requires_editor() should check for it?

Not sure, what does it do now without Qt? Does it create the editor project and then fail to build? If so, it probably make sense to do the check for Qt.

brightening-eyes:
as far as it goes, it seems the build succeeded. both in my machine and in the pipelines.
It seems the editor was not built and it can detect that.

PR #1288, opened 2024-05-12
Render Buffer Refactor and SRV / UAV split

Sanakan8472:
This PR is mostly boilerplate code, the only change of interest is in *Code/Engine/RendererFoundation/Descriptors/Descriptors.h* and *Code/Engine/RendererFoundation/Descriptors/Enumerations.h*.

## Buffer Refactor
* `ezGALBufferCreationDescription` has all its bools and `ezGALBufferType` removed. Especially `ezGALBufferType::Generic` was very confusing. This has been replaced with a single bitflags field: `ezGALBufferFlags`
* `ezGALBufferFlags` defines what you are going to use the buffer for: `VertexBuffer`, `IndexBuffer`, `ConstantBuffer`, `TexelBuffer`, `StructuredBuffer` or `ByteAddressBuffer`. These use the same name as you woud find in the `ezGALShaderResourceType` enum. Note that these are flags, as you can have for example a `VertexBuffer` that you want to write as a `ByteAddressBuffer`.
* Additionally, `ezGALBufferFlags` contains flags where the buffer can be used: `ShaderResource`, `UnorderedAccess` and `DrawIndirect`.

## SRV / UAV split
* `ezGALResourceViewCreationDescription` has been split into `ezGALBufferResourceViewCreationDescription` and `ezGALTextureResourceViewCreationDescription`.
* `ezGALUnorderedAccessViewCreationDescription` has been split into `ezGALBufferUnorderedAccessViewCreationDescription` and `ezGALTextureUnorderedAccessViewCreationDescription`.
* This remove ambiguity and a source of errors as you could bind idiotic views to descriptors of incompatible types and the low level renderer needed to account for this. It also makes it much clearer which fields are actually usable when creating the view.
* The old code forced all UAV / SRV / RenderTarget caches to be on the `ezGALResourceBase` causing every resource type to be bloated with the cache arrays although no resource outside of texture and buffer uses these. The caches have now been moved into the `ezGALBuffer` and `ezGALTexture` classes.
* Removed `m_bAppend` from `ezGALBufferUnorderedAccessViewCreationDescription` as it was not used and this feature (append / consume) requires a dedicated resource type to work cross platform.

## Misc
* Disable debug break on warnings in DX11 debug layer.
* Added `ezGALDevice::GetRendererPlatform` to query what renderer is currently active.
* Increased image compare threshold for Nvidia cards running on DX11 as the latest driver decided to render all lines twice as thick for no reason.

### DIFF
--- a/Code/Engine/RendererFoundation/Descriptors/Enumerations.h
+++ b/Code/Engine/RendererFoundation/Descriptors/Enumerations.h
@@ -19,7 +19,7 @@ struct ezGALShaderResourceType
     // Read-only struct. Set directly via ezGALCommandEncoder::SetPushConstants. HLSL: Use macro BEGIN_PUSH_CONSTANTS, END_PUSH_CONSTANTS, GET_PUSH_CONSTANT
     PushConstants,
 
-    /// \name Shader Resource Views (SRVs). These are set via ezGALResourceViewHandle.
+    /// \name Shader Resource Views (SRVs). These are set via ezGALTextureResourceViewHandle.
     ///@{
 
     /// Read-only texture view. When set, ezGALShaderTextureType is also set. HLSL: Texture*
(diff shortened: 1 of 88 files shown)


Code/Engine/RendererFoundation/Descriptors/Enumerations.h
```diff
@@ -162,24 +166,42 @@ struct ezGALVertexAttributeSemantic
   };
 };
 
-/// \brief General type of buffer.
+/// \brief Defines for what purpose a buffer can be used for.
 /// \sa ezGALBufferCreationDescription
-struct ezGALBufferType
+struct ezGALBufferFlags
```
C-Core: Maybe we could call it BufferUsageFlags. I don't have a strong preference though.

C-Core (approved):

Code/Engine/RendererFoundation/Descriptors/Enumerations.h
Sanakan8472: That works too, fixed!

PR #1292, opened 2024-05-14
drag drop importing

PlasmaDev5:
Adds the ability to drag source files onto the asset browser in order to import it into the current directory. 
not sure what else to say, it does what it says on the tin.

This change also adds the ability to move assets in editor via drag and drop.

### DIFF
--- a/Code/Editor/EditorFramework/Assets/Implementation/AssetBrowserView.cpp
+++ b/Code/Editor/EditorFramework/Assets/Implementation/AssetBrowserView.cpp
@@ -109,37 +109,32 @@ void ezQtAssetBrowserView::dropEvent(QDropEvent* pEvent)
     src.Shrink(1, 0); // remove prepending '/' in name that appears for some reason
 
     ezStringBuilder target = targetPar;
-    target.AppendFormat("/{}", qtToEzString(it->fileName()));
+    target.AppendPath(qtToEzString(it->fileName()));
 
     if (targetPar == src)
     {
-      ezLog::Error("Can't drop a file or folder on itself");
+      ezLog::Error("Can't drop a file or folder onto itself.");
       continue;
     }
 
 
     if (!ezOSFile::ExistsDirectory(src) && !ezOSFile::ExistsFile(src))
     {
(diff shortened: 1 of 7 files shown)


PlasmaDev5:
@jankrassnigg not sure what im missing on that one as ezOSFile seams valid locally and on other platforms

jankrassnigg:
It's just missing a #include.
Typically we build with so called "unity files", ie. multiple CPPs are combined into one, but on this build machine we disable that and then sometimes if built individually, a CPP doesn't have all the includes that it needs.

Code/Editor/EditorFramework/Assets/Implementation/AssetBrowserView.cpp
```diff
@@ -71,6 +74,122 @@ ezInt32 ezQtAssetBrowserView::GetIconScale() const
...
+{
+  pEvent->acceptProposedAction();
+}
+
+void ezQtAssetBrowserView::dragLeaveEvent(QDragLeaveEvent* pEvent)
+{
+  pEvent->accept();
+}
+
+void ezQtAssetBrowserView::dropEvent(QDropEvent* pEvent)
```
jankrassnigg, line 101: When is this codepath ever executed?

Also, why does this drop handler only copy files, and the other one on the widget triggers an import?

PlasmaDev5: So when my junior developer put this together we spent a while confused by this. From what we could find the drop events for internal drag and drop only triggers on the view but external drag and drop only triggers on the widget. 

Due to this is means this will only be used for moving files within the asset browser such as drag asset into a folder.

PlasmaDev5:
Thanks for the small fixes :) was going to do the import fix now but you beat me to it

Code/Editor/EditorFramework/Assets/Implementation/AssetBrowserView.cpp
jankrassnigg: Okay, i'll give that another try tomorrow.

PR #1293, opened 2024-05-15
jolt can now be built like other libs of ezengine

### LINKED ISSUE #1289: jolt: different options based on the architecture and preference in build
hello,
I opened this issue in order to represent jolt definitions better in ezengine's cmake scripts
* first, no matter how you build ezengine (weather static or dynamic), jolt gets built as a dynamic library. like enet, lua etc it should be set as the preference for each user.
* second, jolt.cmake has many useful defines (like double precision, different instruction sets) etc which can be tuned in the cmake script to have better performance. maybe modifying and then including jolt.cmake directly should make sense? because it handles stuff like fma, floating point exceptions, fastmath etc.
* jolt.h should be the precompiled header I suppose, as the jolt.cmake script sets it.
* I also suppose in cmake, things like USE_AVX, USE_AVX2 etc need to be have EZ_ prefix? maybe its not needed at all and I am biased a little.
thanks

brightening-eyes:
fixes #1289

### DIFF
--- a/Code/ThirdParty/Jolt/CMakeLists.txt
+++ b/Code/ThirdParty/Jolt/CMakeLists.txt
@@ -5,9 +5,11 @@ ez_requires(EZ_3RDPARTY_JOLT_SUPPORT)
 # Get the name of this folder as the project name
 get_filename_component(PROJECT_NAME ${CMAKE_CURRENT_SOURCE_DIR} NAME_WE)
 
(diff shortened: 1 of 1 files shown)


brightening-eyes:
thing is that only and only one check (android x64) is failing.
maybe its due to the defines?

jankrassnigg:
> thing is that only and only one check (android x64) is failing. maybe its due to the defines?

It looks like the failure is unrelated to the actual change, so I'll just merge it anyway.

Thanks for the contribution!

PR #1298, opened 2024-05-20
Fix particles disappearing on Nvidia DX11

Sanakan8472:
* The DX11 shader compiler no longer sets binding information for samplers and SRVs as Nvidia drivers can't cope with binding mappings that have gaps.
* This relaxes the renderer resource binding contraints in that it allows a resource name to exist multiple times inside the shader resource binding. This is fine as we don't do name lookups on `ezGALShader::GetBindingMapping` except for in `ezMaterialResource::UpdateConstantBuffer` which uses `ezGALShader::GetShaderResourceBinding(const ezTempHashedString& sName)` to lookup the first binding of a given name but as it's only used for constant buffers no ambiguity is possible.

### DIFF
--- a/Code/Engine/RendererCore/Shader/Implementation/ShaderPermutationBinary.cpp
+++ b/Code/Engine/RendererCore/Shader/Implementation/ShaderPermutationBinary.cpp
@@ -11,6 +11,7 @@ struct ezShaderPermutationBinaryVersion
     Version3 = 3,
     Version4 = 4,
     Version5 = 5,
+    Version6 = 6, // Fixed DX11 particles vanishing
 
     // Increase this version number to trigger shader recompilation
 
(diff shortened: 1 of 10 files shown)


jankrassnigg:
Was this the reason why the particle editor sometimes stopped working?

Sanakan8472:
> Was this the reason why the particle editor sometimes stopped working?

Yes. It's a driver bug on Nvidia. You will need to clear your shader cache for the fix to take effect.

jankrassnigg:
> > Was this the reason why the particle editor sometimes stopped working?
> 
> Yes. It's a driver bug on Nvidia. You will need to clear your shader cache for the fix to take effect.

We have a shader version number somewhere to force that.

Sanakan8472:
> We have a shader version number somewhere to force that.

Fixed!

C-Core (approved):

PR #1299, opened 2024-05-21
Core: Add Center Window On Display Option.

TheophilusE:
* Fixes window not showing up on selected display.


C-Core (approved):
Thanks for the fix and the additional option looks useful too.

PR #1302, opened 2024-05-22
Draw ezDebugRenderer last in the frame

RiederAlex:
Changed the order of the debug renderer to be drawn even after GUI.

This is important to avoid GUI (e.g. RMLUI) overdrawing important debug info (e.g. `App.ShowFPS=true`).
As a user I expect my debug info to always show up and not be hidden by anything.

Pre-Fix:
PreFix

Post-Fix:
PostFix

### DIFF
--- a/Code/Engine/RendererCore/Debug/DebugRenderer.h
+++ b/Code/Engine/RendererCore/Debug/DebugRenderer.h
@@ -225,8 +225,11 @@ public:
 private:
   friend class ezSimpleRenderPass;
 
-  static void Render(const ezRenderViewContext& renderViewContext);
-  static void RenderInternal(const ezDebugRendererContext& context, const ezRenderViewContext& renderViewContext);
+  static void RenderScreenSpace(const ezRenderViewContext& renderViewContext);
+  static void RenderInternalScreenSpace(const ezDebugRendererContext& context, const ezRenderViewContext& renderViewContext);
+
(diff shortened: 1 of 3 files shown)


C-Core (changes requested):
I thought about this as well but this is not a proper fix. Now also the 3D debug geometry is rendered on top of the UI which is not what we want. The ezDebugRenderer::Render function has to be split in a 3D and a 2D part. The 3D part stays at the place as it is now and the 2D part is after the UI.

RiederAlex:
Good point - I'll try to adapt the ezDebugRenderer as you've suggested.

RiederAlex:
I've split the 2D and 3D parts into its own sections, so now it looks like this - the yellow looking bars are RMLUI divs.

Pre-Fix:
PreFix

Post-Fix:
PostFix

Notice that 3D Text still gets clipped by RMLUI.

C-Core:
The game engine unit tests now fail. Most likely needs updated reference images. If you don't know how to do that I can take a look later.

RiederAlex:
Sorry - my bad for not running the unit tests before pushing something.

As I currently only have an AMD card at my disposal and I seem to have touched the one unit test which has image reference overrides for multiple platforms help would be greatly appreciated!

RiederAlex:
I've found & fixed the issue - it was caused by a one frame delay introduced by my change. 
The unit tests should now run through without having to update the reference images.

C-Core (approved):

PR #1303, opened 2024-05-22
Fix Changing SetWorldSimulationEnabled creates inconsistent behavior

### LINKED ISSUE #1300: [BUG] Changing SetWorldSimulationEnabled creates inconsistent behavior
Games need a way to pause/unpause the simulation - however doing this the most straighforward way via `pWorld->SetWorldSimulationEnabled(true/false)` inside a components update causes inconsistencies.

The main issue is that `pWorld->SetWorldSimulationEnabled(true/false)` instantly resumes simulating during `ezWorld::Update`, causing only parts of a frame being simulated.
This is double bad if unpausing, as this causes `GetWorld()->GetClock().GetTimeDiff().AsFloatInSeconds()` to be 0 for the rest of the frame while updating components which do not expect it (e.g. where `m_bOnlyUpdateWhenSimulating` = true), causing assertions/divides by 0/ect. 
See `void ezWorld::UpdateSynchronous` for an example where this happens.

**Proposed solutions**

1. `ezWorld::Get/Set WorldSimulationEnabled` are delayed
    A new boolean is introduced which remembers the desired state of the world simulation and flushes it into `m_Data.m_bSimulateWorld` at the beginning of the Update() loop
    With this solution I'm afraid of side effects I do not know about and certain decisions have to be made which might not be easily understandable as a user, like if GetWorldSimulationEnabled() return the current state of or the desired state.

2. Pause/Unpause is implemented in GameState and delayed
    Similar to proposed solution 1, but with a lot less side effects as its a new API - we implement a Pause/Unpause/IsPaused in GameState and flush it in `GameState::ProcessInput()`. 
This would give users a clear way to pause the simulation, but it would introduce a new API which accomplishes something that can also be done differently (via directly accessing the world), which is always a sub-optimal solution.

3. `ezWorld::Update()` saves the state of m_bSimulateWorld at the beginning of the call and runs the rest of the update loop with that state
      So calls like `ezWorld::UpdateSynchronous` or `ezWorld::UpdateAsynchronous` would no longer use `m_Data.m_bSimulateWorld`, but instead something like `bSimulateThisFrame`, which is a constant value set at the beginning of `ezWorld::Update()`

Personally I think solution 3 is the one desired here, as it creates consistency (a frame is always completely simulated, or not), does not introduce a new API or complicated addtional state in ezWorld.

If desired, I would take over the implementation of the solution for this issue.

RiederAlex:
`ezWorld::SetWorldSimulation` will now be cached and is only flushed into the actual simulation state in the beginning/end of `ezWorld::Update()` calls.

This ensures consistency for the Update-functions which only execute when the simulation is active - either all of them are run, or none. Its no longer possible to get an in between state where only some of them are run if the simulation state is changed during an `ezWorld::Update()` call.

This fixes #1300

### DIFF
--- a/Code/Engine/Core/World/Implementation/WorldData.h
+++ b/Code/Engine/Core/World/Implementation/WorldData.h
@@ -241,7 +241,7 @@ namespace ezInternal
     mutable ezAtomicInteger32 m_iReadCounter;
 
     ezUInt32 m_uiUpdateCounter = 0;
-    bool m_bDesiredSimulateWorld = true;
+    bool m_bShouldSimulateWorld = true;
     bool m_bSimulateWorld = true;
     bool m_bReportErrorWhenStaticObjectMoves = true;
 
--- a/Code/Engine/Core/World/Implementation/WorldData.h
+++ b/Code/Engine/Core/World/Implementation/WorldData.h
@@ -241,6 +241,7 @@ namespace ezInternal
     mutable ezAtomicInteger32 m_iReadCounter;
 
     ezUInt32 m_uiUpdateCounter = 0;
+    bool m_bDesiredSimulateWorld = true;
     bool m_bSimulateWorld = true;
     bool m_bReportErrorWhenStaticObjectMoves = true;
 
--- a/Code/Engine/Core/World/World.h
+++ b/Code/Engine/Core/World/World.h
@@ -251,12 +251,12 @@ public:
   ///@}
 
   /// \brief If enabled, the full simulation should be executed, otherwise only the rendering related updates should be done
-  /// Note the state returned by \link GetWorldSimulationEnabled() will not reflect state of \param bEnable immediately,
-  /// it is only updated in the beginning/end of an \link Update() call.
+  /// Note the state returned by ezWorld::GetWorldSimulationEnabled() will not reflect state of \param bEnable immediately,
+  /// it is only updated in the beginning/end of an ezWorld::Update() call.
   void SetWorldSimulationEnabled(bool bEnable);
 
   /// \brief If enabled, the full simulation should be executed, otherwise only the rendering related updates should be done
-  /// \link SetWorldSimulationEnabled() for implementation specifics.
+  /// See ezWorld::SetWorldSimulationEnabled() for implementation specifics.
   bool GetWorldSimulationEnabled() const;
 
   /// \brief Updates the world by calling the various update methods on the component managers and also updates the transformation data of
--- a/Code/Engine/Core/World/World.h
+++ b/Code/Engine/Core/World/World.h
@@ -251,9 +251,12 @@ public:
   ///@}
 
   /// \brief If enabled, the full simulation should be executed, otherwise only the rendering related updates should be done
+  /// Note the state returned by \link GetWorldSimulationEnabled() will not reflect state of \param bEnable immediately,
+  /// it is only updated in the beginning/end of an \link Update() call.
   void SetWorldSimulationEnabled(bool bEnable);
 
   /// \brief If enabled, the full simulation should be executed, otherwise only the rendering related updates should be done
+  /// \link SetWorldSimulationEnabled() for implementation specifics.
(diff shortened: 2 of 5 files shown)


Code/Engine/Core/World/World.h
```diff
@@ -251,9 +251,12 @@ class EZ_CORE_DLL ezWorld final
   ///@}
 
   /// \brief If enabled, the full simulation should be executed, otherwise only the rendering related updates should be done
+  /// Note the state returned by \link GetWorldSimulationEnabled() will not reflect state of \param bEnable immediately,
```
C-Core: There is no need to \link we don't use it elsewhere, links will be automatically inserted.

Code/Engine/Core/World/Implementation/WorldData.h
```diff
@@ -241,6 +241,7 @@ namespace ezInternal
     mutable ezAtomicInteger32 m_iReadCounter;
 
     ezUInt32 m_uiUpdateCounter = 0;
+    bool m_bDesiredSimulateWorld = true;
```
C-Core: Typically we would name a variable something with "should" or "want". I think "m_bShouldSimulateWorld" sounds good and easy to understand.

Code/UnitTests/CoreTest/World/WorldTest.cpp
```diff
@@ -198,6 +198,62 @@ namespace
...
+      if (m_bToggleSimulationNextAlwaysUpdate)
+      {
+        GetWorld()->SetWorldSimulationEnabled(!GetWorld()->GetWorldSimulationEnabled());
+        m_bToggleSimulationNextAlwaysUpdate = false;
+      }
+
+      m_bWasSimulationUpdateCalledLast = false;
+    }
+
+    ezInt32 m_nTimesSimulationUpdateCalled = 0;
```
C-Core: For integer we use the "i" prefix, so "m_iTimesSimulationUpdateCalled" etc. See https://github.com/ezEngine/ezEngine/blob/dev/CodingGuidelines.md

Code/UnitTests/CoreTest/World/WorldTest.cpp
```diff
@@ -710,6 +766,55 @@ EZ_CREATE_SIMPLE_TEST(World, World)
...
+    world.SetWorldSimulationEnabled(true);
+
+    auto pModule = world.GetOrCreateModule<SimulationTestModule>();
+
+    // World simulation is active, both calls should update
+    world.Update();
+    EZ_TEST_INT(pModule->m_nTimesAlwaysUpdateCalled, 1);
+    EZ_TEST_INT(pModule->m_nTimesSimulationUpdateCalled, 1);
+
+    // This is a sanity check to ensure the order of the Update() calls is AlwaysUpdate() and then SimulationUpdate()
```
C-Core: If there is no priority given update functions in the same phase are sorted by name, so AlwaysUpdate will come before SimulationUpdate. If you want to enforce that use a priority on the desc when registering the update function. Higher priorities (higher number) are called earlier.

C-Core (approved):
Thanks for the fix and even a unit test, love that! I've commented a few minor things.

Code/Engine/Core/World/World.h
RiederAlex: Fixed in 63e08d6

Code/Engine/Core/World/Implementation/WorldData.h
RiederAlex: Fixed in 63e08d6

Code/UnitTests/CoreTest/World/WorldTest.cpp
RiederAlex: Fixed in 63e08d6

Code/UnitTests/CoreTest/World/WorldTest.cpp
RiederAlex: Fixed in 63e08d6

PR #1305, opened 2024-05-23
Debug Text and other improvements

### LINKED ISSUE #1058: DebugRenderer text rendering should use fixed resolution
Debug renderer text becomes very small on 4K monitors.

We should use a fixed, virtual resolution (say 1000 horizontally) and adjust the screen position and size accordingly.

C-Core:
* Changed the debug font to have a black outline to improve readability against bright backgrounds
* Added a cvar to scale the size of the debug text. In the editor this cvar is set to the dpi scale to make debug text more readable on e.g. 4k screens. Fixes #1058

image

* Exposed a few more math functions to visual script
* Exposed jolt sleep settings since they are pretty aggressive by default


Code/EnginePlugins/JoltPlugin/System/JoltWorldModule.cpp
```diff
@@ -797,6 +797,14 @@ void ezJoltWorldModule::UpdateSettingsCfg()
 void ezJoltWorldModule::ApplySettingsCfg()
 {
   SetGravity(m_Settings.m_vObjectGravity, m_Settings.m_vCharacterGravity);
+
+  if (m_pSystem)
+  {
+    auto physicsSettings = m_pSystem->GetPhysicsSettings();
+    physicsSettings.mPointVelocitySleepThreshold = m_Settings.m_fSleepVelocityThreshold;
```
jankrassnigg, line 804: Are there other thresholds?

jankrassnigg (approved):

Code/EnginePlugins/JoltPlugin/System/JoltWorldModule.cpp
C-Core: There are tons of settings. But none super relevant, at least not for now.

PR #1306, opened 2024-05-23
Renderer: CommandEncoder Refactor

Sanakan8472:
## Window Management
* Added `ezGameApplicationBase::Run_AcquireImage`: This function can be used to acquire a new image from a swap-chain or do any other update operations on windows before the multi-threaded rendering and update phase starts.
  * Currently, we don't acquire the image here yet, but we do the recreation of the swapchain on window resize as it is outside the rendering loop and thus a safe point to modify.
  * `ezGameApplication` implements this function to call `ezWindowOutputTargetBase::AcquireImage`.
* Renamed `ezGameApplicationBase::Run_Present` to `Run_PresentImage` to make the connection clearer between the two functions.
  * `ezGameApplication` implements this function to call `ezWindowOutputTargetBase::PresentImage` (renamed as well).
* Added `ezWindowBase::IsVisible`. Describes whether the window can potentially be seen by the user. E.g. windows that are minimized or hidden are not visible. Currently we don't support hiding windows so only minimization is implemented. This can be used to not draw anything while the window can't be seen, see ShaderExplorer.cpp as an example.
* Fixed a bug in which minimizing the window would trigger a resize event to 0x0 on Windows.
* `ezWindowOutputTargetXR::RenderCompanionView` logic has been split into `CompanionViewBeginFrame` and `CompanionViewEndFrame` to account for swap-chain changes below and fix capturing the frame that was just rendered instead of the previous content of the swap-chain texture. This fixes Vulkan validation issues due to reading from the swap-chain image after presenting already.

## Swap-chain Changes
* Swap-chain acquire / present has been moved from the `BeginPipeline` / `EndPipeline` calls and is now done via `ezGALDevice::EnqueueFrameSwapChain`. The function adds a swap-chain to be used for the next frame. Must be called before or during the `ezGALDeviceEvent::BeforeBeginFrame` event (`BeginFrame` function) and repeated for every frame the swap-chain is to be used. This approach guarantees that all swap-chains of a frame acquire and present at the same time, which improves frame pacing. It also allows for other code paths to inspect or otherwise depend on the content of the swap-chain texture before presenting. The device will ensure to acquire an image from the swap-chain during `BeginFrame` and present it when calling `EndFrame`.
* In theory, the swap-chain aquire / present logic could be moved to the `ezGameApplicationBase::Run_AcquireImage` and `ezGameApplicationBase::Run_PresentImage` calls instead at some point but right now the logic whether we draw is too far detached from the window output code to make that work.

## Renderer Changes
* Removed `ezGALPass` and `ezGALDevice::BeginPass` / `EndPass` function. Instead, `BeginCommands` / `EndCommands` has been added which returns a `ezGALCommandEncoder` directly. Functionality of the `ezGALPass` has been merged into the `ezGALCommandEncoder` or `ezGALDevice`.
* All `BeginPass` calls have simply been deleted. Most of them were in passes which have been replaced by a single profiling scope in the for each pass render pipeline loop.
* Removed `ezGALRenderCommandEncoder` and `ezGALComputeCommandEncoder` by merging into `ezGALCommandEncoder`. The separation made little sense given that a dynamic subset of functions switches from accessible to inaccessible and vise versa when starting to a render phase. Logic will be added to those function in question to assert when violated once the high level renderer can guarantee no violations.
* Removed `ezGALDevice::BeginPipeline` / `EndPipeline`. See swap-chain changes. Where used, `BeginCommands` / `EndCommands` was added instead.
* `ezRenderContext::BeginRendering` and `BeginCompute` no longer return an `ezGALCommandEncoder` or require one as an input. Instead, it is always present and can be retrieved via `ezRenderContext::GetCommandEncoder`. For custom `ezRenderContext` instances it is passed in via the costructor. For the default instance the encoder is present as long as you are within a device `BeginCommands` / `EndCommands` scope.
* Moved `ezGALDevice` logic from `ezGameApplication` into the `ezRenderWorld`.
* Fixed uninitialized variable in `CalculateShadowTerm` shader function as the param was `out` but not written in every code path.
* Fixed stereo debug text rendering. This is now used in the stereo tests since the images needed to be re-generated anyways as we are now capturing the current rendered image. This makes it easier to reason about test failure. Due to multi-threaded rendering, we will capture the content of frame 99 during the 100th update loop though.


Code/Engine/RendererFoundation/CommandEncoder/CommandEncoder.h
```diff
@@ -5,11 +5,16 @@
...
 #include <RendererFoundation/CommandEncoder/CommandEncoderState.h>
 
+struct ezGALRenderingSetup;
+
 class EZ_RENDERERFOUNDATION_DLL ezGALCommandEncoder
 {
   EZ_DISALLOW_COPY_AND_ASSIGN(ezGALCommandEncoder);
 
 public:
+  ezGALCommandEncoder(ezGALDevice& in_device, ezGALCommandEncoderCommonPlatformInterface& in_commonImpl);
```
C-Core: We don't use in_ prefix anywhere else

C-Core (approved):

Code/Engine/RendererFoundation/CommandEncoder/CommandEncoder.h
Sanakan8472: I was forced to pick one out of in / out / ref by clangTidy, gosh so many options, hmm ref?

PR #1309, opened 2024-05-25
Fix Incorrect Window Position Calculation.

TheophilusE:
This occurred when either the expected window width or height was greater than the monitor's width or height.

On a vertical monitor setup, I encountered this scenario that the window won't be visible when the width of the window exceeds that of the monitor. It turns out that the window parameters are unsigned integers, so the subtraction gets promoted to that (becomes UINT_MAX).

image


Sanakan8472:
Thanks for the fix @TheophilusE ! Much appreciated!

jankrassnigg (approved):

PR #1310, opened 2024-05-25
Various Build and Test improvements

Sanakan8472:
* Add GCC checks to ezUtils.cmake as clang doesn't require the Ob1 fix.
* Fixed project compilation on Linux. The `architecture` field is only supported on MSVC and must be removed otherwise. Removing is now outside the MSVC scope and marks the settings as dirty.
* Allow pulling monster game on Linux by removing the .exe from the git command.
* Clang fix in StackTraceLogParser.h: `EZ_EDITORFRAMEWORK_DLL` must be at the start of the function declaration or a compiler error is triggered.
* Android testing: Tests now capture all logcat logs but only print ezEngine logs to console. Added exponential backoff in case of app install errors
* Reduced cmake preset version rquirements so the default Ubuntu LTS CMake version is working out of the box.

### DIFF
--- a/Code/Editor/EditorFramework/CodeGen/CppProject.cpp
+++ b/Code/Editor/EditorFramework/CodeGen/CppProject.cpp
@@ -1001,6 +1001,16 @@ namespace
     inout_json[sName] = sValue;
     inout_modified = ezCppProject::ModifyResult::MODIFIED;
   }
+
+  void Remove(ezVariantDictionary& inout_json, ezStringView sName, ezCppProject::ModifyResult& inout_modified)
+  {
+    if (inout_json.Contains(sName))
+    {
+      inout_json.Remove(sName);
+      inout_modified = ezCppProject::ModifyResult::MODIFIED;
+    }
+  }
+
 } // namespace
 
 ezCppProject::ModifyResult ezCppProject::ModifyCMakeUserPresetsJson(const ezCppSettings& cfg, ezVariantDictionary& inout_json)
@@ -1068,10 +1078,11 @@ ezCppProject::ModifyResult ezCppProject::ModifyCMakeUserPresetsJson(const ezCppS
       Modify(presetDict, "architecture", "x64", result);
     }
     else
-    {
-      presetDict.Remove("architecture");
-    }
 #endif
+    {
+      Remove(presetDict, "architecture", result);
+    }
+
   }
 
   return result;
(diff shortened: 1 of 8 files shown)


Code/Editor/EditorFramework/CodeGen/CppProject.cpp
```diff
@@ -1001,6 +1001,16 @@ namespace
     inout_json[sName] = sValue;
     inout_modified = ezCppProject::ModifyResult::MODIFIED;
   }
+
+  void Remove(ezVariantDictionary& inout_json, ezStringView sName, ezCppProject::ModifyResult& inout_modified)
+  {
+    if (inout_json.Contains(sName))
+    {
+      inout_json.Remove(sName);
```
jankrassnigg: Doesn't this remove function already return whether something was removed?

Code/Editor/EditorFramework/EditorApp/EditorApp.cpp
```diff
@@ -367,7 +367,11 @@ ezStatus ezQtEditorApp::MakeRemoteProjectLocal(ezStringBuilder& inout_sFilePath)
 
     QProcess proc;
     proc.setWorkingDirectory(sTargetDir.GetData());
+#if EZ_ENABLED(EZ_PLATFORM_WINDOWS_DESKTOP)
```
jankrassnigg, line 370: We have similar code elsewhere, not sure whether or does the same thing, it whether it appends a platform specific suffix or calls some function to do that.

jankrassnigg (approved):

Code/Editor/EditorFramework/CodeGen/CppProject.cpp
Sanakan8472: Oh it does! Much better, thanks!

Code/Editor/EditorFramework/EditorApp/EditorApp.cpp
Sanakan8472: So I've looked around and we have multiple cases of the platform ifdef pattern like this one here, but there is also this pattern:
```
#  if EZ_ENABLED(EZ_PLATFORM_WINDOWS)
  pathToArchiveTool.Append(".exe");
#  endif
```
And a lot of: "ups we forgot to make this work on linux"!
Haven't found anything more fancy than those.

PR #1316, opened 2024-06-03
Adding parent mapping support for action maps

Sanakan8472:
Previously, it was impossible to make a plugin that adds an action to every editor as you would manually have to add the action to every editor's mappings. To fix this, inheritance has been introduced into the action mapping system. When registering an action via `ezActionMapManager::RegisterActionMap` a  parent mapping can be provided. Once a mapping is mapped to a view, all actions of the entire parent hierarchy as show in the view.

The `ezActionMap` had to be changed slightly to account for this:
1. `MapAction` just accumulates the mappings in an array until `BuildActionTree` is called which builds the hierarchy of the inherited and local actions. This is necessary as mappings can be defined in any order so the parent mapping might not be defined yet and mapping an action into the parent mapping's hierarchy might not be possible yet. 
2. `UnmapAction` has been removed from the public interface and replaced with `HideAction` which practically does the same, but later at the `BuildActionTree` phase.
3. `SearchPathForAction` has been made private as is requires the hierarchy to be built already which is not possible when defining the mapping across parent hierarchies which are not defined yet. The few use cases we had, have been fixed by adding a parameter `const ezBitflags<ezStandardMenuTypes> menus` to `ezProjectActions::MapActions` to allow for partial mapping of actions.

Right now, the following base action mappings have been defined which can now be extended via plugins to easily map to every asset document editor's UI:
1. **AssetMenuBar**: The bare minimum of actions every asset document's menu bar should have.
2. **AssetToolbar**: The bare minimum of actions every asset document's toolbar should have.
3. **AssetViewToolbar**: Empty but should be the base mapping of every view.
4. **SimpleAssetViewToolbar**: Convenience mapping that is derived from *AssetViewToolbar* and provides the most common actions used by engine views.

### DIFF
--- a/Code/EditorPlugins/Fmod/EditorPluginFmod/EditorPluginFmod.cpp
+++ b/Code/EditorPlugins/Fmod/EditorPluginFmod/EditorPluginFmod.cpp
@@ -18,19 +18,11 @@ void OnLoadPlugin()
   // Mesh
   {
     // Menu Bar
-    ezActionMapManager::RegisterActionMap("SoundBankAssetMenuBar").IgnoreResult();
-    ezStandardMenus::MapActions("SoundBankAssetMenuBar", ezStandardMenuTypes::Default | ezStandardMenuTypes::Edit);
-    ezProjectActions::MapActions("SoundBankAssetMenuBar");
-    ezDocumentActions::MapMenuActions("SoundBankAssetMenuBar");
-    ezAssetActions::MapMenuActions("SoundBankAssetMenuBar");
-    ezCommandHistoryActions::MapActions("SoundBankAssetMenuBar");
+    ezActionMapManager::RegisterActionMap("SoundBankAssetMenuBar", "AssetMenuBar").IgnoreResult();
 
     // Tool Bar
     {
-      ezActionMapManager::RegisterActionMap("SoundBankAssetToolBar").IgnoreResult();
-      ezDocumentActions::MapToolbarActions("SoundBankAssetToolBar");
-      ezCommandHistoryActions::MapActions("SoundBankAssetToolBar", "");
-      ezAssetActions::MapToolBarActions("SoundBankAssetToolBar", true);
+      ezActionMapManager::RegisterActionMap("SoundBankAssetToolBar", "AssetToolbar").IgnoreResult();
     }
   }
 
--- a/Code/EditorPlugins/Fmod/EditorPluginFmod/EditorPluginFmod.cpp
+++ b/Code/EditorPlugins/Fmod/EditorPluginFmod/EditorPluginFmod.cpp
@@ -18,11 +18,11 @@ void OnLoadPlugin()
   // Mesh
   {
     // Menu Bar
-    ezActionMapManager::RegisterActionMap("SoundBankAssetMenuBar", "AssetMenuBar").IgnoreResult();
+    ezActionMapManager::RegisterActionMap("SoundBankAssetMenuBar", "AssetMenuBar");
 
     // Tool Bar
     {
-      ezActionMapManager::RegisterActionMap("SoundBankAssetToolBar", "AssetToolbar").IgnoreResult();
+      ezActionMapManager::RegisterActionMap("SoundBankAssetToolBar", "AssetToolbar");
     }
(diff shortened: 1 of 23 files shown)


Code/EditorPlugins/Fmod/EditorPluginFmod/EditorPluginFmod.cpp
```diff
@@ -18,19 +18,11 @@ void OnLoadPlugin()
   // Mesh
   {
     // Menu Bar
-    ezActionMapManager::RegisterActionMap("SoundBankAssetMenuBar").IgnoreResult();
-    ezStandardMenus::MapActions("SoundBankAssetMenuBar", ezStandardMenuTypes::Default | ezStandardMenuTypes::Edit);
-    ezProjectActions::MapActions("SoundBankAssetMenuBar");
-    ezDocumentActions::MapMenuActions("SoundBankAssetMenuBar");
-    ezAssetActions::MapMenuActions("SoundBankAssetMenuBar");
-    ezCommandHistoryActions::MapActions("SoundBankAssetMenuBar");
+    ezActionMapManager::RegisterActionMap("SoundBankAssetMenuBar", "AssetMenuBar").IgnoreResult();
```
jankrassnigg: Is there a point to returning ezResult here, if we always ignore it anyway ?

Code/Tools/Libs/GuiFoundation/Action/ActionMapManager.h
```diff
@@ -8,7 +8,7 @@ class EZ_GUIFOUNDATION_DLL ezActionMapManager
 {
 public:
   /// \brief Adds a new action map with the given name. Returns EZ_FAILURE if the name was already used before.
-  static ezResult RegisterActionMap(ezStringView sMapping);
+  static ezResult RegisterActionMap(ezStringView sMapping, ezStringView sParentMapping = {});
```
jankrassnigg: Needs a better comment to explain this.

Code/Tools/Libs/GuiFoundation/Action/Implementation/ActionMap.cpp
```diff
@@ -266,3 +297,53 @@ const ezTreeNode<ezActionMapDescriptor>* ezActionMap::GetChildByName(const ezTre
...
+  if (uiCurrentTransitiveEditCounter == m_uiTransitiveEditCounterOfRoot)
+  {
+    return &m_Root;
+  }
+  m_uiTransitiveEditCounterOfRoot = uiCurrentTransitiveEditCounter;
+
+  m_Root = TreeNode();
+  m_Descriptors.Clear();
+
+  for (ezInt32 i = (ezInt32)mappings.GetCount() - 1; i >= 0; --i)
```
jankrassnigg, line 324: When I need to loop backwards with uint, I often do this:

```cpp
for (ezUInt32 iPlus1 = arr.GetCount(); iPlus1 > 0; --iPlus1)
{
  const ezUInt32 i = iPlus1 - 1;
}
```

jankrassnigg (approved):
This looks useful, definitely gets rid of a lot of copy&paste code which makes it more maintainable. I think the interface is fine.

Code/EditorPlugins/Fmod/EditorPluginFmod/EditorPluginFmod.cpp
Sanakan8472: No I don't think so, so I removed it and replaced with asserts for both register and unregister functions.

Code/Tools/Libs/GuiFoundation/Action/ActionMapManager.h
Sanakan8472: Done!

Code/Tools/Libs/GuiFoundation/Action/Implementation/ActionMap.cpp
Sanakan8472: I would do that if I absolutely require an unsigned value. If not, I'd rather have one line less and the benefit of not accidentally picking the wrong counter variable in my code inside the loop.

PR #1317, opened 2024-06-04
Some ezFoundation changes

jankrassnigg:
* Tweaks to ezFileSystem API
* Minor STL compat fixes
* Some math utility functions
* Removed unused pointer offset in debug builds
* constexpr ezUuid
* StringUtils: CalculateLineNumber

### DIFF
--- a/Code/Engine/Foundation/Math/Transform.h
+++ b/Code/Engine/Foundation/Math/Transform.h
@@ -74,7 +74,7 @@ public:
   Type GetMaxScale() const;
 
   /// \brief Returns whether this transform contains negative scaling aka mirroring.
-  bool ContainsNegativeScale() const;
+  bool HasMirrorScaling() const;
 
   /// \brief Returns whether this transform contains uniform scaling.
   bool ContainsUniformScale() const;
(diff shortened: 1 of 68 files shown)


Code/Engine/Foundation/Math/Transform.h
```diff
@@ -82,6 +82,9 @@ class ezTransformTemplate
   /// \brief Checks that all components are valid (no NaN, only finite numbers).
   bool IsValid() const;
 
+  /// \brief Check if scale is flipped.
+  bool IsFlipped() const;
```
C-Core: We already have a function like that on ezSimdTransform where it is called "ContainsNegativeScale" which is more decriptive I think. Would be could if the api stays consistent.

jankrassnigg: Though isn't that name incorrect, as well? If it has two negative scales, it would return false, even though the name suggests it would return true.

jankrassnigg: I'd be happy to change both names to something better, but can't think of a good name. Maybe "HasMirrorScaling" ?

Also, actually ezTransform already has ContainsNegativeScale() and IsFlipped can just be removed.

C-Core: HasMirrorScaling or ContainsMirrorScaling sounds good to me.

PR #1319, opened 2024-06-05
Generalized support for injecting textures into a render pipeline

Sanakan8472:
Previously, the only input to a render pipeline was the `ezTargetPass` which was hardcoded in the `ezRenderPipeline`. The concept has now been generalized with these changes:
1. New node pins `ezRenderPipelineNodeOutputProviderPin` and `ezRenderPipelineNodeInputProviderPin` have been added.
2. When the new pin type is present, the render pipeline asks the node for a texture to be provided for this pin by calling `ezRenderPipelinePass::QueryTextureProvider` with the pin and texture description that is expected.
3. If the pass returns a valid texture, this texture is used instead for the connection instead of using the texture pool. In case an invalid texture is returned, the pin reverts to the normal input / output variant.
4. The hard-coded logic has been moved from the render pipeline to the `ezTargetPass::QueryTextureProvider` implementation.

## ezHistorySourcePass / ezHistoryTargetPass
New render passes have been added that built on top of the new feature and allows to access data from a previous frame. To preserve textures across the next frame you need to create a `ezHistorySourcePass` node to define the type of texture and initial state. This node's output pin will give access to the previous frame's content. 

Next, create an `ezHistoryTargetPass`. It's input pin exposes the same texture as provided by the source node but allows you to write to by connecting the input pin to another pass that produces the image that you want to carry to the next frame. To connect an ezHistoryTargetPass to its counterpart you need to set it's "SourcePassName" property to the name of the `ezHistorySourcePass` you want to match. Note that this requires giving the `ezHistorySourcePass` an explicit name, the default type name shown on the node is not valid. 

As both nodes expose the same texture, special care has to be taken that it's not used as input and output of another pass at the same time. In those cases, add a `ezCopyTexturePass` to break up invalid state. Here is an example of a basic feedback loop of the color buffer across frames that requires a copy to be inserted:
image
Without the copy, the blend pass would read and write to the same texture.

## ezBlendPass
Added a very simple blend pass that blends the two input pins by an exposed BlendFactor.

## Misc
* Fixed typo in `ezRenderPipelineNodePassThroug[h]Pin`. While the class is reflected, it is not relevant during serialization as it never is referenced by others or hold any properties, so no patch needs to be applied.
* Fixed Vulkan Copy Texture going out of bounds when copying textures with mip maps.
* Fixed clang compilation by adding `<algorithm>` include before `<Texture/DirectXTex/BC.h>`.


Data/Base/Shaders/Pipeline/Blend.ezShader
```diff
@@ -0,0 +1,45 @@
...
+#include <Shaders/Pipeline/FullscreenTriangleVertexShader.h>
+
+[GEOMETRYSHADER]
+
+#include <Shaders/Pipeline/FullscreenTriangleStereoGeometryShader.h>
+
+[PIXELSHADER]
+
+#include <Shaders/Pipeline/FullscreenTriangleInterpolator.h>
+#include <Shaders/Common/Common.h>
```
C-Core, line 27: Most of the includes are not necessary I guess besides FullscreenTriangleInterpolator.h and BlendConstants.h, can you remove the rest?

Code/Engine/RendererVulkan/CommandEncoder/Implementation/CommandEncoderImplVulkan.cpp
```diff
@@ -347,9 +347,7 @@ void ezGALCommandEncoderImplVulkan::CopyTexturePlatform(const ezGALTexture* pDes
     imageCopy.dstSubresource.baseArrayLayer = 0;
     imageCopy.dstSubresource.layerCount = destDesc.m_uiArraySize;
     imageCopy.dstSubresource.mipLevel = i;
-    imageCopy.extent.width = destDesc.m_uiWidth;
-    imageCopy.extent.height = destDesc.m_uiHeight;
-    imageCopy.extent.depth = destDesc.m_uiDepth;
+    imageCopy.extent = destination->GetMipLevelSize(i);
```
C-Core, line 348: Is the // TODO above still relevant?

Code/Engine/RendererCore/Pipeline/Passes/SourcePass.h
```diff
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "RendererCore/Pipeline/FrameDataProvider.h"
```
C-Core: This include shouldn't be necessary?

Code/Engine/RendererCore/Pipeline/Implementation/Passes/SourcePass.cpp
```diff
@@ -41,14 +41,13 @@ ezSourcePass::ezSourcePass(const char* szName)
 {
   m_Format = ezSourceFormat::Default;
```
C-Core: Nit: move initialization to header

Code/Engine/RendererCore/Pipeline/Implementation/Passes/BlendPass.cpp
```diff
@@ -0,0 +1,114 @@
...
+  return true;
+}
+
+void ezBlendPass::Execute(const ezRenderViewContext& renderViewContext, const ezArrayPtr<ezRenderPipelinePassConnection* const> inputs, const ezArrayPtr<ezRenderPipelinePassConnection* const> outputs)
+{
+  if (outputs[m_PinOutput.m_uiOutputIndex])
+  {
+    ezGALDevice* pDevice = ezGALDevice::GetDefaultDevice();
+
+    ezBlendConstants* cb = ezRenderContext::GetConstantBufferData<ezBlendConstants>(m_hBlendCB);
```
C-Core: Would be a nice use case for push constants

C-Core (approved):

Code/Engine/RendererCore/Pipeline/Implementation/Passes/BlendPass.cpp
```diff
@@ -0,0 +1,114 @@
...
+ezResult ezBlendPass::Deserialize(ezStreamReader& inout_stream)
+{
+  EZ_SUCCEED_OR_RETURN(SUPER::Deserialize(inout_stream));
+  const ezUInt32 uiVersion = ezTypeVersionReadContext::GetContext()->GetTypeVersion(GetStaticRTTI());
+  EZ_IGNORE_UNUSED(uiVersion);
+  inout_stream >> m_fBlendFactor;
+  return EZ_SUCCESS;
+}
+
+EZ_STATICLINK_FILE(RendererCore, RendererCore_Pipeline_Implementation_Passes_BlurPass);
```
C-Core: Copy paste from blur pass

Data/Base/Shaders/Pipeline/Blend.ezShader
Sanakan8472: Fixed! Copy pasted out of laziness.

Code/Engine/RendererVulkan/CommandEncoder/Implementation/CommandEncoderImplVulkan.cpp
Sanakan8472: Ah see I know something was missing all along ;-)

Code/Engine/RendererCore/Pipeline/Implementation/Passes/SourcePass.cpp
Sanakan8472: Damn copy&paste laziness strikes again!

Code/Engine/RendererCore/Pipeline/Implementation/Passes/BlendPass.cpp
Sanakan8472: Woops, ran static link tool over it to fix.

Code/Engine/RendererCore/Pipeline/Implementation/Passes/BlendPass.cpp
Sanakan8472: Good point! Fixed!

Issue #1149, opened 2024-01-09
[BUG] Shortcuts (copy / paste) broken in Visual Shader editor (material assets)


jankrassnigg:
Shortcuts don't work in the visual shader editor window anymore. Probably a similiar problem to this https://github.com/ezEngine/ezEngine/pull/1112

PR #1321, opened 2024-06-06
Fix shortcuts (e.g. Ctrl+S) while focus is inside the Visual Shader Editor

RiederAlex:
Fix shortcuts (e.g. Ctrl+S) while focus is inside the Visual Shader Editor

This issue only occurred for the Visual Shader Editor because of the workaround present for centralWidgets in ezQtDocumentWindow::eventFilter, but the Visual Shader Editor is not a centralWidget.

How to test:
Open Material, do some changes and while focus is inside Visual Shader Editor, press Ctrl+S.
With this change, the asset should now be saved.

### DIFF
--- a/Code/Tools/Libs/GuiFoundation/NodeEditor/Implementation/NodeScene.cpp
+++ b/Code/Tools/Libs/GuiFoundation/NodeEditor/Implementation/NodeScene.cpp
@@ -379,6 +379,12 @@ void ezQtNodeScene::keyPressEvent(QKeyEvent* event)
   {
     OpenSearchMenu(QCursor::pos());
   }
+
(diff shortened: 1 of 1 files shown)


jankrassnigg:
You're my hero, I've been annoyed by this for quite a while now.
See this ticket: https://github.com/ezEngine/ezEngine/issues/1149

The referenced commit solves this issue in a different way, though, not sure which one is better. Maybe something for @Sanakan8472 to comment on.

Sanakan8472:
Looks good to me! Calling `QGraphicsScene::keyPressEvent(event);` would work as well as the default behavior of any widget is to ignore any event but I think the graphics view has some undesirable effects so we explicitly don't call the base impl.

PR #1323, opened 2024-06-06
Add experimental support for DXVK-native on Linux

Sanakan8472:
This allows compiling the DX11 renderer on Linux. The main goal is to make it easier to make renderer changes on Linux without having to boot Windows. To use:
1. Clone DXVK
1. Change `meson.build` to `lib_glfw = dependency('glfw3', required: false)` instead of just `'glfw'`.
1. Run `./package-native.sh master ~/libDXVK --no-package`
1. `export PKG_CONFIG_PATH="~/libDXVK/dxvk-native-master/usr/lib/pkgconfig"`
1. Enable `EZ_BUILD_EXPERIMENTAL_DXVK`

Note that you won't be able to run this without considerable effort so right now it's merely to make sure the code compiles:
1. There is no shader compiler so if you want to run DX11, you need to get a shader cache from a Windows machine or from the Windows CI build artifacts.
3. We conflict between two GLFW versions as EZ does not use the system version but DXVK does. You can delete our version and replace it with the system version but it's tedious.
4. For some reason you can't load two renderers at the same time under Linux. Sub-system init for DX11 is simply skipped. Can't figure out why.
5. Latest release of dxvk has a bug, so you need to apply at least this patch: https://github.com/doitsujin/dxvk/pull/4061
6. You need to set the env var `DXVK_WSI_DRIVER=GLFW` before running any app.

### DIFF
--- a/Code/BuildSystem/AzurePipelines/Windows-x64.yml
+++ b/Code/BuildSystem/AzurePipelines/Windows-x64.yml
@@ -200,4 +200,10 @@ jobs:
     inputs:
       PathtoPublish: $(Build.ArtifactStagingDirectory)\AssetCache
       ArtifactName: AssetCache
+  - task: PublishBuildArtifacts@1
+    displayName: 'Publish Artifact: ShaderCache'
+    condition: eq(variables['task.MSBuild.status'], 'success')
+    inputs:
+      PathtoPublish: $(System.DefaultWorkingDirectory)\Output\ShaderCache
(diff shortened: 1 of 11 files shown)


jankrassnigg (commented):
Sounds like great fun.

Sanakan8472:
> Sounds like great fun.

Anything is more fun than having to boot windows all the time...

PR #1331, opened 2024-06-10
Fixed wrong thresholds in placement patterns

C-Core:
Previous patterns contained zero thresholds which is wrong because it would place objects even if the density is zero.


jankrassnigg (approved):
9314d601-8547-42ed-a9d9-c8e49832835b_text

C-Core:
If you want to see pretty points take a look at the source website ;)
image

PR #1332, opened 2024-06-14
MSFT Contributions June 2024

C-Core:
* Renamed std_image library to just stb and added std_rect_pack
* Changed ezTexturePacker implemention to use std_rect_pack to allow for runtime use cases. The previous implementation was to slow for that.
* Smaller additions and fixes


Code/Engine/Foundation/IO/FileSystem/Implementation/FileSystem.cpp
```diff
@@ -30,6 +30,8 @@ ezMap<ezString, ezString> ezFileSystem::s_SpecialDirectories;
 
 void ezFileSystem::RegisterDataDirectoryFactory(ezDataDirFactory factory, float fPriority /*= 0*/)
 {
+  EZ_ASSERT_DEV(s_pData != nullptr, "FileSystem is not initialized.");
```
jankrassnigg, line 35: This will only happen if someone generally didn't initialize ez yet. The nullptr check is quite redundant, because the next line will literally do a nullptr access and crash with the appropriate error message from the compiler.

I think this should at most be an EZ_ASSERT_DEBUG.

jankrassnigg (approved):

Code/Engine/Foundation/IO/FileSystem/Implementation/FileSystem.cpp
C-Core: In some projects the lifetime of ez can be more complicated. These asserts have been added to catch lifetime issues on CI machines. Asserts are much easier to see in logs than crashes and we typically have dev builds on CI. 
That's the background and I think these asserts don't hurt but we can also keep this change internal if you would prefer that.

PR #1336, opened 2024-06-20
Fixed file system iterator, particle orientation and added reverse string hash lookup

jankrassnigg:


Data/Base/Shaders/Particles/ParticleCommonVS.h
```diff
@@ -60,14 +60,17 @@ Quad CalcQuadOutputPositionWithAlignedAxis(uint vertexIndex, float3 inPosition,
...
   float3 offsetRight = orthoDir * ((QuadTexCoords[vertexIndex].x - 0.5) * inSize);
   float3 offsetUp = axisDir * ((1.0 - QuadTexCoords[vertexIndex].y) * inSize * stretch);
 
   Quad quad;
-  quad.worldPosition = mul(ObjectToWorldMatrix, float4(inPosition + offsetRight + offsetUp, 1));
+  quad.worldPosition = mul(ObjectToWorldMatrix, float4(inPosition, 1));
+  quad.worldPosition.xyz += offsetRight + offsetUp;
   quad.screenPosition = mul(GetWorldToScreenMatrix(), quad.worldPosition);
 
   float3 centerNormal = cross(axisDir, orthoDir);
```
C-Core, line 76: The normal is now wrong since axisDir and orthoDir and the offsets are already in world space.

C-Core (changes requested):

Data/Base/Shaders/Particles/ParticleCommonVS.h
jankrassnigg: You mean 'centerNormal' ? I don't see that.

As far as I can tell, the code originally assumed that axisDir and orthoDir are in world-space, but they only where that, when the effect wasn't rotated (therefore it always looked right in the particle editor).

Now they are always in world space and I think the rest of the calculation doesn't need to change.

C-Core: The normals are transformed into world space 2 lines below that. Github doesn't allow me to comment on that line :(
Now the normals are double transformed.

jankrassnigg: Good catch. Fixed it.

C-Core (approved):

PR #1338, opened 2024-06-23
Support ezDirectoryWatcher on non NTFS drives

Sanakan8472:
Several changes were necessary to make this work:
1. The logic of enumerating the changes has been separated from processing the changes. `ezDirectoryWatcher::EnumerateChanges` now only does the processing and the OS enumeration is done by `ezDirectoryWatcherImpl::EnumerateChangesImpl`.
2. `ezDirectoryWatcher::OpenDirectory` now checks the file system of the root of the drive via `GetVolumeInformationW`. Depending on whether NTFS is supported, `m_bNTFS` is set. If set, `ReadDirectoryChangesExW` is used as before. If not, `ReadDirectoryChangesW` is used instead.
3. Depending on `m_bNTFS`, `ezDirectoryWatcherImpl::EnumerateChangesImpl` will call either `GetChangesNTFS` and process those directly or call `GetChangesNonNTFS` in a loop until all changes are collected and then call `PostProcessNonNTFSChanges`.
4. `PostProcessNonNTFSChanges` main goal is to recover the type of the change (file ord directory). This can't be done perfectly, but it uses the following heursitics:
    1. First, a linked list of changes is built that belong to the same object (e.g. create, rename, rename again, modify etc.).
    2. For each object we itterate through all its changes and try to detect its type.
    3. Objects that have a state stored in the mirror can be given a type with 100% acuracy.
    4. Objects that still exit on disk can get their type by calling `ezOSFile::GetFileStats`. This is almost always correct except for cases that shouldn't appear in practice.
    5. Objects that got created and deleted in the same enumeration call can't be figured out, so we guess by looking for a `.` in the file and if present determine that it probably was a file.
5. A unit test (Windows only) has been added that forces logic for non-NTFS drives. All current Directory Watcher tests succeed.

### DIFF
--- a/Code/Engine/Foundation/Platform/Win/DirectoryWatcher_Win.cpp
+++ b/Code/Engine/Foundation/Platform/Win/DirectoryWatcher_Win.cpp
@@ -3,6 +3,7 @@
 #if (EZ_ENABLED(EZ_PLATFORM_WINDOWS_DESKTOP) && EZ_ENABLED(EZ_SUPPORTS_DIRECTORY_WATCHER))
 
 #  include <Foundation/Basics/Platform/Win/IncludeWindows.h>
+#  include <Foundation/Configuration/CVar.h>
 #  include <Foundation/Containers/DynamicArray.h>
 #  include <Foundation/IO/DirectoryWatcher.h>
 #  include <Foundation/IO/Implementation/FileSystemMirror.h>
@@ -20,6 +21,8 @@
 
 namespace
 {
+  ezCVarBool cvar_ForceNonNTFS("DirectoryWatcher.ForceNonNTFS", false, ezCVarFlags::Default, "Forces the use of ReadDirectoryChanges instead of ReadDirectoryChangesEx");
+
   struct MoveEvent
   {
     ezString path;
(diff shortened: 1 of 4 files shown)


Code/Engine/Foundation/Platform/Win/DirectoryWatcher_Win.cpp
```diff
@@ -120,16 +382,79 @@ void ezDirectoryWatcherImpl::DoRead()
...
+  {
+    BOOL success =
+      ReadDirectoryChangesW(m_directoryHandle, m_buffer.GetData(), m_buffer.GetCount(), m_whatToWatch.IsSet(ezDirectoryWatcher::Watch::Subdirectories), m_filter, nullptr, &m_overlapped, nullptr);
+    EZ_ASSERT_DEV(success, "ReadDirectoryChangesW failed.");
+  }
+}
+
+void ezDirectoryWatcherImpl::EnumerateChangesImpl(ezStringView sDirectoryPath, ezTime waitUpTo, const ezDelegate<void(const Change&)>& callback)
+{
+  ezDynamicArray<Change> changes;
```
jankrassnigg: hybrid array ?

Code/Engine/Foundation/Platform/Win/DirectoryWatcher_Win.cpp
```diff
@@ -36,13 +39,246 @@ namespace
...
+      ezMap<ezStringView, ezUInt32> lastChangeAtPath;
+      nextOp.SetCount(ref_changes.GetCount(), -1);
+
+      ezInt32 pendingRemoveOrRename = -1;
+      ezInt32 lastMoveFrom = -1;
+
+      for (ezUInt32 i = 0; i < ref_changes.GetCount(); i++)
+      {
+        const auto& currentChange = ref_changes[i];
+        if (pendingRemoveOrRename != -1 && currentChange.Action == FILE_ACTION_RENAMED_OLD_NAME && ref_changes[pendingRemoveOrRename].eventFilePath == currentChange.eventFilePath)
```
jankrassnigg: more parenthesis

Code/Engine/Foundation/Platform/Win/DirectoryWatcher_Win.cpp
```diff
@@ -36,13 +39,246 @@ namespace
...
+            lastChangeAtPath.Remove(ref_changes[lastMoveFrom].eventFilePath);
+            lastChangeAtPath.Insert(currentChange.eventFilePath, i);
+            break;
+        }
+      }
+    }
+
+    // Anything that is chained via the nextOp linked list must be given the same type.
+    // Instead of building arrays of arrays, we create a bit field of all changes and then flatten the linked list at the first set bit. While iterating we remove everything we reached via the linked list so on the next call to get the first bit we will find another object that needs processing. As the operations are ordered, the first bit will always point to the very first operation of an object (nextOp can never point to a previous element).
+    ezBitfield<ezHybridArray<ezUInt32, 4>> pendingChanges;
```
jankrassnigg: ezHybridBitfield<128>

Code/Engine/Foundation/Platform/Win/DirectoryWatcher_Win.cpp
```diff
@@ -36,13 +39,246 @@ namespace
...
+        bool isFile = true;
+        bool typeFound = false;
+        for (Change* currentChange : objectChanges)
+        {
+          ezFileSystemMirrorType::Type type;
+          if (pMirror->GetType(currentChange->eventFilePath, type).Succeeded())
+          {
+            isFile = type == ezFileSystemMirrorType::Type::File;
+            typeFound = true;
+            break;
```
jankrassnigg: goto typeFound; 😄

Code/Engine/Foundation/Platform/Win/DirectoryWatcher_Win.cpp
```diff
@@ -36,13 +39,246 @@ namespace
...
+        {
+          // No stats and no entry in mirror: It's guessing time!
+          isFile = objectChanges[0]->eventFilePath.FindSubString(".") != nullptr;
+        }
+
+        // Apply type to all objects in the chain.
+        for (Change* currentChange : objectChanges)
+        {
+          currentChange->isFile = isFile;
+        }
```
jankrassnigg, line 266: I don't understand the chain logic yet, but wouldn't you want to apply the change differently, depending on how confident you are ?

Ie when we know the result, apply it, if we don't know it and just guessed, store it, but wait for a better result? And don't override a result that was already certain?

jankrassnigg (approved):
I tried it on my dev drive and it works now 👍

Code/Engine/Foundation/Platform/Win/DirectoryWatcher_Win.cpp
Sanakan8472: So a chain of events is something like:
1. "Create A -> Rename start A -> Rename end B -> Modify B". We know that all of these must refer to the same object but just looking at the first event "Create A" we can't figure out the type given that at the location 'A' no object exists anymore on disk and we have obviously not observed A in the mirror at that point as it is new.
If we can find a file at location B, then we know the entire chain must belong to a file. The only cases where this assumption is wrong are non-practical like creating a file and immediately deleting it and replacing it with a folder of the same name which will not appear in practise.
2. Another common change is something like "Modify C -> Modify C" or "Rename C -> Rename D". In these cases we have observed the object before so the very first op can be looked up in the mirror. Here we can always be certain that this is correct.
3. The only case that we have to guess is something like this: "Create A -> Delete A". Or cases where the monitor is overwhelmed and the kernel drops events.

As for waiting longer, I already flush all events that are in the kernel by running the code after the outer loop so no more events can be observed for these objects at the current time. So waiting any longer would probably not improve things are the state on disk will diverge even further from the observed events over time. And the question then also is when to stop and make a decision.

I'm not sure if there is a case in which waiting longer would result in a different decision. The file won't suddenly appear in the monitor and if we can't find it on disk anymore it will probably not come back anytime soon. There is of course always the possibility of race conditions between the ops on disk and the events in the monitor but the only case in which you would be able to trick the logic here would be if you rename a file multiple times in rapid succession so that the evaluation of the monitor results will not be able to observe the current location on disk as the file has already been renamed again, which is probably an unlikely case in practice.

The only way to improve this is probably making up heuristics based on the exact type of events in the chain but that would make the code a lot more complicated.
Before making any more detailed changes to the logic, I'd say we should wait for practical examples that we observe where things don't work as expected and then come up with better heuristics for these use cases.

Code/Engine/Foundation/Platform/Win/DirectoryWatcher_Win.cpp
Sanakan8472: That would work as well, but I'm old fashioned and have a goto phobia (or rather I forgot how it works).

Code/Engine/Foundation/Platform/Win/DirectoryWatcher_Win.cpp
jankrassnigg: Okay makes sense.

Though I didn't mean "wait longer" like actually letting more time pass.

To me the code looks like it makes a decision based on different criteria, one of them is the fallback to look at the filename and check for dots. However, you have multiple events in your chain and one may give you a very certain answer, the other only the guessed answer, but to me it looks like you basically just take what you get and don't store how confident you are about it.

Ie if the first event gave you a perfect answer, and the second one is one where you have to guess, you shouldn't override the answer. The other way round, you SHOULD override the answer, because the first one wasn't as confident.

Probably your code is doing the right thing and I just don't see it, but that was what I meant with "wait for a better result".

Sanakan8472: Ah I see what you mean, so this is happening implicitly as in the loop the mirror condition is checked first. I've changed the code now to make that clearer: Only the very first element of the chain needs to be looked up in the mirror as this is a list of ordered operations. If it's in the mirror, we are done. If not, we check all changes against the disk and if that also doesn't yield any results, we fall back to guessing.

PR #1341, opened 2024-06-29
GPU -> CPU query functions

Sanakan8472:
Timestamps, occlusion queries and fences all use the same pattern now: inserting creates a handle which can then be queried on the device for results. Each result method returns `ezGALAsyncResult` which can be `Ready`, `Pending` or `Expired`. Results are only cached for up to 4 frames after which they will be disposed of and `Expired` will be returned. This was already the case before for timestamps, but the state couldn't be queried.

## Timestamps
* The implementation details in DX11 have been moved into the new `ezQueryPoolDX11` class.

## Occlusion Queries
* The `ezGALQuery` class has been removed. Instead, occlusion queries are now just handles named `ezGALOcclusionHandle`.
* `ezGALCommandEncoder::BeginOcclusionQuery` returns a handle that has to be passed into `ezGALCommandEncoder::EndOcclusionQuery` within the same render scope.
* The result can be retrieved via `ezGALDevice::GetOcclusionResult`.

## Fences
* Fences can be inserted into a command encoder via `ezGALCommandEncoder::InsertFence`. An `ezGALFenceHandle` is returned that can be queried via `ezGALDevice::GetFenceResult`.
* `GetFenceResult` supports a timeout if you need to wait for a fence instead of just querying the current state. Note that you need to flush commands to the GPU in order to be able to wait for a fence via either ending a frame or calling `ezCommandEncoder::Flush` explicitly.
* Fence logic is implemented via the new classes `ezFenceQueueVulkan` and `ezFenceQueueDX11`.
* The DX11 fence logic has been moved from the device into `ezFencePoolDX11`.

## Safe Frames
* Added `ezGALDevice::GetCurrentFrame`: This is a monotonically increasing number which changes +1 every time EndFrame is called. You can use this to synchronize read/writes between CPU and GPU, see GetSafeFrame.
* Added `ezGALDevice::GetSafeFrame`: The latest frame that has been fully executed on the GPU. Whenever you execute any work that requires synchronization between CPU and GPU, remember the `GetCurrentFrame` result in which the operation was done. When `GetSafeFrame` reaches this number, you know for sure that the GPU has completed all operations of that frame.
* To make this work, each device implementation will now in `BeginFramePlatform` query all pending frame fences and do any cleanup related to those. Previously, this was only done after 4 frames.

## Misc Changes

* `ezCommandEncoder::Flush` now ensures that all commands are submitted to the GPU. The function was previously just flushing internal state to the underlying API.
* Added support for 64-bit-based `ezGenericId`.

## Known Issues
* On Vulkan, occlusion queries will fail validation on the first draw call in a render scope as the render pass is started lazily and thus isn't present then the begin call is executed. This will be fixed once the lazy evaluation has been removed from Vulkan.
* The DX11 query pools are ring buffers of a fixed size which can potentially case an overrun. This was already the case for timestamps but the same pool implementation has now been applied to occlusion queries as well. You can change `ezQueryPoolDX11::Initialize` to increase the max limit (2k for timestamps, 512 for occlusion queries).


Code/Engine/RendererFoundation/Profiling/Implementation/Profiling.cpp
```diff
@@ -30,13 +30,18 @@ class GPUProfilingSystem
...
+
+      if (resBegin == ezGALAsyncResult::Expired || resEnd == ezGALAsyncResult::Expired)
       {
-        ezTime beginTime;
-        EZ_VERIFY(e.m_pDevice->GetTimestampResult(timingScope.m_BeginTimestamp, beginTime).Succeeded(),
-          "Begin timestamp should be finished before end timestamp");
+        s_TimingScopes.PopFront();
+      }
 
+      if (resBegin == ezGALAsyncResult::Ready || resEnd == ezGALAsyncResult::Ready)
```
C-Core: Shouldn't that be resBegin == ezGALAsyncResult::Ready && resEnd == ezGALAsyncResult::Ready

C-Core (approved):

Code/Engine/RendererFoundation/Profiling/Implementation/Profiling.cpp
Sanakan8472: Yes totally!

PR #1343, opened 2024-07-05
Various improvements for AssetBrowser related to Show/Hide all children

RiederAlex:
- Icon now has correct color/positioning
- <root> folder no longer shows all items all the time
- "Select in AssetBrowser" now clears the Search text, sets the path filter and always selected the asset correctly
- "Show Items in hidden folders" (right click context menu in tree view) can now always be toggled on/off
- Fixed issues with AssetBrowser in Dialog mode


jankrassnigg (approved):
Not sure about switching the default for showing the items in the folders, but we can give it a try for a while.

PR #1350, opened 2024-07-13
FMOD, Project Export and other fixes

jankrassnigg:


Code/EnginePlugins/FmodPlugin/Components/FmodEventComponent.cpp
```diff
@@ -770,8 +790,17 @@ void ezFmodEventComponent::UpdateParameters(FMOD::Studio::EventInstance* pInstan
...
-  // have to update pitch every time, in case the clock speed changes
-  EZ_FMOD_ASSERT(pInstance->setPitch(m_fPitch * (float)GetWorld()->GetClock().GetSpeed()));
+  if (GetNoGlobalPitch())
+  {
+    EZ_FMOD_ASSERT(pInstance->setPitch(m_fPitch));
+  }
+  else
+  {
+    // have to update pitch every time, in case the clock speed changes
+    /// \todo Global pitch might better be a bus setting
```
C-Core: There is no bus setting for pitch. You can change the frequency of a sound with a pitch shifter effect on a bus but it does not change the playback speed, which makes sense since a bus is basically just a post processing chain. So this comment can be removed. (It appears multiple times in this file)

C-Core (approved):

PR #1351, opened 2024-07-18
Sort log-messages before copy pasting

RiederAlex:
This ensures that the C&P log will always be from top to bottom, no matter in what order the selection happened inside the log view.

e.g. select log entries in a random order via Ctrl+Left Click and copy&paste into a text editor. 
Previously the order would be preserved (which is unintuitive), now the log messages are always from olders to newest


jankrassnigg (approved):
Never noticed this, but thanks for fixing !

PR #1356, opened 2024-07-23
Fixed crash in JoltDefaultCharacterComponent

RiederAlex:
Fixed crash inJoltDefaultCharacterComponent when m_vVelocityLateral is nearly zero, but not quite, causing NaN values.

image

This actually happened pretty easily when jumping against slightly slanted walls, see attached video.
Without the fix this crashes nearly 100% of the time

https://github.com/user-attachments/assets/6bf4f449-28b1-4b3c-96c2-e515232e30a4


jankrassnigg (approved):
Thanks!

PR #1360, opened 2024-07-29
Fix compilation error when building with EZ_3RDPARTY_ZSTD_SUPPORT=OFF

x0r04rg:
uiMaxNumWorkerThreads emits 'unused variable' warning which gets treated as an error


jankrassnigg (approved):
👍

PR #1363, opened 2024-08-09
Added component to trigger scene transitions

jankrassnigg:
Adds a component that makes it easy to switch between levels.

`ezSceneTransitionComponent` is a component that either reacts to trigger events, or can be invoked via script to make the game state load into another scene.
It can also trigger (or cancel) a preload, and forward relative player positions to the new scene, to make it possible to build levels with very natural transitions. You can also select at which start point a player should spawn, in case the target level has multiple.

There is a new demo scene that shows different possibilities. The video below shows the setup for preloading a scene with one trigger and then entering the scene through one or another entry point.

https://github.com/user-attachments/assets/f1da1767-1cf6-4ccf-84c6-0820ea1644b1


Code/Engine/Core/GameApplication/Implementation/GameApplicationBase.cpp
```diff
@@ -213,15 +213,15 @@ void ezGameApplicationBase::ExecuteFrameCapture(ezWindowHandle targetWindowHandl
 
 //////////////////////////////////////////////////////////////////////////
 
-void ezGameApplicationBase::ActivateGameState(ezWorld* pWorld /*= nullptr*/, ezStringView sStartPosition /*= {}*/, const ezTransform* pStartPosition /*= nullptr*/)
+void ezGameApplicationBase::ActivateGameState(ezWorld* pWorld /*= nullptr*/, ezStringView sStartPosition /*= {}*/, const ezTransform& startPositionOffset /*= nullptr*/)
```
C-Core: Function has no default values anymore, remove comments

Code/Engine/GameEngine/GameApplication/Implementation/GameApplicationInit.cpp
```diff
@@ -69,6 +69,8 @@ void ezGameApplication::Init_ConfigureAssetManagement()
   ezResourceManager::RegisterResourceForAssetType("Surface", ezGetStaticRTTI<ezSurfaceResource>());
   ezResourceManager::RegisterResourceForAssetType("Texture 2D", ezGetStaticRTTI<ezTexture2DResource>());
   ezResourceManager::RegisterResourceForAssetType("Texture Cube", ezGetStaticRTTI<ezTextureCubeResource>());
+  // ezResourceManager::RegisterResourceForAssetType("Scene", nullptr);
```
C-Core: Can be removed?

Code/Engine/GameEngine/GameState/Implementation/GameState.cpp
```diff
@@ -296,38 +312,61 @@ ezResult ezGameState::SpawnPlayer(ezStringView sStartPosition, const ezTransform
...
+      }
 
-        if (pStartPosition)
-        {
-          startPos = *pStartPosition;
-          startPos.m_vScale.Set(1.0f);
-          startPos.m_vPosition.z += 1.0f; // do not spawn player prefabs on the ground, they may not have their origin there
-        }
+      startPos.m_vScale.Set(1.0f);
+      // startPos.m_vPosition.z += 0.1f; // do not spawn player prefabs on the ground, they may not have their origin there
```
C-Core: remove commented out code?

Code/Engine/GameEngine/Gameplay/SceneTransitionComponent.h
```diff
@@ -0,0 +1,99 @@
+#pragma once
+
+#include <Core/World/EventMessageHandlerComponent.h>
+#include <GameEngine/GameEngineDLL.h>
+
+struct ezMsgTriggerTriggered;
+
+using ezSceneTransitionComponentManager = ezComponentManager<class ezSceneTransitionComponent, ezBlockStorageType::Compact>;
+
+/// \brief What ezSceneTransitionComponent should do when it gets triggered through a ezMsgTriggerTriggered
```
C-Core, line 10: Would it make more sense to have one scene transition component that can handle both the preload and scene switch. The trigger messages can have a string to differentiate what each trigger should do. The expected trigger messages could be exposed as properties on this component with a default value so they can be customized but don't have to.
I think that would make it easier to setup and keep all the logic in one place.

jankrassnigg: Hm, maybe. I mean you can already do that now, if you add a visual script. I think the current setup is the easiest to use and understand. Not sure yet, it would be easy to change, though.

PR #1375, opened 2024-08-20
Fixed #1280: Added tracking of missing package dependencies

### LINKED ISSUE #1280: [BUG] Missing sub-assets are not detected by asset transform
Create a prefab that contains a Sound Event component, set a sound event on it.
Open FMOD Studio, remove that sound event and re-build the sound banks (F7).
Now the sound event asset doesn't show up in the ezEditor anymore.
Asset transform, however, doesn't complain, the missing reference doesn't show up in the Asset Curator. You have to manually open the prefab to see that the asset dependency can't be found.

jankrassnigg:
This is a fix for #1280, but in practice this didn't only affect sub-assets, but really all package dependencies.

So far we completely ignored looking at package dependencies.

They are technically not that important, because they are already covered by the regular asset file hash, since they are just strings that point to another asset, but they are not needed for transforming assets, and if they aren't also tagged as thumbnail dependencies, then they are also not needed for that.

Unfortunately, this meant that if an asset was deleted, the asset curator would not point this missing dependency out. Which makes it really, really hard to find assets that have a missing dependeny (which may crash at runtime).

So I've now added that we also track this, similar to the thumbnail dependencies.

This does mean, that an asset gets retransformed when a missing package dependency is fixed, even though technically this isn't needed. Well, at least if the editor was already open. It may not get retransformed, if the editor was not open while the missing asset gets restored.

Not 100% certain this is "the right fix" for this, though.

### DIFF
--- a/Code/Editor/EditorFramework/Assets/AssetCurator.h
+++ b/Code/Editor/EditorFramework/Assets/AssetCurator.h
@@ -75,6 +75,7 @@ struct EZ_EDITORFRAMEWORK_DLL ezAssetInfo
     TransformError,
     MissingTransformDependency,
     MissingThumbnailDependency,
+    MissingPackageDependency,
     CircularDependency,
     COUNT,
   };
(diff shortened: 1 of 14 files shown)


Sanakan8472:
What is missing here is inverse dependency tracking. If you have a dependency that is a pure packaging dependency and you delete it while the editor is open, the dependent assets won't be flagged as MissingPackageDependency because there is no inverse tracking. See AssetCurator.h: m_InverseTransformDeps / m_UnresolvedTransformDeps for example.
These are also used to find all uses and other search functionalities that require inverse lookups.

PR #1383, opened 2024-08-29
Assert fix, icon polishing

jankrassnigg:


Code/Editor/EditorFramework/Document/Implementation/GameObjectDocument.cpp
```diff
@@ -308,10 +308,11 @@ void ezGameObjectDocument::DetermineNodeName(const ezDocumentObject* pObject, co
 
     for (auto pProperty : properties)
     {
+      const auto type = pProperty->GetSpecificType();
+
       // search for string properties that also have an asset browser property -> they reference an asset, so this is most likely the most
       // relevant property
-      if (
```
jankrassnigg, line 313: Bugfix for the resource names not showing up anymore in the scenegraph window as node names

Code/EditorPlugins/Scene/EditorPluginScene/Actions/LayerActions.cpp
```diff
@@ -153,7 +153,11 @@ void ezLayerAction::ToggleLayerLoaded(ezScene2Document* pSceneDocument, ezUuid l
   }
 
   pSceneDocument->SetLayerLoaded(layerGuid, bLoad).LogFailure();
-  pSceneDocument->SetActiveLayer(layerGuid).LogFailure();
+
+  if (bLoad)
+  {
+    pSceneDocument->SetActiveLayer(layerGuid).LogFailure();
```
jankrassnigg, line 159: Fixes unnecessary failure log output when deactivating a layer

Code/EditorPlugins/Scene/EditorPluginScene/Scene/SceneDocument.cpp
```diff
@@ -188,6 +188,26 @@ void ezSceneDocument::GroupSelection()
   ShowDocumentStatus(ezFmt("Grouped {} objects", numSel));
 }
 
+void ezSceneDocument::SelectParentObject()
+{
+  const auto& Sel = GetSelectionManager()->GetSelection();
```
jankrassnigg, line 193: New feature: option to quickly select the parent of an object

Code/Tools/Libs/ToolsFoundation/Command/Implementation/TreeCommands.cpp
```diff
@@ -702,7 +702,7 @@ ezStatus ezSetObjectPropertyCommand::DoInternal(bool bRedo)
 
   if (!bRedo)
   {
-    EZ_ASSERT_DEBUG(m_NewValue.GetType() != ezVariantType::StringView && m_NewValue.GetType() != ezVariantType::TypedPointer, "Variants that are stored in the command history must hold ownership of their value.");
```
jankrassnigg, line 705: This assert would often fire (now) when editing arrays that contain resource references (string views).
Looks like this assert isn't needed? Nothing breaks without it.

Sanakan8472: So this assert makes sure that the undo history does not have pointers to external memory which could be freed and cause crashes later. In this specific case it only works because the stringview points to the RTTI default value property so it is static. Unless it's a phantom type in which case updating type info would invalidate the default property and thus invalidate this memory address. So the code should not be changed.
Instead, go up the callstack on the assert and look for a good location to fix the type of the change.

One option would be to update `ezDocumentObjectConverterReader::ApplyDiff` to call:
```C++
      if (pProp->GetFlags().IsAnySet(ezPropertyFlags::IsEnum | ezPropertyFlags::Bitflags) || bIsValueType)
      {
        const ezVariantType::Enum memberType = ezToolsReflectionUtils::GetStorageType(pProp);
        if (bIsValueType && op.m_Value.GetType() != memberType)
          op.m_Value = op.m_Value.ConvertTo(memberType);

        pObjectAccessor->SetValue(pObject, pProp, op.m_Value).IgnoreResult();
      }
```
`ezToolsReflectionUtils::GetStorageType` returns a value that can be stored in the object as well as in the undo history. There might be a better place to update the type, for example when the default value is computed, but I'd have to have a deeper look at the code. One example would be to replace all calls in tools code to `ezReflectionUtils::GetDefaultValue` with `ezToolsReflectionUtils::GetStorageDefault` which does the same but takes care of the storage type. However, the later one does not support given a sub-index to the function so some more work is required for that kind of fix.

jankrassnigg: fixed

PR #1390, opened 2024-09-10
Light settings tweaks and other minor code changes

jankrassnigg:


Code/Engine/RendererFoundation/Descriptors/Descriptors.h
```diff
@@ -139,14 +139,12 @@ struct EZ_RENDERERFOUNDATION_DLL ezGALVertexAttribute
...
 
-  ezGALVertexAttribute(ezGALVertexAttributeSemantic::Enum semantic, ezGALResourceFormat::Enum format, ezUInt16 uiOffset, ezUInt8 uiVertexBufferSlot,
-    bool bInstanceData);
+  ezGALVertexAttribute(ezGALVertexAttributeSemantic::Enum semantic, ezGALResourceFormat::Enum format, ezUInt16 uiOffset, ezUInt8 uiVertexBufferSlot);
 
   ezGALVertexAttributeSemantic::Enum m_eSemantic = ezGALVertexAttributeSemantic::Position;
   ezGALResourceFormat::Enum m_eFormat = ezGALResourceFormat::XYZFloat;
   ezUInt16 m_uiOffset = 0;
   ezUInt8 m_uiVertexBufferSlot = 0;
-  bool m_bInstanceData = false;
```
C-Core, line 149: Why has this been removed? We don't use it at the moment but will in the near future.

Code/Engine/RendererFoundation/RendererFoundationDLL.h
```diff
@@ -281,7 +281,8 @@ struct ezGALCompareFunc
 
   enum Enum
   {
-    Never = 0,
+    None = 0,
+    Never,
```
C-Core: What is the difference between none and never?

Code/Engine/RendererFoundation/Shader/Types.h
```diff
@@ -30,6 +30,31 @@ class ezShaderMat3
   float m_Data[12];
 };
 
+/// \brief A wrapper class that converts a ezMat4 into the correct data layout for shaders.
```
C-Core, line 34: We have used ezMat4 directly for constant buffers etc. so far, why is this needed?

Code/Engine/RendererFoundation/Descriptors/Descriptors.h
jankrassnigg: It's not supported on all platforms, and since we don't use it at the moment, I find it unnecessary to expose it.
I did keep the relevant D3D and Vulkan implementation code in (commented out), in case we do want to re-add it, but as long as we don't use it, I would remove it from the interface.

Code/Engine/RendererFoundation/RendererFoundationDLL.h
jankrassnigg: We currently don't have an "off" state for this. D3D doesn't care, but other APIs do. So this is for the situation where the comparison isn't used, at all. It could also be stored separately, but that feels unnecessary.

Code/Engine/RendererFoundation/Shader/Types.h
jankrassnigg: Because of bugs in other APIs: https://issues.chromium.org/issues/364267168
Basically it allows me to inject a matrix transpose at some point.

Code/Engine/RendererFoundation/Descriptors/Descriptors.h
C-Core: I would be very surprised if there is a platform that doesn't support it. It has been around since Dx9 and very old opengl version. And yes webgpu does support it.

Code/Engine/RendererFoundation/RendererFoundationDLL.h
C-Core: Always is your "off" state

jankrassnigg: WebGPU differentiates between "Always" and "Undefined". In some situations "Always" does make sense (I think), so you have to set that, but if you set something other than "Undefined", it will complain about that, if you use it in situations where you can't compare things.

Code/Engine/RendererFoundation/Descriptors/Descriptors.h
jankrassnigg: We store it on a per-vertex stream basis (see code above). In WebGPU this is a single value on an entire VertexBufferLayout, which combines all streams.
So yes, it supports it, but no, not the way that we have it in the API.

Code/Engine/RendererFoundation/RendererFoundationDLL.h
jankrassnigg: I've reverted this change for now, because it breaks the renderer tests (though not on my machine).

PR #1391, opened 2024-09-15
Add support for cross-compiling for the SteamRT Sniper

Ingrater:
This adds support for cross-compiling the ezEngine runtime for the SteamRT Sniper. This includes a CMakeUserPresets.json setup to do so. This allows ezEngine to run on the SteamRT, which means it can be shipped as a Linux game on Steam and requires only one build to be done and will then run on all Linux distributions Steam supports. With this there is no need to do a separate ezEngine build for each Linux distribution.


Sanakan8472 (approved):

Code/ThirdParty/assimp
jankrassnigg, line 1: Can you push the submodule changes (also for FMOD below) to their main branch ?

jankrassnigg: Didn't see that you have PRs for those already. But you probably need to update these submodule references after they are completed.

PR #1392, opened 2024-09-15
Fix clang linker error and add Ubuntu 24 support

Sanakan8472:
For some reason, clang requires dll export of static thread_local members while all other copilers don't.

### DIFF
--- a/Code/ThirdParty/Jolt/Physics/Body/BodyAccess.h
+++ b/Code/ThirdParty/Jolt/Physics/Body/BodyAccess.h
@@ -8,17 +8,6 @@
 
 JPH_NAMESPACE_BEGIN
 
-/// EZ Modification start
-/// This is a workaround for clang requiring dll export of static thread_local members.
-#ifdef __clang__
-  #undef EZ_CLANG_JPH_EXPORT
-  #define EZ_CLANG_JPH_EXPORT JPH_EXPORT
-#else
-  #undef EZ_CLANG_JPH_EXPORT
-  #define EZ_CLANG_JPH_EXPORT
-#endif
-/// EZ Modification end
-
 class EZ_CLANG_JPH_EXPORT BodyAccess
 {
 public:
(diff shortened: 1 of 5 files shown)


Code/ThirdParty/Jolt/Physics/Body/BodyAccess.h
```diff
@@ -8,7 +8,18 @@
 
 JPH_NAMESPACE_BEGIN
 
-class BodyAccess
+/// EZ Modification start
+/// This is a workaround for clang requiring dll export of static thread_local members.
+#ifdef __clang__
+  #undef EZ_CLANG_JPH_EXPORT
```
Ingrater: Let's put this next to the definition of `JPH_EXPORT` so we don't have to repeat it multiple times.

Ingrater (changes requested):

Code/ThirdParty/Jolt/Physics/Body/BodyAccess.h
Ingrater: Also I'm 90% sure this requirement is linux only. Your fix however also kicks in when compiling with clang on windows. My assumption would be, that this now breaks the debug build on windows when compiling with clang.

jankrassnigg: I'm not thrilled to have an ez specific change in this library, that I can't merge back. Could you maybe rename this and clean it up, so that I can submit this to jolt?

Sanakan8472: I can try. Need to look into how Jolt defines their stuff. THe problem seems to be that:
1. Windows: "data with thread storage duration may not have dll interface"
2. Linux clang: exact oposite of windows.
3. Linux gcc: does not care one way or the other.

jankrassnigg: You can probably keep this extra define, but rename it to something like JPH_EXPORT_CLANG. And change the comments to not mention EZ.

Sanakan8472: I've pushed new changes, pleaes have a look!

jankrassnigg (approved):
Looks good!

PR #1397, opened 2024-09-19
Various improvements

### LINKED ISSUE #1395: [BUG] ezFileSystem::GetFileReader impl seems wrong in case bAllowFileEvents=false is passed
**Bug description:**
To be clear: I haven't tested this and I'm also not blocked by this, but it struck my eyes when I stepped through the code.

Look into the loop inside ezFileSystem::GetFileReader that tests all data directories. It tries to get the reader:

ezDataDirectoryReader* pReader = ...
(line 498 in my checkout)

but the (valid) pReader is never returned in the bAllowFileEvents==false case. Instead it proceeds to the error case.

jankrassnigg:
* Better configuration options for SIMD level
* ezRandom interface tweak to prevent accidentally using it wrong
* Fixed #1395
* Some new math convenience functions
* SSE compile fixes


Code/Engine/Foundation/IO/FileSystem/Implementation/FileSystem.cpp
```diff
@@ -502,15 +502,18 @@ ezDataDirectoryReader* ezFileSystem::GetFileReader(ezStringView sFile, ezFileSha
     // Let the data directory try to open the file.
     ezDataDirectoryReader* pReader = s_pData->m_DataDirectories[i].m_pDataDirType->OpenFileToRead(sRelPath, FileShareMode, bOneSpecificDataDir);
 
-    if (bAllowFileEvents && pReader != nullptr)
+    if (pReader != nullptr)
```
C-Core, line 505: Same issue with the GetFileWriter method below

jankrassnigg: Thanks! Fixed.

PR #1405, opened 2024-10-09
Some GAL modifications

jankrassnigg:
These changes are WIP, but pushed here to let others already review and give feedback.

Generally it introduces texture 2d and cube array types, because relying on array size doesn't always work.


Code/Engine/RendererCore/Pipeline/Implementation/Passes/BloomPass.cpp
```diff
@@ -63,7 +63,7 @@ bool ezBloomPass::GetRenderTargetDescriptions(const ezView& view, const ezArrayP
     ezGALTextureCreationDescription desc = *inputs[m_PinInput.m_uiInputIndex];
     desc.m_uiWidth = desc.m_uiWidth / 2;
     desc.m_uiHeight = desc.m_uiHeight / 2;
-    desc.m_Format = ezGALResourceFormat::RG11B10Float;
+    desc.m_Format = ezGALResourceFormat::RGBAHalf; // rg10b11 is not renderable on all platforms
```
jankrassnigg: Not sure about this one, can also be a dynamic check.

Code/Engine/RendererCore/Pipeline/Implementation/Passes/SourcePass.cpp
```diff
@@ -52,6 +52,7 @@ ezGALTextureCreationDescription ezSourcePass::GetOutputDescription(const ezView&
   const ezGALRenderTargets& renderTargets = view.GetActiveRenderTargets();
 
   ezGALTextureCreationDescription desc;
+  desc.m_Type = ezGALTextureType::Texture2DArray;
```
jankrassnigg, line 55: Not sure whether all input textures should simply be array textures.

Code/Engine/RendererVulkan/CommandEncoder/Implementation/CommandEncoderImplVulkan.cpp
```diff
@@ -1,5 +1,6 @@
 #include <RendererVulkan/RendererVulkanPCH.h>
 
+#include <RendererFoundation/Resources/RendererFallbackResources.h>
```
jankrassnigg, line 3: Haven't tested the Vulkan changes yet, they probably don't compile.

Data/Base/Shaders/Common/Lighting.h
```diff
@@ -74,7 +74,7 @@ float3 SampleSceneColor(float2 screenPosition)
 
 float SampleSceneDepth(float2 screenPosition)
 {
-  float depthFromZBuffer = SceneDepth.SampleLevel(PointClampSampler, float3(screenPosition.xy * ViewportSize.zw, s_ActiveCameraEyeIndex), 0.0f).r;
```
jankrassnigg, line 77: Probably should be reverted, but this may need a platform specific solution.

Data/Base/Shaders/Common/ObjectConstants.h
```diff
@@ -23,7 +23,7 @@ StructuredBuffer<ezPerInstanceData> perInstanceData;
 StructuredBuffer<Transform> skinningTransforms;
 #  endif
 
-Buffer<float4> perInstanceVertexColors;
```
jankrassnigg, line 26: Not sure what to do about this one, doesn't work on all platforms.

Data/Base/Shaders/Pipeline/DownscaleDepth.ezShader
```diff
@@ -51,7 +50,7 @@ float4 main(PS_IN Input) : SV_Target
   for (int i = 0; i < 4; ++i)
   {
     float2 samplePos = (screenPosition.xy + offsets[i]) * PixelSize;
-    depth = max(depth, DepthTexture.SampleLevel(DepthSampler, float3(samplePos, s_ActiveCameraEyeIndex), 0).r);
```
jankrassnigg, line 54: Also needs a platform specific solution.

Data/Base/Shaders/Pipeline/ReflectionFilteredSpecular.ezShader
```diff
@@ -22,7 +22,7 @@ CullMode = CullMode_None
 #include "ReflectionFilteredSpecularConstants.h"
 
 TextureCube InputCubemap;
-RWTexture2DArray<float4> ReflectionOutput;
+RWTexture2DArray<half4> ReflectionOutput;
```
jankrassnigg, line 25: WebGPU doesn't like that this wasn't the identical type that is bound.

Data/Base/Shaders/Pipeline/SSAO.ezShader
```diff
@@ -38,16 +37,16 @@ Texture2D NoiseTexture;
...
 
 float SampleLowResDepth(float2 texCoords, float mipLevel)
 {
 #if 0
-  return LinearizeZBufferDepth(DepthTexture.SampleLevel(DepthSampler, float3(texCoords, s_ActiveCameraEyeIndex), 0).r);
+  return LinearizeZBufferDepth(DepthTexture.Load(int4(texCoords.x, texCoords.y, s_ActiveCameraEyeIndex, 0)).r);
 #else
   texCoords = texCoords * TexCoordsScale;
-  return LowResDepthTexture.SampleLevel(DepthSampler, float3(texCoords, s_ActiveCameraEyeIndex), mipLevel).r;
+  return LowResDepthTexture.Load(int4(texCoords.x, texCoords.y, s_ActiveCameraEyeIndex, mipLevel)).r;
```
jankrassnigg: Not sure what to do about this, WebGPU doesn't like sampling from depth textures (though may need to be investigated further).

Code/Engine/RendererCore/Pipeline/Implementation/Passes/BloomPass.cpp
C-Core: This has to be a dynamic check, RGBAHalf is double the size and thus double the memory bandwidth of RG11B10Float so we should use the smaller format on platforms that support it.

Code/Engine/RendererCore/Pipeline/Implementation/Passes/SourcePass.cpp
C-Core: Why is this needed?

Data/Base/Shaders/Common/Lighting.h
C-Core: Since we are using a point filter we can switch to Load instead of Sample, but there are a few things that make Load tricky:
- Load takes absolute pixel coordinates instead of relative coordinates like Sample, so the coordinates in your changes are wrong.
- We need to manually take care of clamping in the shader code.

Data/Base/Shaders/Common/ObjectConstants.h
C-Core: This is not the same, the underlying buffer is one uint per element and we utilize the hardware to do the conversion to float4 for us. StructuredBuffer<float4> indicates that the underlying buffer already is a float4 per element, which is not the case.

Data/Base/Shaders/Pipeline/SSAO.ezShader
C-Core: Again the coordinates are in the wrong space. Furthermore we use a border clamp sampler here so apart from the coordinate clamping we would also need to emulate the border behavior in shader which will be a waste of performance on platforms that support depth sampling. So I guess we would need a SAMPLE_DEPTH macro or function to abstract that.

Code/Engine/RendererCore/Pipeline/Implementation/Passes/SourcePass.cpp
jankrassnigg: Uh, I'm not sure I remember the exact details. @Sanakan8472 maybe?
I think there are cases where we create a texture-array resource view to input textures somewhere later. The source textures are then just arrays of size 1. This works in D3D11, but WebGPU absolutely doesn't like this. You can create a non-array view to an array texture, but not the other way round. So if you don't know what other code will do with the texture later, this works better.

Data/Base/Shaders/Common/Lighting.h
jankrassnigg: The absolute coordinates explain some artifacts I've been seeing, good to know.

Data/Base/Shaders/Common/ObjectConstants.h
jankrassnigg: Ah ok, didn't know such a thing existed. Afaik WebGPU doesn't have this feature. So the correct change would be to use a StructuredBuffer<uint32> and then unpack and rescale to float4 in the shader? Do you think that would make a performance difference? I'd assume D3D11 just injects such code into the shader for you, or is that indeed dedicated hardware ?

C-Core: A StructuredBuffer<uint> is not a good idea, the struct in a structured buffer shouldn't be that small as a single uint, 16bytes at least for good perf, ideally also 16bytes aligned. Would Buffer<uint> work on WebGPU or does it not support raw buffers at all? 
No D3D11 does not inject shader code, maybe the driver does, but in most cases it will just utilize the same HW that is used for texture fetches which already have conversions like these build in. In theory we could also use a texture but then you have stricter size limitations, so Buffer<uint> with manual conversion in the shader would be the best solution if it is supported.

jankrassnigg: IIRC the Spir-V to WGSL compiler couldn't handle this. So could be a not implemented feature, or something that has no equivalent there. WebGPU has storage buffers, which are, to my knowledge just raw buffers. I think they should be used, but not sure what change in HLSL would be necessary.

I'll probably revert some of these changes for now, and come back to them later, but now I at least already know better what to do about them.

C-Core: Storage buffers are structured buffers, here is a discussion regarding storage **texel** buffers (that's the Buffer equivalent in vulkan or opengl) in WebGPU: https://github.com/gpuweb/gpuweb/issues/162
TLDR they are not supported yet.

Code/Engine/RendererCore/Pipeline/Implementation/Passes/SourcePass.cpp
C-Core: Ok fine by me, we can create all pipeline textures as array textures if the shader wants to have an array texture view for stereo support. For D3D11 it shouldn't matter at all and if it does we can always force it to be a texture2D in the D3D11 implementation if it only has one slice.

Data/Base/Shaders/Pipeline/SSAO.ezShader
jankrassnigg: Removed from this PR.

Data/Base/Shaders/Pipeline/DownscaleDepth.ezShader
jankrassnigg: Removed from this PR.

Data/Base/Shaders/Common/ObjectConstants.h
jankrassnigg: Removed from this PR.

Data/Base/Shaders/Common/Lighting.h
jankrassnigg: Added clamping and should now sample at the correct coordinate.

Code/Engine/RendererVulkan/CommandEncoder/Implementation/CommandEncoderImplVulkan.cpp
jankrassnigg: Tested Vulkan and made several adjustments. It now runs and shows no additional validation errors, however the ambient lighting is much darker, so maybe something with the probes doesn't work quite right, but not sure whether that was already the case before this change.

Code/Engine/RendererCore/Pipeline/Implementation/Passes/BloomPass.cpp
jankrassnigg: Removed from this PR.

Data/Base/Shaders/Common/Lighting.h
```diff
@@ -74,7 +74,10 @@ float3 SampleSceneColor(float2 screenPosition)
...
+  int screenX = clamp(screenPosition.x, 0, ViewportSize.x - 1);
+  int screenY = clamp(screenPosition.y, 0, ViewportSize.x - 1);
```
C-Core: ViewportSize.y, or you just write both lines like:
int2 screenXY = clamp(screenPosition.xy, 0, ViewportSize.xy - 1);
the int4 in load can also be constructed like this
int4(screenXY, s_ActiveCameraEyeIndex, 0)

jankrassnigg: fixed

Code/Engine/RendererCore/Pipeline/Implementation/Passes/SourcePass.cpp
Sanakan8472: Yep the problem is that some APIs are stricter when binding missmatching views to descriptors so we should be able to explicitly say whether we mean a 2d texture or a 2d texture array regardless of slice count. The same needs to be done for views as well at which point we could also remove the proxy texture support as it's a clutch to emable a sub-set of having explicit type enums.

Data/Base/Shaders/Pipeline/ReflectionFilteredSpecular.ezShader
Sanakan8472: The same actually happens on some Vulkan devices. In theory, your descriptor types and render target output types should match exactly to what you bind to them.

Code/Engine/RendererCore/Pipeline/Implementation/Passes/BloomPass.cpp
Sanakan8472: Re-added with format support checks. Will fall back to RGBAHalf if RG11B10Float is not supported as render target and sampling.

Code/Engine/RendererVulkan/CommandEncoder/Implementation/CommandEncoderImplVulkan.cpp
Sanakan8472: Issue has been fixed. Implementation of render target view and resource view will be much easier once we add an explicit type to the views and forbit mixing and matching of incompatible views with descriptors.

Data/Base/Shaders/Pipeline/DownscaleDepth.ezShader
Sanakan8472: I've added a platform function `ezSampleLevel_PointClampBorder` which abstracts this as well as a unit test.

Data/Base/Shaders/Pipeline/SSAO.ezShader
Sanakan8472: Replaced as well with `ezSampleLevel_PointClampBorder`.

Code/Engine/RendererCore/Pipeline/Implementation/Passes/BloomPass.cpp
```diff
@@ -63,7 +63,9 @@ bool ezBloomPass::GetRenderTargetDescriptions(const ezView& view, const ezArrayP
     ezGALTextureCreationDescription desc = *inputs[m_PinInput.m_uiInputIndex];
     desc.m_uiWidth = desc.m_uiWidth / 2;
     desc.m_uiHeight = desc.m_uiHeight / 2;
-    desc.m_Format = ezGALResourceFormat::RG11B10Float;
+    const ezGALDeviceCapabilities& caps = ezGALDevice::GetDefaultDevice()->GetCapabilities();
+    const bool bSupportsRG11B10Float = caps.m_FormatSupport[ezGALResourceFormat::RG11B10Float].AreAllSet(ezGALResourceFormatSupport::RenderTarget | ezGALResourceFormatSupport::Texture);
+    desc.m_Format = bSupportsRG11B10Float ? ezGALResourceFormat::RG11B10Float : ezGALResourceFormat::RGBAHalf;
```
jankrassnigg: the used format also needs to be adjusted in line 114 and 119 below (can't comment there)

Data/Base/Shaders/Common/ObjectConstants.h
Sanakan8472: Added `ezGALDeviceCapabilities::m_bTexelBuffer` and platform define `SUPPORTS_TEXEL_BUFFER`. These must match and are static per platform as the later one is defined in the shader platform headers. Implementation passes proc gen test if forced on.

Code/Engine/RendererCore/Pipeline/Implementation/Passes/BloomPass.cpp
Sanakan8472: Good point! Fixed all instances here and in SourcePass.

Code/Engine/RendererCore/Pipeline/Implementation/Passes/SourcePass.cpp
Sanakan8472: On Vulkan we currently need to create two native views per GAL resource because of this ambiguity. This needs to be made explicit at some point.

Data/Base/Shaders/Common/Platform_Web.h
```diff
@@ -83,4 +83,20 @@ float4 ezEvaluateAttributeAtSample(float4 Attribute, uint SampleIndex, uint NumM
...
+  // Get the texture size at the specified mip level
+  uint width, height, elements, levels;
+  DepthTexture.GetDimensions(0, width, height, elements, levels);
+  // Convert normalized coordinates to texel space
+  float2 texelCoords = SamplePos * int2(width, height);
+  // Get the integer parts of the coordinates
+  int2 texelBase = int2(floor(texelCoords));
+  float4 texel = DepthTexture.Load(int4(texelBase, ArrayIndex, MipLevel));
+  // validUV is one on each axis if within [0, 1] range.
+  float2 validUV = step(float2(0, 0), SamplePos) - step(float2(1, 1), SamplePos);
```
C-Core: bool validUV = all(saturate(SamplePos) == SamplePos);
return validUV ? texel : BorderColor;

C-Core (approved):

PR #1408, opened 2024-10-16
Fixed cross platform serialization issues with visual scripts

### LINKED ISSUE #1404: [BUG] Visual Scripts don't work on 32 bit systems
Data can't be loaded, due to alignment differences.

C-Core:
Visual scripts had a couple of things pre-compiled that depend on the target platform like byteoffsets for local variables etc. These offsets could differ e.g. from a 64bit to 32bit platform.
Now only platform independent data is serialized and and things like byteoffsets for local variables are calculated at load time.


Code/EnginePlugins/VisualScriptPlugin/Runtime/VisualScript_inl.h
```diff
@@ -87,6 +87,24 @@ EZ_ALWAYS_INLINE ezVisualScriptGraphDescription::DataOffset ezVisualScriptGraphD
...
+{
+  return m_NumOutputDataOffsets <= EZ_ARRAY_SIZE(m_OutputDataOffsets.m_Embedded) ? m_OutputDataOffsets.m_Embedded : m_OutputDataOffsets.m_Ptr;
+}
+
+// static
+template <typename T>
+EZ_ALWAYS_INLINE constexpr ezUInt32 ezVisualScriptGraphDescription::Node::GetUserDataAlignment()
+{
+  // Ensures at least 8 byte alignment, thus making it compatible between 32 and 64 bit platforms.
+  return ezMath::Max<ezUInt32>(EZ_ALIGNMENT_OF(T), 8u);
```
jankrassnigg: I never understood why EZ_ALIGNMENT_OF works the way it does. Is this really the correct choice here? Because IIRC it also uses some MAX value internally.

jankrassnigg (approved):

Code/EnginePlugins/VisualScriptPlugin/Runtime/VisualScript_inl.h
C-Core: Looks like we don't need the EZ_ALIGNMENT_OF macro anymore, at least it in msvc as far as I tested it. Let's see what CI and the other platforms say.

PR #1414, opened 2024-10-22
Refactor of platform specific includes

jankrassnigg:
If a library has a folder "Platform/XYZ" (with XYZ being the target platform name), then this is automatically added as a public include directory. This allows to do #include<Feature.h> where "Feature.h" will redirect to a platform specific header.

This means, that we don't need to have the #ifdef PLATFORM_XYZ cascades in many places anymore, and adding a platform becomes much easier, because you don't need to modify existing code.

If code for a platform is missing, the #include fails at compile time, and no "#else error" is needed.

* Moved many platform specific files around to clean them up and automatically include them.
* Removed now unnecessary #ifdef sections.
* FileServe now automatically loads all available shader compiler plugins

There's still quite a lot more to do, to separate platform code from the main code, but that may come later.

This is part of #1119.


Code/Tools/HeaderCheck/HeaderCheck.cpp
```diff
@@ -406,7 +406,8 @@ class ezHeaderCheckApp : public ezApplication
       ezStringBuilder currentFileLower = sCurrentFile.GetFileNameAndExtension();
       currentFileLower.ToLower();
 
-      bool ignore = m_IgnoreTarget.m_byName.Contains(includeFileLower) || m_IgnoreSource.m_byName.Contains(currentFileLower);
+      // ignore includes ending in "_Platform.h", they redirect to platform specific EZ headers
+      const bool ignore = m_IgnoreTarget.m_byName.Contains(includeFileLower) || m_IgnoreSource.m_byName.Contains(currentFileLower) || includeFileLower.EndsWith_NoCase("_Platform.h");
```
jankrassnigg, line 410: Note this addition, to automatically ignore includes of the form

```cpp
#include <Stuff_Platform.h>
```

Otherwise every platform specific include is assumed to be a 'system include' and reported as an error.

Code/Tools/Fileserve/FileserveApp.cpp
```diff
@@ -26,8 +26,23 @@ void ezFileserverApp::AfterCoreSystemsStartup()
...
+    {
+      ezStringBuilder sName = it.GetStats().m_sName;
+
+      if (sName.HasExtension("DLL"))
+      {
+        sName.RemoveFileExtension();
+        ezPlugin::LoadPlugin(sName, ezPluginLoadFlags::PluginIsOptional).IgnoreResult();
+      }
+    }
+  }
```
jankrassnigg, line 45: Automatic shader compiler plugin loading.

Code/BuildSystem/CMake/CMakeUtils/ezUtilsTarget.cmake
```diff
@@ -127,6 +127,11 @@ macro(ez_create_target TYPE TARGET_NAME)
 		endif()
 	endif()
 
+	# add a platform specific include directory
+	if (EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/Platform/${EZ_CMAKE_PLATFORM_POSTFIX}")
+		target_include_directories(${TARGET_NAME} PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}/Platform/${EZ_CMAKE_PLATFORM_POSTFIX}")
+	endif()
```
jankrassnigg, line 133: Automatic addition of platform specific include path.

PR #1415, opened 2024-10-23
Further refactor of platform specific code

jankrassnigg:
* Moved more platform specific code into separate files and the right folders.
* Removed more #ifdef's for the different platforms

Part of #1119


Code/Engine/Foundation/Platform/Linux/FormatString_Linux.cpp
```diff
@@ -0,0 +1,17 @@
+#include <Foundation/FoundationPCH.h>
+
+#if EZ_ENABLED(EZ_PLATFORM_ANDROID)
```
Sanakan8472: This should be EZ_PLATFORM_LINUX

PR #1417, opened 2024-10-27
Readback refactor

Sanakan8472:
## Readback Changes

* The readback logic has been extracted from the `ezGALTexture` code and placed into its own resource: `ezGALReadbackTexture`. This has the benefit of not wasting memory on every potential texture you may want to readback at some point. An equivalent `ezGALReadbackBuffer` was created as well.
* Removed `ezGALWindowSwapChainCreationDescription::m_bAllowScreenshots` and `ezGALResourceAccess::m_bReadBack` as the texture resource itself no longer hold the readback staging resources. It is safe to potentially readback any resource, setting e.g. `vk::BufferUsageFlagBits::eTransferSrc` has no impact on performance.
* Readback of any kind is no longer allowed inside a rendering scope.
* A texture readback looks like this now:
```C++
ezGALTextureHandle hTexture = ...;
const ezGALTexture* pTexture = m_pDevice->GetTexture(hTexture);
// Create readback texture using the description of the texture you intent to read back:
ezGALReadbackTextureHandle hReadbackTexture = m_pDevice->CreateReadbackTexture(pTexture->GetDescription());
// Start the readback async operation
encoder.ReadbackTexture(m_hReadbackTexture, hTexture);
ezGALFenceHandle hFence = encoder.InsertFence();
...
// Once the fence has been reached, it is safe to lock the memory of the readback texture:
ezEnum<ezGALAsyncResult> res = pDevice->GetFenceResult(hFence);
if (res == ezGALAsyncResult::Ready)
{
    ezGALTextureSubresource sourceSubResource;
    ezArrayPtr<ezGALTextureSubresource> sourceSubResources(&sourceSubResource, 1);
    ezHybridArray<ezGALSystemMemoryDescription, 1> memory;
    ezReadbackTextureLock lock = m_Readback.LockTexture(sourceSubResources, memory);
    EZ_ASSERT_ALWAYS(lock, "Failed to lock readback texture");
    // Do something with the locked memory, e.g. create an ezImage:
    ezImage image;
    ezTextureUtils::CreateSubResourceImage(pTexture->GetDescription(), sourceSubResource, memory[0], image, true);
}
```
* A few helper functions have been added to make this easier:
  * `ezGALReadbackTextureHelper` and `ezGALReadbackBufferHelper` automate creation / destruction of the readback texture / buffer resource. They also ensure that you are not locking memory that still has a readback pending. The class allows repeated readbacks of resources of different formats - the underlying readback resource will be re-created if it is not suitable for the newly requested readback operation.
  * `ezTextureUtils::CreateSubResourceImage`: Copies the given texture subresource from a locked memory `ezGALSystemMemoryDescription`.
  * `ezTextureUtils::CreateSubResourceView`: Tries to create a texture view by aliasing the given locked memory `ezGALSystemMemoryDescription`. Falls back to `CreateSubResourceImage` if not possible.
  * `ezTextureUtils::CopySubResource`: Copies a texture subresource memory to a new location with a different row pitch.

## Misc

* `ezGALSystemMemoryDescription::m_pData` is now an `ezBlobPtr` instead of a raw ptr. This allows for sanity checks to prevent running out of the buffer memory.
* `ezBlobPtr` can now be initialized from an `ezArrayPtr<T>`.
* Fixed bug that moved the cursor even if `ezMouseCursorClipMode::NoClip` was set.
* Fixed data race in `ezReflectionPool::SetConstantSkyIrradiance`.
* Better barriers on vulkan buffer readback / upload.
* Fixed a bug in the Vulkan fence implementation that caused the wrong result to be returned when querying the fence status.
* Fixed a crash in Vulkan in case the timestamp pools runs out in the middle of a rendering scope. Turns out, you can't reset pools inside a rendering scope so for now we just ensure 3 free pools before entering any rendering scope.
* Worked around a Visual Studio compiler bug in ozz that caused an access violation reading 0xFFFFFFFFFFFF.
* Fixed `Point.ezShader` on Vulkan. Turns out, DX9 supported PSIZE vertex interpolator attribute, DX11 removed it and Vulkan requires it. So had to add an ifdef to the shader.

### DIFF
--- a/Code/Engine/RendererCore/Textures/TextureUtils.h
+++ b/Code/Engine/RendererCore/Textures/TextureUtils.h
@@ -14,6 +14,13 @@ struct EZ_RENDERERCORE_DLL ezTextureUtils
 
   static void ConfigureSampler(ezTextureFilterSetting::Enum filter, ezGALSamplerStateCreationDescription& out_sampler);
 
+  /// \brief Copies the given texture subresource from `memory` into `out_Image` according to the texture description.
+  static void CreateSubResourceImage(const ezGALTextureCreationDescription& desc, const ezGALTextureSubresource& subResource, ezGALSystemMemoryDescription& memory, ezImage& out_Image, bool bRemoveSRGB);
+  /// \brief Returns an image view of the texture subresource in `memory`. If the format allows for it, the memory will be aliased, removing the need to copy the data but the view becomes invalid once the memory does. If this is not possible, the function reverts to calling CreateSubResourceImage with `ref_Temp` used as the data storage. You can check if `ref_Temp` is valid to figure out which code path was taken.
+  static ezImageView CreateSubResourceView(const ezGALTextureCreationDescription& desc, const ezGALTextureSubresource& subResource, ezGALSystemMemoryDescription& memory, ezImage& ref_Temp, bool bRemoveSRGB);
+  /// \brief Copies a texture subresource memory to a new location with a different row pitch.
+  static void CopySubResource(const ezGALTextureCreationDescription& desc, const ezGALTextureSubresource& subResource, const ezGALSystemMemoryDescription& sourceMemory, ezArrayPtr<ezUInt8> targetData, ezUInt32 uiTargetRowPitch);
+
   /// \brief If enabled, textures are always loaded to full quality immediately. Mostly necessary for image comparison unit tests.
   static bool s_bForceFullQualityAlways;
 };
--- a/Code/Engine/RendererCore/Textures/TextureUtils.h
+++ b/Code/Engine/RendererCore/Textures/TextureUtils.h
@@ -15,11 +15,11 @@ struct EZ_RENDERERCORE_DLL ezTextureUtils
   static void ConfigureSampler(ezTextureFilterSetting::Enum filter, ezGALSamplerStateCreationDescription& out_sampler);
 
   /// \brief Copies the given texture subresource from `memory` into `out_Image` according to the texture description.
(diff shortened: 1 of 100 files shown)


jankrassnigg:
Somethings wrong, GitHub doesn't run any builds on this PR.

Code/Engine/RendererCore/Textures/TextureUtils.h
```diff
@@ -14,6 +14,13 @@ struct EZ_RENDERERCORE_DLL ezTextureUtils
 
   static void ConfigureSampler(ezTextureFilterSetting::Enum filter, ezGALSamplerStateCreationDescription& out_sampler);
 
+  /// \brief Copies the given texture subresource from `memory` into `out_Image` according to the texture description.
+  static void CreateSubResourceImage(const ezGALTextureCreationDescription& desc, const ezGALTextureSubresource& subResource, ezGALSystemMemoryDescription& memory, ezImage& out_Image, bool bRemoveSRGB);
```
C-Core: Maybe the function should be named something like CopySubResourceToImage? Create... sounds misleading.

C-Core (approved):

Code/Engine/RendererCore/Textures/TextureUtils.h
Sanakan8472: Fixed!

Sanakan8472:
So I created a new PR https://github.com/ezEngine/ezEngine/pull/1418 and that one also fails to run triggers except for clang format. So probably the azure pipelines are broken. Not sure why though, they can be started manually without issue.

Sanakan8472:
Well the pipelines do show up if I manually trigger them so for now I'll just do that. I can't seem to find any fault anywhere. Maybe azure is broken?

PR #1425, opened 2024-11-09
Buffer Upload Refactor

Sanakan8472:
## Renderer Foundation Changes

* Added a call to `AssertOutsideRenderingScope` in all function in `ezGALCommandEncoder` that are not allowed to be called inside a render scope. This includes all copy, update, resolve and readback related functions.
* Added a new flag `ezGALBufferUsageFlags::Transient`. When set, it declares that the content of the buffer's content will not persist accross frames. This allows for various optimizations:
  * On DX11 this means that we can use the `D3D11_MAP_WRITE_DISCARD` followed by `D3D11_MAP_WRITE_NO_OVERWRITE` pattern each frame on these buffers.
  * Transient constant buffers have the addional contraint that they always need to be updated in their entirety. On Vulkan we can replace these updates with a dynamic uniform ring buffer in a later PR.
  * See `ezGALBufferPool` below.
* `ezGALDevice::CreateConstantBuffer` will create transient buffers by default now.
* Buffers requested from the `ezGPUResourcePool` will be marked as transient and only given out once per frame.
* `ezGALUpdateMode` has been revised to these new modes:
  * `TransientConstantBuffer`: Can be executed at any time in a command encoder. Only allowed on transient constant buffers. Buffer must be completely overwritten. Data will not persist across frames.
  * `AheadOfTime`: Can be executed at any time in a command encoder. Copy is ensured to happen before the next command in the command encoder. The same memory location can't be updated twice in one frame. Note that no GPU access must have happened to the modified memory range in the current command encoder before this call or undefined behavior will occur.
  * `CopyToTempStorage`: Only allowed outside a render pass. Uploads to temp buffer and then calls a buffer to buffer transfer at the current time in the command buffer. Only use this if the same buffer must have different content in the same frame.
* `ezGALRendererFallbackResources` now correctly handles 2d vs array textures.
* Added `ezGALBufferPool` class which manages a growing ring buffer of one or more buffers. The behaviour of the class changes depending on whether `Initialize` is called with or without a `ezGALBufferUsageFlags::Transient` buffer descriptor:
  * **Transient**: You must call `GetNewBuffer` at the start of every frame and whenever your current buffer has run out of space. Use this for transient data that changes every frame.
  * **Persistent**: `GetNewBuffer` must be called once after which you can re-use it every frame and the content in it will persist. You can use this fact to allocate the buffer lazily only in case you actually need it. If you feel the need to call `GetNewBuffer` after the first lazy allocation, you should consider switching to a transient buffer or using a resizable buffer instead.
  ```C++
  // Header:
      ezGALBufferPool m_VertexBuffer;
  // Cpp Init:
      ezGALBufferCreationDescription desc;
      desc.m_BufferFlags = ezGALBufferUsageFlags::VertexBuffer | ezGALBufferUsageFlags::Transient;
      desc.m_uiStructSize = sizeof(ezImguiVertex);
      desc.m_uiTotalSize = s_uiVertexBufferSize * desc.m_uiStructSize;
      desc.m_ResourceAccess.m_bImmutable = false;
      m_VertexBuffer.Initialize(desc, "DearImguiRenderer - VertexBuffer");
  // Cpp Usage:
      ezGALBufferHandle hVertexBuffer = m_VertexBuffer.GetNewBuffer();
      pCommandEncoder->UpdateBuffer(hVertexBuffer, 0, ezMakeArrayPtr(pRenderData->m_Vertices.GetPtr(), pRenderData->m_Vertices.GetCount()).ToByteArray(), ezGALUpdateMode::AheadOfTime);
      pRenderContext->BindMeshBuffer(hVertexBuffer, hIndexBuffer, &m_VertexDeclarationInfo, ezGALPrimitiveTopology::Triangles, pRenderData->m_Indices.GetCount() / 3);
  ```

## High Level Renderer Changes

* All the `ezRenderer` implementations have had the buffers they maintained replaced with `ezGALBufferPool` instances except for sprite and lensflare renderers which use `ezGPUResourcePool` buffers. This should be discouraged and replaced with `ezGALBufferPool` instances as well as it create uncontrollable descriptor trash creations if you effectively get random buffers each time. Most of pools are transient with the exception of the dynamic mesh buffer resource.
* Changed `ezInstanceData` to use a `ezGALBufferPool` with a `bTransient` flag in the contructor. To keep the diff small, the `ezInstanceData::GetInstanceData` function was extended with a `ezRenderContext` parameters so the implementation can request and bind a new buffer.
* `ezConstantBufferStorageBase` creates transient buffers now. To make sure these are updated before use as they don't persist across frames, the function `BeforeBeginFrame` is called at the start of the frame by the `ezRenderContext`. This marks the buffer as dirty no matter the internal state. This can probably be made more efficient in later itterations.

## Vulkan Changes

* All buffers are now created in device local memory.
* `ezInitContextVulkan` was extended to support device local buffer updates.
* Fixed a crash in `ezPipelineBarrierVulkan`.


Code/Engine/RendererFoundation/Resources/Implementation/BufferPool.cpp
```diff
@@ -0,0 +1,97 @@
...
+  ezGALDevice* pDevice = ezGALDevice::GetDefaultDevice();
+  buffer = pDevice->CreateBuffer(m_Desc);
+  if (!m_sDebugName.IsEmpty())
+  {
+
+    ezStringBuilder sTmp;
+    sTmp.SetFormat("{}-{}#{}", m_sDebugName, m_iUniqueID, m_UsedBuffers.GetCount() + m_FreeBuffers.GetCount());
+    pDevice->GetBuffer(buffer)->SetDebugName(sTmp);
+  }
+  EZ_ASSERT_DEV(!buffer.IsInvalidated(), "Failed to creat pool buffer");
```
C-Core: typo

C-Core (approved):

Code/Engine/RendererFoundation/Resources/Implementation/BufferPool.cpp
Sanakan8472: Fixed!

PR #1430, opened 2024-11-17
Delay Editor picking readback by one frame

Sanakan8472:
* Added various profiling scopes to make editor / engine / gpu corrrelation easier.
* Picking is now done with one frame delay, preventing blocking on the readback operation.
* Removed duplicate profiling scopes.


jankrassnigg (approved):
Any difference in how picking feels?

Also please test stuff like drag & drop of meshes into the scene, that they don't pick against themselves or behave weird in different ways. Same for gizmos like the drag-to-position gizmo. Not sure whether the readback delay has any effect on this.
Same for marquee selection (hold space + drag mouse).

Sanakan8472:
> Any difference in how picking feels?

So the good thing is that as we always download the entire image, the picking result does not change unless the content of the image changes due to movement or camera rotation.  So for things like dragging into the scene for example you would get the same result as before with slightly less delay as the picking buffer doesn't change between frames.
So far I haven't noticed any change from before the change.

PR #1435, opened 2024-11-21
Added shadow fade out

C-Core:
This adds functionality to fade out shadows of point and spot lights when they reach a certain size in screen space. For that a separate range called "ShadowFadeOutRange" has been introduced that can be controlled independent of the actual light range and therefore give better control. By default this range is the same as the light range.

Also tweaked the slope bias to reduce artifacts regardless of spot angle and light range.


Data/Base/Shaders/Common/Lighting.h
```diff
@@ -285,6 +286,8 @@ float CalculateShadowTerm(float3 worldPosition, float3 vertexNormal, float3 ligh
     shadowPosition += shadowDataBuffer[matrixIndex + 2] * worldPosition.z;
 
     debugColor = debugColors[matrixIndex];
+
+    fadeOut = ((shadowDataOffsetAndFadeOut >> 20) & 0xFFF) / 4095.0;
```
jankrassnigg: why the & 0xFFF ? isn't that redundant after the >> 20 ?

Code/Engine/RendererCore/Lights/Implementation/SpotLightComponent.cpp
```diff
@@ -2,26 +2,24 @@
...
 #include <Core/WorldSerializer/WorldWriter.h>
-#include <RendererCore/Debug/DebugRenderer.h>
 #include <RendererCore/Lights/Implementation/ShadowPool.h>
 #include <RendererCore/Lights/SpotLightComponent.h>
 #include <RendererCore/Pipeline/View.h>
 
-#if EZ_ENABLED(EZ_COMPILE_FOR_DEVELOPMENT)
-ezCVarBool cvar_RenderingLightingVisScreenSpaceSize("Rendering.Lighting.VisScreenSpaceSize", false, ezCVarFlags::Default, "Enables debug visualization of light screen space size calculation");
-#endif
+constexpr ezAngle c_MaxSpotAngle = ezAngle::MakeFromDegree(160.0f);
```
jankrassnigg, line 9: was there a problem with 179 degree? other than getting worse quality at such large angles, I mean

Code/Engine/RendererCore/Lights/Implementation/ShadowPool.cpp
```diff
@@ -44,6 +44,9 @@ EZ_END_SUBSYSTEM_DECLARATION;
 #if EZ_ENABLED(EZ_COMPILE_FOR_DEVELOPMENT)
 ezCVarBool cvar_RenderingShadowsShowPoolStats("Rendering.Shadows.ShowPoolStats", false, ezCVarFlags::Default, "Display same stats of the shadow pool");
 ezCVarBool cvar_RenderingShadowsVisCascadeBounds("Rendering.Shadows.VisCascadeBounds", false, ezCVarFlags::Default, "Visualizes the bounding volumes of shadow cascades");
+ezCVarFloat cvar_RenderingShadowsScaleMappingExponent("Rendering.Shadows.ScaleMappingExponent", 1.5f, ezCVarFlags::Default, "Determines how fast the shadow map size is reduced with screen space size");
+#else
+float cvar_RenderingShadowsScaleMappingExponent = 1.5f;
```
jankrassnigg: Wouldn't this be useful to allow games to tweak this in in shipping builds?

jankrassnigg (approved):

Code/Engine/RendererCore/Lights/Implementation/SpotLightComponent.cpp
C-Core: The frustum generation code would assert and even with something like 170 the shadows are completely broken due to the lack of precision.

PR #1436, opened 2024-11-23
Vulkan Ring Buffers

Sanakan8472:
## ezRingBufferTracker
* Tracks memory in a ring buffer on a frame by frame basis. The intended usage pattern is that for frame N, multiple allocations are called. This is followed by a call to `SubmitFrame` for frame N (optional) which allows the user to handle to memory ranges that were modified in this frame. Eventually at a later point in time `Free` is called for frame N to allow reuse of the memory. Note that the assumption is that N is always increasing and that `Free` is called for each frame in order.
* Added unit test.

## ezUniformBufferPoolVulkan
* `ezGALBufferVulkan` created with `ezGALBufferUsageFlags::Transient` no longer allocate any memory. Instead, allocations are handled by the new class `ezUniformBufferPoolVulkan`.
* Uses `ezRingBufferTracker` to track pools of memory. Each pool is a dynamic uniform ringbuffer.

## ezStagingBufferPoolVulkan
* Allocates temporary staging buffers from a large pool. Allocations will automatically be freed at the end of the frame. New (larger) pools will be created if the existing ones run out of space. Pools will be deleted after a certain time of no usage.
* Uses `ezRingBufferTracker` to track pools of memory. Each pool is a Vulkan buffer with `ezVulkanAllocationCreateFlags::HostAccessSequentialWrite` and `ezVulkanAllocationCreateFlags::Mapped` flags set and allocations will point directly into the permanently mapped buffer. Only memcpy operations should be used on the allocated buffers.

### DIFF
--- a/Code/Engine/RendererVulkan/CommandEncoder/Implementation/CommandEncoderImplVulkan.cpp
+++ b/Code/Engine/RendererVulkan/CommandEncoder/Implementation/CommandEncoderImplVulkan.cpp
@@ -1207,6 +1207,10 @@ ezResult ezGALCommandEncoderImplVulkan::FlushDeferredStateChanges()
     m_TextureAndSampler.Clear();
     const ezUInt32 uiSets = m_PipelineDesc.m_pCurrentShader->GetSetCount();
     m_DescriptorSets.SetCount(uiSets);
+
+    ezDeque<vk::DescriptorBufferInfo> dynamicUniformBuffers;
+    ezHybridArray<ezUInt32, 6> dynamicUniformBufferOffsets;
+
     for (ezUInt32 uiSet = 0; uiSet < uiSets; ++uiSet)
     {
       m_DescriptorSets[uiSet] = ezDescriptorSetPoolVulkan::CreateDescriptorSet(m_LayoutDesc.m_layout[uiSet]);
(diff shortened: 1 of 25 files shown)


Code/Engine/RendererVulkan/CommandEncoder/Implementation/CommandEncoderImplVulkan.cpp
```diff
@@ -1195,6 +1207,10 @@ ezResult ezGALCommandEncoderImplVulkan::FlushDeferredStateChanges()
     m_TextureAndSampler.Clear();
     const ezUInt32 uiSets = m_PipelineDesc.m_pCurrentShader->GetSetCount();
     m_DescriptorSets.SetCount(uiSets);
+
+    ezDeque<vk::DescriptorBufferInfo> dynamicUniformBuffers;
```
C-Core: Per frame allocation(s), keep around as member like the others above.

C-Core (approved):

Code/Engine/RendererVulkan/CommandEncoder/Implementation/CommandEncoderImplVulkan.cpp
Sanakan8472: Fixed!

PR #1461, opened 2024-12-19
Crash fixes

Sanakan8472:
* `ozz::math::Float4x4` needs to be aligned to 16 bytes so it is not safe to cast `ezMat4` to it as the later one is not 16 bytes aligned.
* Replaced all casts from `ezStringView` to `const char*` which ended up being put back into an `ezStringView`. Some of these produced invalid reads with a guarding allocator.
* Wrong brackets in AssetBrowerWidget.cpp. The old code was always true as `==` binds stronger than `?`.

### DIFF
--- a/Code/Engine/GameEngine/Animation/Skeletal/Implementation/AnimatedMeshComponent.cpp
+++ b/Code/Engine/GameEngine/Animation/Skeletal/Implementation/AnimatedMeshComponent.cpp
@@ -106,18 +106,18 @@ void ezAnimatedMeshComponent::InitializeAnimationPose()
     const ozz::animation::Skeleton* pOzzSkeleton = &pSkeleton->GetDescriptor().m_Skeleton.GetOzzSkeleton();
     const ezUInt32 uiNumSkeletonJoints = pOzzSkeleton->num_joints();
 
-    ezArrayPtr<ezMat4> pPoseMatrices = EZ_NEW_ARRAY(ezFrameAllocator::GetCurrentAllocator(), ezMat4, uiNumSkeletonJoints);
-
+    ezArrayPtr<ozz::math::Float4x4> pPoseMatrices = EZ_NEW_ARRAY(ezFrameAllocator::GetCurrentAllocator(), ozz::math::Float4x4, uiNumSkeletonJoints);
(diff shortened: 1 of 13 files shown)


Code/Engine/GameEngine/Animation/Skeletal/Implementation/AnimatedMeshComponent.cpp
```diff
@@ -106,18 +106,18 @@ void ezAnimatedMeshComponent::InitializeAnimationPose()
     const ozz::animation::Skeleton* pOzzSkeleton = &pSkeleton->GetDescriptor().m_Skeleton.GetOzzSkeleton();
     const ezUInt32 uiNumSkeletonJoints = pOzzSkeleton->num_joints();
 
-    ezArrayPtr<ezMat4> pPoseMatrices = EZ_NEW_ARRAY(ezFrameAllocator::GetCurrentAllocator(), ezMat4, uiNumSkeletonJoints);
-
+    ezArrayPtr<ozz::math::Float4x4> pPoseMatrices = EZ_NEW_ARRAY(ezFrameAllocator::GetCurrentAllocator(), ozz::math::Float4x4, uiNumSkeletonJoints);
```
jankrassnigg, line 109: I'm not sure that this would make difference?

Does the frame allocator use the alignment of the type in any way? I thought it just always alignes on 16 byte, so the casting back and forth "should" work either way.

jankrassnigg (approved):

Code/Engine/GameEngine/Animation/Skeletal/Implementation/AnimatedMeshComponent.cpp
Sanakan8472: Unless you are using a special aligned allocator, the normal allocator behaviour is to allocate memory that is aligned to the target type's alignment. And `ezMat4` is 4 byte aligned while `ozz::math::Float4x4` is 16  byte aligned. Thus, it is not safe to cast `ezMat4` to `ozz::math::Float4x4` as an unaligned SIMD register will cause nonsensical runtime errors (mostly access to 0x0 or 0xFFFFFFFFFFFFFFFF).

Sanakan8472: Looking at the code, the default allocator is aligned to 8 as that is what malloc produces.

PR #1470, opened 2025-01-05
Linux build fixes

Sanakan8472:
* Fixes for `RunCmake.sh` on a clean Ubuntu 24.04 install.
* Build fix: Newer gcc versions do not like our `EZ_ADD_DEFAULT_OPERATOR_NOTEQUAL` macro when used on a forward declared type.

### DIFF
--- a/RunCMake.sh
+++ b/RunCMake.sh
@@ -92,8 +92,14 @@ verlt() {
 }
 
 if [ "$Distribution" = "Ubuntu" -a "$Version" = "22" ] || [ "$Distribution" = "Ubuntu" -a "$Version" = "24" ] || [ "$Distribution" = "Mint" -a "$Version" = "21" ]; then
-  packages=(cmake build-essential ninja-build libxrandr-dev libxinerama-dev libomp-dev libxcursor-dev libxi-dev uuid-dev mold libfreetype-dev libtinfo5 libomp-dev)
+  packages=(cmake build-essential ninja-build libxrandr-dev libxinerama-dev libxcursor-dev libxi-dev uuid-dev mold libfreetype-dev libxkbcommon-dev)
 
+  if [ "$Distribution" = "Ubuntu" -a "$Version" = "24" ]; then
+    packages+=(libtinfo6)
+  else  
+    packages+=(libtinfo5)
+  fi
+  
   if [ "$UseClang" = true ]; then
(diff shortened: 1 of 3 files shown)


RunCMake.sh
```diff
@@ -92,8 +92,14 @@ verlt() {
 }
 
 if [ "$Distribution" = "Ubuntu" -a "$Version" = "22" ] || [ "$Distribution" = "Ubuntu" -a "$Version" = "24" ] || [ "$Distribution" = "Mint" -a "$Version" = "21" ]; then
-  packages=(cmake build-essential ninja-build libxrandr-dev libxinerama-dev libomp-dev libxcursor-dev libxi-dev uuid-dev mold libfreetype-dev libtinfo5 libomp-dev)
```
Ingrater, line 95: You did remove libomp-dev. Is this no longer needed when building with gcc on Ubuntu 22?

RunCMake.sh
```diff
@@ -92,8 +92,14 @@ verlt() {
...
+  packages=(cmake build-essential ninja-build libxrandr-dev libxinerama-dev libxcursor-dev libxi-dev uuid-dev mold libfreetype-dev libxkbcommon-dev)
 
+  if [ "$Distribution" = "Ubuntu" -a "$Version" = "24" ]; then
+    packages+=(libtinfo6)
+  else  
+    packages+=(libtinfo5)
+  fi
+  
   if [ "$UseClang" = true ]; then
     packages+=(clang-14 libomp-14-dev libstdc++-12-dev)
```
Ingrater, line 104: Ubuntu 24 is using clang 17, I think we should add a if statement here to use the clang-17 packages when running this script on ubuntu 24.

Ingrater (changes requested):

RunCMake.sh
Sanakan8472: Using clang-17 leads to both gdb and lldb being unable to set breakpoints on the build so I didn't switch over to it.

RunCMake.sh
Sanakan8472: The package is still present further down under clang packages as `libomp-14-dev`. We need to use the version that matches the clang version. Using `libomp-dev` points to the latest version which conflicts with `libomp-14-dev` as there can be only one version installed apparently.

PR #1475, opened 2025-01-29

jankrassnigg:


Code/Engine/Foundation/Basics/Platform/BlackMagic.h
```diff
@@ -111,6 +111,12 @@
   op(a0, a1), op(a2, a3), op(a4, a5), op(a6, a7), op(a8, a9), op(a10, a11), op(a12, a13), op(a14, a15), op(a16, a17)
 #define EZ_EXPAND_ARGS_PAIR_COMMA_20(op, a0, a1, a2, a3, a4, a5, a6, a7, a8, a9, a10, a11, a12, a13, a14, a15, a16, a17, a18, a19) \
   op(a0, a1), op(a2, a3), op(a4, a5), op(a6, a7), op(a8, a9), op(a10, a11), op(a12, a13), op(a14, a15), op(a16, a17), op(a18, a19)
+#define EZ_EXPAND_ARGS_PAIR_COMMA_22(op, a0, a1, a2, a3, a4, a5, a6, a7, a8, a9, a10, a11, a12, a13, a14, a15, a16, a17, a18, a19, a20, a21) \
+  op(a0, a1), op(a2, a3), op(a4, a5), op(a6, a7), op(a8, a9), op(a10, a11), op(a12, a13), op(a14, a15), op(a16, a17), op(a18, a19), op(a20, a21)
+#define EZ_EXPAND_ARGS_PAIR_COMMA_24(op, a0, a1, a2, a3, a4, a5, a6, a7, a8, a9, a10, a11, a12, a13, a14, a15, a16, a17, a18, a19, a20, a21, a22, a23) \
+  op(a0, a1), op(a2, a3), op(a4, a5), op(a6, a7), op(a8, a9), op(a10, a11), op(a12, a13), op(a14, a15), op(a16, a17), op(a18, a19), op(a20, a21), op(a22, a23)
+#define EZ_EXPAND_ARGS_PAIR_COMMA_25(op, a0, a1, a2, a3, a4, a5, a6, a7, a8, a9, a10, a11, a12, a13, a14, a15, a16, a17, a18, a19, a20, a21, a22, a23, a24, a25) \
```
C-Core: Should be EZ_EXPAND_ARGS_PAIR_COMMA_26

Code/Engine/Core/Scripting/ScriptClasses/Implementation/ScriptExtensionClass_Physics.cpp
```diff
@@ -0,0 +1,130 @@
...
+#include <Core/Scripting/ScriptAttributes.h>
+#include <Core/Scripting/ScriptClasses/ScriptExtensionClass_Physics.h>
+
+// clang-format off
+EZ_BEGIN_STATIC_REFLECTED_TYPE(ezScriptExtensionClass_Physics, ezNoBase, 1, ezRTTINoAllocator)
+{
+  EZ_BEGIN_FUNCTIONS
+  {
+    EZ_SCRIPT_FUNCTION_PROPERTY(GetGravity, In, "World"),
+    EZ_SCRIPT_FUNCTION_PROPERTY(Raycast, Out, "HitPosition", Out, "HitNormal", Out, "HitObject", In, "World", In, "Start", In, "Direction", In, "Distance", In, "CollisionLayer", In, "ShapeTypes")->AddAttributes(new ezFunctionArgumentAttributes(8, new ezDefaultValueAttribute((ezInt32)ezPhysicsShapeType::Static | (ezInt32)ezPhysicsShapeType::Dynamic))),
```
C-Core: Do we have an dynamic enum for collision layer?

Code/Engine/RendererCore/Debug/Implementation/DebugRenderer.cpp
```diff
@@ -2100,35 +2100,45 @@ EZ_BEGIN_STATIC_REFLECTED_TYPE(ezScriptExtensionClass_Debug, ezNoBase, 1, ezRTTI
   {
     EZ_SCRIPT_FUNCTION_PROPERTY(DrawCross, In, "World", In, "Position", In, "Size", In, "Color", In, "Transform")->AddAttributes(
       new ezFunctionArgumentAttributes(2, new ezDefaultValueAttribute(0.1f)),
-      new ezFunctionArgumentAttributes(3, new ezExposeColorAlphaAttribute())),
+      new ezFunctionArgumentAttributes(3, new ezDefaultValueAttribute(ezColor::White), new ezExposeColorAlphaAttribute()),
```
C-Core: I don't think that these default value attribute should be needed, for properties you also don't need to specify them when they have the type's default value. Should be the same here.

jankrassnigg: They are needed when auto generating function signatures in languages that work like C, ie when argument N has a default value, then argument N+1 also has to have a default value.

Also, for these functions they do make sense to have.

Code/Engine/Core/Scripting/ScriptClasses/Implementation/ScriptExtensionClass_Physics.cpp
jankrassnigg: Yes, added.

Code/Engine/RendererCore/Debug/Implementation/DebugRenderer.cpp
jankrassnigg: removed some of them again

PR #1476, opened 2025-01-30
Fixed particle culling bug, improved ezRaycastComponent

jankrassnigg:


Code/EnginePlugins/ParticlePlugin/System/ParticleSystemInstance.cpp
```diff
@@ -342,6 +342,7 @@ ezParticleSystemInstance::ezParticleSystemInstance()
 
 void ezParticleSystemInstance::Construct(ezUInt32 uiMaxParticles, ezWorld* pWorld, ezParticleEffectInstance* pOwnerEffect, float fSpawnCountMultiplier)
 {
+  m_BoundingVolume = ezBoundingBoxSphere::MakeInvalid();
```
jankrassnigg, line 345: ezParticleSystemInstance's are reused, thus not resetting the variable would mean new instances used values from previous runs, so they might show up in the wrong location in the first few frames, and then deemed to be invisible, and then not updated...

Took me several hours to find this issue. It was mostly showing up with the gas cylinders, where the fire would not show up.

Code/EnginePlugins/GameComponentsPlugin/Gameplay/Implementation/RaycastComponent.cpp
```diff
@@ -54,10 +55,16 @@ EZ_BEGIN_COMPONENT_TYPE(ezRaycastComponent, 3, ezComponentMode::Static)
...
     EZ_BITFLAGS_MEMBER_PROPERTY("ShapeTypesToHit", ezPhysicsShapeType, m_ShapeTypesToHit)->AddAttributes(new ezDefaultValueAttribute(ezVariant(ezPhysicsShapeType::Default & ~(ezPhysicsShapeType::Trigger)))),
     EZ_MEMBER_PROPERTY("CollisionLayerEndPoint", m_uiCollisionLayerEndPoint)->AddAttributes(new ezDynamicEnumAttribute("PhysicsCollisionLayer")),
-    EZ_MEMBER_PROPERTY("CollisionLayerTrigger", m_uiCollisionLayerTrigger)->AddAttributes(new ezDynamicEnumAttribute("PhysicsCollisionLayer")),
-    EZ_ACCESSOR_PROPERTY("TriggerMessage", GetTriggerMessage, SetTriggerMessage),
+    EZ_MEMBER_PROPERTY("ChangeNotificationMsg", m_sChangeNotificationMsg),
   }
   EZ_END_PROPERTIES;
+  EZ_BEGIN_FUNCTIONS
+  {
+    EZ_SCRIPT_FUNCTION_PROPERTY(GetCurrentDistance)->AddFlags(ezPropertyFlags::Const),
```
C-Core, line 63: There is no need to explicitely specify the const flag if the function itself is const already. Then the const flag is already added by our reflection automatically.

C-Core (approved):

PR #1483, opened 2025-02-10
Simpler way to update buffers and textures from game code

C-Core:
* Added two new methods to GALDevice: UpdateBufferForNextFrame and UpdateTextureForNextFrame. These can be directly called from game code or during extraction to update a buffer or texture for the next render frame. E.g. this is now used for skinning buffer update or for the occlusion debug texture.
* Removed CopyToTempBuffer update mode for buffers since this is now basically replaced with the new methods above and was not used in our code anyways.
* Added a max draw distance to the debug text component which can be used to prevent clutter from too much debug text on screen.
* Various other smaller fixes and optimizations.

### DIFF
--- a/Code/Engine/RendererCore/Debug/DebugTextComponent.h
+++ b/Code/Engine/RendererCore/Debug/DebugTextComponent.h
@@ -32,7 +32,7 @@ public:
   float m_fValue2 = 0.0f;                                         // [ property ]
   float m_fValue3 = 0.0f;                                         // [ property ]
 
-  float m_fMaxDistance = 10.0f;                                  // [ property ]
+  float m_fMaxDistance = 10.0f;                                   // [ property ]
 
 protected:
   void OnMsgExtractRenderData(ezMsgExtractRenderData& msg) const; // [ msg handler ]
--- a/Code/Engine/RendererCore/Debug/DebugTextComponent.h
+++ b/Code/Engine/RendererCore/Debug/DebugTextComponent.h
@@ -32,7 +32,7 @@ public:
   float m_fValue2 = 0.0f;                                         // [ property ]
   float m_fValue3 = 0.0f;                                         // [ property ]
 
-  float m_fMaxDistance = 100.0f;                                  // [ property ]
+  float m_fMaxDistance = 10.0f;                                  // [ property ]
 
 protected:
   void OnMsgExtractRenderData(ezMsgExtractRenderData& msg) const; // [ msg handler ]
--- a/Code/Engine/RendererCore/Debug/DebugTextComponent.h
+++ b/Code/Engine/RendererCore/Debug/DebugTextComponent.h
@@ -32,6 +32,8 @@ public:
   float m_fValue2 = 0.0f;                                         // [ property ]
   float m_fValue3 = 0.0f;                                         // [ property ]
 
+  float m_fMaxDistance = 100.0f;                                  // [ property ]
+
 protected:
   void OnMsgExtractRenderData(ezMsgExtractRenderData& msg) const; // [ msg handler ]
 };
--- a/Code/Engine/RendererCore/Debug/Implementation/DebugTextComponent.cpp
+++ b/Code/Engine/RendererCore/Debug/Implementation/DebugTextComponent.cpp
@@ -95,7 +95,7 @@ void ezDebugTextComponent::OnMsgExtractRenderData(ezMsgExtractRenderData& msg) c
   c.a *= fFade;
 
   ezStringBuilder sb;
-  sb.SetFormat(m_sText, m_fValue0, m_fValue1, m_fValue2, m_fValue3);  
+  sb.SetFormat(m_sText, m_fValue0, m_fValue1, m_fValue2, m_fValue3);
 
   ezDebugRenderer::Draw3DText(msg.m_pView->GetHandle(), sb, GetOwner()->GetGlobalPosition(), c);
 }
--- a/Code/Engine/RendererCore/Debug/Implementation/DebugTextComponent.cpp
+++ b/Code/Engine/RendererCore/Debug/Implementation/DebugTextComponent.cpp
@@ -85,8 +85,8 @@ void ezDebugTextComponent::OnMsgExtractRenderData(ezMsgExtractRenderData& msg) c
   if (m_sText.IsEmpty())
     return;
 
-  const float fDistance = msg.m_pView->GetCullingCamera()->GetCenterPosition().GetDistanceTo(GetOwner()->GetGlobalPosition());
-  if (fDistance > m_fMaxDistance)
+  const float fSquaredDistance = msg.m_pView->GetCullingCamera()->GetCenterPosition().GetSquaredDistanceTo(GetOwner()->GetGlobalPosition());
+  if (fSquaredDistance > m_fMaxDistance * m_fMaxDistance)
     return;
 
   ezStringBuilder sb;
--- a/Code/Engine/RendererCore/Debug/Implementation/DebugTextComponent.cpp
+++ b/Code/Engine/RendererCore/Debug/Implementation/DebugTextComponent.cpp
@@ -18,7 +18,7 @@ EZ_BEGIN_COMPONENT_TYPE(ezDebugTextComponent, 2, ezComponentMode::Static)
     EZ_MEMBER_PROPERTY("Value2", m_fValue2),
     EZ_MEMBER_PROPERTY("Value3", m_fValue3),
     EZ_MEMBER_PROPERTY("Color", m_Color),
-    EZ_MEMBER_PROPERTY("MaxDistance", m_fMaxDistance)->AddAttributes(new ezDefaultValueAttribute(100.0f), new ezClampValueAttribute(0.0f, ezVariant())),
+    EZ_MEMBER_PROPERTY("MaxDistance", m_fMaxDistance)->AddAttributes(new ezDefaultValueAttribute(10.0f), new ezClampValueAttribute(0.0f, ezVariant())),
   }
   EZ_END_PROPERTIES;
   EZ_BEGIN_MESSAGEHANDLERS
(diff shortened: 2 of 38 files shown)


Code/Engine/RendererCore/Debug/DebugTextComponent.h
```diff
@@ -32,6 +32,8 @@ class EZ_RENDERERCORE_DLL ezDebugTextComponent : public ezComponent
   float m_fValue2 = 0.0f;                                         // [ property ]
   float m_fValue3 = 0.0f;                                         // [ property ]
 
+  float m_fMaxDistance = 100.0f;                                  // [ property ]
```
jankrassnigg: 100m is already pretty far way. The default could probably be significantly lower (somewhere between 10 and 25 meters maybe).

Code/Engine/RendererCore/Debug/Implementation/DebugTextComponent.cpp
```diff
@@ -68,20 +70,29 @@ void ezDebugTextComponent::DeserializeComponent(ezWorldReader& inout_stream)
...
-  if (!m_sText.IsEmpty())
-  {
-    ezStringBuilder sb;
-    sb.SetFormat(m_sText, m_fValue0, m_fValue1, m_fValue2, m_fValue3);
+  if (m_sText.IsEmpty())
+    return;
 
-    ezDebugRenderer::Draw3DText(msg.m_pView->GetHandle(), sb, GetOwner()->GetGlobalPosition(), m_Color);
-  }
+  const float fDistance = msg.m_pView->GetCullingCamera()->GetCenterPosition().GetDistanceTo(GetOwner()->GetGlobalPosition());
```
jankrassnigg: GetSquaredDistanceTo

Code/Engine/RendererCore/Meshes/Implementation/SkinnedMeshComponent.cpp
```diff
@@ -54,30 +52,10 @@ void ezSkinningState::TransformsChanged()
...
-  const ezUInt32 uiExIdx = ezRenderWorld::GetDataIndexForExtraction();
-
-  if (m_bTransformsUpdated[uiExIdx] && *m_bTransformsUpdated[uiExIdx] == false)
-  {
-    auto pSkinningMatrices = EZ_NEW_ARRAY(ezFrameAllocator::GetCurrentAllocator(), ezShaderTransform, m_Transforms.GetCount());
-    pSkinningMatrices.CopyFrom(m_Transforms);
-
-    ref_renderData.m_pNewSkinningTransformData = pSkinningMatrices.ToByteArray();
-    ref_renderData.m_bTransformsUpdated = m_bTransformsUpdated[uiExIdx];
+    ezGALDevice::GetDefaultDevice()->UpdateBufferForNextFrame(m_hGpuBuffer, m_Transforms.GetByteArrayPtr());
```
jankrassnigg, line 58: Does this introduce a frame delay ?
Generally, does this affect performance in any way?

Code/Engine/RendererVulkan/Device/Implementation/DeviceVulkan.cpp
```diff
@@ -1344,6 +1341,20 @@ void ezGALDeviceVulkan::DestroyVertexDeclarationPlatform(ezGALVertexDeclaration*
...
+  const ezGALBufferVulkan* pBufferVulkan = static_cast<const ezGALBufferVulkan*>(pBuffer);
+
+  m_pInitContext->UpdateBuffer(pBufferVulkan, uiDestOffset, sourceData);
+}
+
+void ezGALDeviceVulkan::UpdateTextureForNextFramePlatform(const ezGALTexture* pTexture, const ezGALSystemMemoryDescription& sourceData, const ezGALTextureSubresource& destinationSubResource, const ezBoundingBoxu32& destinationBox)
+{
+  const ezGALTextureVulkan* pTextureVulkan = static_cast<const ezGALTextureVulkan*>(pTexture);
+
+  // m_pInitContext->UpdateTexture(pTextureVulkan, sourceData, destinationSubResource, destinationBox);
```
jankrassnigg: TODO ?

Code/Engine/RendererCore/Debug/DebugTextComponent.h
C-Core: I chose 100 to not break existing behavior too much, so people don't wonder where their debug text went.

Code/Engine/RendererCore/Meshes/Implementation/SkinnedMeshComponent.cpp
C-Core: No it does not introduce a frame delay and it allows us to have even better performance in the future than before. Although atm the performance is exactly the same.

Code/Engine/RendererVulkan/Device/Implementation/DeviceVulkan.cpp
C-Core: Already fixed.

Code/Engine/RendererCore/Debug/DebugTextComponent.h
jankrassnigg: I don't think there's much existing data that we need to worry about. But for example the Physics test scene is quite annoying with all the debug text, and I think that's a good reference to set a useful default value.

C-Core: Alright, reduced it to 10 and added some fading.

jankrassnigg (approved):

Code/Engine/RendererFoundation/Device/Implementation/Device.cpp
```diff
@@ -931,6 +933,76 @@ void ezGALDevice::DestroyReadbackBuffer(ezGALReadbackBufferHandle hBuffer)
...
+
+    const ezUInt32 uiWidth = bDestBoxIsValid ? ezMath::Max(destinationBox.m_vMax.x - destinationBox.m_vMin.x, 1u) : desc.m_uiWidth;
+    const ezUInt32 uiHeight = bDestBoxIsValid ? ezMath::Max(destinationBox.m_vMax.y - destinationBox.m_vMin.y, 1u) : desc.m_uiHeight;
+    const ezUInt32 uiDepth = bDestBoxIsValid ? ezMath::Max(destinationBox.m_vMax.z - destinationBox.m_vMin.z, 1u) : desc.m_uiDepth;
+
+    const ezUInt32 uiRowPitch = uiWidth * ezGALResourceFormat::GetBitsPerElement(desc.m_Format) / 8;
+    const ezUInt32 uiSlicePitch = uiRowPitch * uiHeight;
+    EZ_ASSERT_DEV(sourceData.m_uiRowPitch == uiRowPitch, "Invalid row pitch. Expected {0} got {1}", uiRowPitch, sourceData.m_uiRowPitch);
+    EZ_ASSERT_DEV(sourceData.m_uiSlicePitch == 0 || sourceData.m_uiSlicePitch == uiSlicePitch, "Invalid slice pitch. Expected {0} got {1}", uiSlicePitch, sourceData.m_uiSlicePitch);
+    EZ_ASSERT_DEBUG(sourceData.m_pData.GetCount() >= uiSlicePitch * uiDepth, "Not enough data provided to update texture");
```
Sanakan8472: We have `ezLog::Error`, `EZ_ASSERT_DEV` and `EZ_ASSERT_DEBUG` in use in this function for validating input. Should we pick just one? As `UpdateBufferForNextFrame` is just using  `ezLog::Error`, should we stick to the same for textures?

Code/Engine/RendererDX11/Device/Implementation/DeviceDX11.cpp
```diff
@@ -691,6 +692,62 @@ void ezGALDeviceDX11::DestroyVertexDeclarationPlatform(ezGALVertexDeclaration* p
...
 }
 
+void ezGALDeviceDX11::UpdateBufferForNextFramePlatform(const ezGALBuffer* pBuffer, ezConstByteArrayPtr sourceData, ezUInt32 uiDestOffset)
+{
+  const ezGALBufferDX11* pBufferDX11 = static_cast<const ezGALBufferDX11*>(pBuffer);
+
+#if EZ_ENABLED(EZ_COMPILE_FOR_DEBUG)
+  for (auto& copy : m_PendingCopies)
+  {
+    if (copy.m_pDestResource == pBufferDX11->GetDXBuffer())
```
Sanakan8472, line 702: Can we allow updating the same buffer twice in a frame if the ranges don't overlap? For larger structured buffers it would be beneficial to only update parts of the buffer, which aren't necesarily in one block. If we don't have a use case for that yet, we can also extend it in a later PR?

Code/Engine/RendererDX11/Device/DeviceDX11.h
```diff
@@ -220,9 +241,20 @@ class EZ_RENDERERDX11_DLL ezGALDeviceDX11 : public ezGALDevice
...
+  struct PendingCopy
+  {
+    TempResource m_SourceResource = {};
+    ID3D11Resource* m_pDestResource = nullptr;
+    ezUInt32 m_uiDestSubResource = 0;
+    ezVec3U32 m_vDestPoint = ezVec3U32::MakeZero();
+    ezVec3U32 m_vSourceSize = ezVec3U32::MakeZero();
+  };
+
+  ezDynamicArray<PendingCopy, ezLocalAllocatorWrapper> m_PendingCopies;
```
Sanakan8472, line 256: Just for confirmation, we can get away with not double buffering this because `BeginFramePlatform` is called before the parallel rendering / update phase, so there is no risk of race conditions?

Code/Engine/RendererVulkan/Device/Implementation/DeviceVulkan.cpp
```diff
@@ -1344,6 +1341,20 @@ void ezGALDeviceVulkan::DestroyVertexDeclarationPlatform(ezGALVertexDeclaration*
   EZ_DELETE(&m_Allocator, pVertexDeclarationVulkan);
 }
 
+void ezGALDeviceVulkan::UpdateBufferForNextFramePlatform(const ezGALBuffer* pBuffer, ezConstByteArrayPtr sourceData, ezUInt32 uiDestOffset)
+{
+  const ezGALBufferVulkan* pBufferVulkan = static_cast<const ezGALBufferVulkan*>(pBuffer);
+
+  m_pInitContext->UpdateBuffer(pBufferVulkan, uiDestOffset, sourceData);
```
Sanakan8472: We can't use `m_pInitContext` for this purpose as this instance is flushed on every submit call, not the start of the frame. This is needed because all resource creation end up in this context and we expect to be able to create resources or block on load in the render code and immediately make use of them.
Thus, in order to ensure these actually end up in the next frame, we need to create a separate instance that is only flushed once at the start of the frame.

Sanakan8472: This is a bit troublesome as we have the following dependencies when creating a resource in the game loop and calling `UpdateBufferForNextFrame`:
1. The resource creation ends up in the `m_pInitContext`.
2. The `UpdateBufferForNextFrame` needs to go in a separate context.
At the start of the frame, before we enter the parallel phase, we need to make one submit call with the command buffer from the `m_pInitContext` followed by the command buffer with all the `UpdateBufferForNextFrame` calls so that resource creation happens before updates.

Sanakan8472 (changes requested):

Code/Engine/RendererDX11/Device/DeviceDX11.h
C-Core: Correct, begin and end frame are fully synchronous with no update running in parallel.

Sanakan8472 (approved):

PR #1486, opened 2025-02-14
AngelScript: #include dependencies, error line mapping

jankrassnigg:
* Asset now tracks #include dependencies and recompiles script on demand
* Asset output now uses compression
* Compiler errors are now correctly translated to the file and line where they occurred
* Update function now gets the time delta
* Added "throw" for critical errors
* Added support for game object and component handles in properties
* fixed an issue that made Jolt crash when applying a force to a just created object

### DIFF
--- a/Data/Samples/SampleGame/TypeScript/ez/AllComponents.ts
+++ b/Data/Samples/SampleGame/TypeScript/ez/AllComponents.ts
@@ -324,6 +324,8 @@ export class DebugTextComponent extends Component
   set Value3(value: number) { __CPP_ComponentProperty_set(this, 1816166542, value); }
   get Color(): Color { return __CPP_ComponentProperty_get(this, 1881354469); }
   set Color(value: Color) { __CPP_ComponentProperty_set(this, 1881354469, value); }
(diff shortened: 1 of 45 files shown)


Data/Samples/SampleGame/TypeScript/ez/AllComponents.ts
```diff
@@ -324,6 +324,8 @@ export class DebugTextComponent extends Component
   set Value3(value: number) { __CPP_ComponentProperty_set(this, 1816166542, value); }
   get Color(): Color { return __CPP_ComponentProperty_get(this, 1881354469); }
   set Color(value: Color) { __CPP_ComponentProperty_set(this, 1881354469, value); }
+  get MaxDistance(): number { return __CPP_ComponentProperty_get(this, 1272349630); }
```
C-Core, line 327: Remove typescript stuff?

jankrassnigg: That's the sample game project. For now it still uses TS, but once I've added array support, it will be replaced as well.

PR #1487, opened 2025-02-17
Added GAL dynamic buffer

C-Core:
* Added a new "buffer" type called GAL dynamic buffer which can be used when a lot of data should be stored in a single large buffer with a dynamic size. It supports allocation/deallocation of single elements or ranges of multiple elements. There is also a compaction step which can move allocations around to reduce fragmentation over time. Only a dirty range of changed elements is then uploaded to the GPU buffer.
* In this PR a dynamic buffer is only used for procedural vertex colors but the plan is to use it for much more in the future, e.g. instance data, skinning data etc.

### DIFF
--- a/Code/Engine/RendererFoundation/Resources/DynamicBuffer.h
+++ b/Code/Engine/RendererFoundation/Resources/DynamicBuffer.h
@@ -70,6 +70,7 @@ private:
   void Resize(ezUInt32 uiNewSize);
 
   ezDynamicArray<ezUInt8, ezAlignedAllocatorWrapper> m_Data;
+  ezUInt32 m_uiNextOffset = 0;
 
   struct Allocation
   {
--- a/Code/Engine/RendererFoundation/Resources/DynamicBuffer.h
+++ b/Code/Engine/RendererFoundation/Resources/DynamicBuffer.h
@@ -50,13 +50,13 @@ public:
   };
 
   /// \brief Tries to compact the buffer by moving allocations to free ranges. All moved allocations are returned in out_changedAllocations.
-  /// 
+  ///
   /// The user data can be used to update the owner of the allocation.
   /// To prevent too many changes per frame only uiMaxSteps are executed which corresponds to the number of allocations which can be moved.
   void RunCompactionSteps(ezDynamicArray<ChangedAllocation>& out_changedAllocations, ezUInt32 uiMaxSteps = 16);
 
   /// \brief This should be called inside the rendering code to retrieve the underlying buffer for rendering.
-  /// 
+  ///
   /// It is ensured that it will always return the same buffer until the next time BeginFrame is called on the GALDevice even if the buffer
   /// has been resized due to more allocations on the game play or extraction side.
   const ezGALBufferHandle& GetBufferForRendering() const
--- a/Code/Engine/RendererFoundation/Resources/DynamicBuffer.h
+++ b/Code/Engine/RendererFoundation/Resources/DynamicBuffer.h
@@ -2,11 +2,22 @@
 
 #include <RendererFoundation/Descriptors/Descriptors.h>
 
+/// \brief A dynamic buffer can be used when a lot of data needs to be stored in a single large buffer with dynamic size.
+///
+/// This class supports allocation and deallocation of single elements or ranges of multiple elements.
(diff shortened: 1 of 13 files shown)


Code/UnitTests/RendererTest/Basics/DynamicBufferTest.cpp
```diff
@@ -0,0 +1,182 @@
...
+      EZ_TEST_INT(pMappedData.GetCount(), uiAllocationSizes[uiIndex]);
+
+      uiTotalOffset += uiAllocationSizes[uiIndex];
+    };
+
+    AllocateAndMap(0);
+
+    pDynamicBuffer->UploadChanges();
+
+    // Internal buffer should be have been created after upload but GetBufferForRendering should still return the old handle until the next frame.
```
jankrassnigg:
```suggestion
    // Internal buffer should have been created after upload but GetBufferForRendering should still return the old handle until the next frame.
```

Code/Engine/RendererFoundation/Resources/DynamicBuffer.h
```diff
@@ -0,0 +1,88 @@
+#pragma once
+
+#include <RendererFoundation/Descriptors/Descriptors.h>
+
+class EZ_RENDERERFOUNDATION_DLL ezGALDynamicBuffer
```
jankrassnigg, line 12: Add a short description for the class's purpose.

Code/Engine/RendererFoundation/Resources/DynamicBuffer.h
```diff
@@ -0,0 +1,88 @@
...
+
+  void UploadChanges();
+
+  struct ChangedAllocation
+  {
+    ezUInt64 m_uiUserData = 0;
+    ezUInt32 m_uiNewOffset = 0;
+  };
+
+  void RunCompactionSteps(ezDynamicArray<ChangedAllocation>& out_changedAllocations, ezUInt32 uiMaxSteps = 16);
```
jankrassnigg, line 56: Needs documentation what this is for and how to use it.

Code/Engine/RendererFoundation/Resources/DynamicBuffer.h
```diff
@@ -0,0 +1,88 @@
...
+
+  const ezGALBufferHandle& GetBufferForRendering() const
```
jankrassnigg, line 62: Needs documentation that describes when you get which handle (frame begin etc)

Code/Engine/RendererFoundation/Resources/DynamicBuffer.h
```diff
@@ -0,0 +1,88 @@
...
+  template <typename T>
+  ezArrayPtr<T> MapForWriting(ezUInt32 uiOffset)
+  {
+    ezUInt32 uiCount = 0;
+    ezByteArrayPtr byteData = MapForWriting(uiOffset, uiCount);
+    EZ_ASSERT_DEBUG(sizeof(T) == m_Desc.m_uiStructSize, "Invalid Type");
+    return ezArrayPtr<T>(reinterpret_cast<T*>(byteData.GetPtr()), uiCount);
+  }
+
+  void UploadChanges();
```
jankrassnigg: Add docs for when the changes are visible.

Code/Engine/RendererFoundation/Resources/DynamicBuffer.h
```diff
@@ -0,0 +1,88 @@
...
+  ezUInt32 Allocate(const U& userData, ezUInt32 uiCount)
+  {
+    static_assert(sizeof(U) <= sizeof(ezUInt64), "userData is too large");
+    ezUInt64 uiUserData = 0;
+    *reinterpret_cast<U*>(&uiUserData) = userData;
+
+    return Allocate(uiUserData, uiCount);
+  }
+
+  void Deallocate(ezUInt32 uiOffset);
```
jankrassnigg, line 31: Add docs talking about holes / compaction. Also why can I only deallocate at an offset? Shouldn't I also be able to give a range?

C-Core: An allocation is identified by its offset just like you would identify a regular allocation by its pointer.

PR #1489, opened 2025-02-18
Removed batch id from render data

C-Core:
* The batch id has been removed from render data. It was weird to work with and had no checks for potential collisions. A new virtual CanBatch method has been added to render data which can just do good old comparisons of all relevant members to determine whether two render data can be batched together.
* Optimized clustered data extraction. The bounding spheres around the cluster cells are now in viewspace and don't need to be recomputed every frame. Cluster bitmasks are directly cleared after use so the expensive zerofill calls could be removed.
* Cleaned up some profiling scopes and other smaller optimizations.

### DIFF
--- a/Code/Engine/RendererCore/ShaderCompiler/Implementation/ShaderManager.cpp
+++ b/Code/Engine/RendererCore/ShaderCompiler/Implementation/ShaderManager.cpp
@@ -105,7 +105,7 @@ namespace
     return false;
   }
 
-  static ezHashTable<ezUInt64, ezString> s_PermutationPaths;
+  static ezHashTable<ezUInt64, ezUntrackedString> s_PermutationPaths;
 } // namespace
 
 //////////////////////////////////////////////////////////////////////////
@@ -330,8 +330,8 @@ ezShaderPermutationResourceHandle ezShaderManager::PreloadSinglePermutationInter
 {
   const ezUInt64 uiPermutationKey = (ezUInt64)ezHashingUtils::StringHashTo32(uiResourceIdHash) << 32 | uiPermutationHash;
 
-  ezString* pPermutationPath = &s_PermutationPaths[uiPermutationKey];
-  if (pPermutationPath->IsEmpty())
+  ezUntrackedString& permutationPath = s_PermutationPaths[uiPermutationKey];
+  if (permutationPath.IsEmpty())
   {
     ezStringBuilder sShaderFile = GetCacheDirectory();
     sShaderFile.AppendPath(GetActivePlatform().GetData());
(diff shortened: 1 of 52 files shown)


Code/Engine/RendererCore/ShaderCompiler/Implementation/ShaderManager.cpp
```diff
@@ -105,7 +105,7 @@ namespace
     return false;
   }
 
-  static ezHashTable<ezUInt64, ezString> s_PermutationPaths;
+  static ezHashTable<ezUInt64, ezUntrackedString> s_PermutationPaths;
```
jankrassnigg, line 108: why is this necessary?

Code/EnginePlugins/KrautPlugin/Renderer/KrautRenderer.cpp
```diff
@@ -22,6 +22,15 @@ EZ_BEGIN_DYNAMIC_REFLECTED_TYPE(ezKrautRenderer, 1, ezRTTIDefaultAllocator<ezKra
 EZ_END_DYNAMIC_REFLECTED_TYPE;
 // clang-format on
 
+bool ezKrautRenderData::CanBatch(const ezRenderData& other0) const
+{
+  const auto& other = ezStaticCast<const ezKrautRenderData&>(other0);
```
jankrassnigg, line 27: what about the material ?

Code/EnginePlugins/ParticlePlugin/Declarations.h
```diff
@@ -82,8 +82,6 @@ struct EZ_PARTICLEPLUGIN_DLL ezParticleTypeRenderMode
     Blended,
     Opaque,
     Distortion,
-    BlendedBackground,
```
jankrassnigg: this is a user facing setting, needed to tweak behavior

Code/Engine/RendererCore/ShaderCompiler/Implementation/ShaderManager.cpp
C-Core: Technically it is not necessary, the root leak detection will not report the string allocations as leaks. But better not track the memory at all.

Code/EnginePlugins/KrautPlugin/Renderer/KrautRenderer.cpp
C-Core: The material comes from the mesh and cannot be overwritten by the component so it is enough to only check the mesh here.

Code/EnginePlugins/ParticlePlugin/Declarations.h
C-Core: Reverted

PR #1493, opened 2025-02-22
Introduce global transform update by flag

AlexandrKedrov:
- Introduce Bitfield to `World::Hierarchy` struct. This struct is used to identify which objects' transforms have been changed since previous update during `WorldData::UpdateGlobalTransforms`. Thus only affected object's transforms are recalculated.
- Introduce `TransformData::m_uiIndex` which corresponds to the index in the bitfield. Currently it is the same as `GameObject::m_InternalId.m_InstanceIndex`.
- Modify global transform operations of GameObject: they set "changed" flag for all children objects of the modified object. (We set global transform manually, so transform shouldn't be recalculated, but children's transforms are still should)
- Modify local transform operations of GameObject: set "changed" flag for the current object. Its global transform will be racalculated

These changes help to solve following problems:
- Unnecessary updates. If we collect thousands of objects in one group, these objects transform will be recalculate despite their parent is unlikely to transform (many devs use groups only for easier navigation in the editor)
- We can control which object's transform should be recalculated. If we changed a game object's global transform manually, this transform was recalulculated based on its local transform during `WorldData::UpdateGlobalTransforms`. It could lead to fast error accumulation due to precision loss.

My current implementation is a bit "rusty". I propose following changes (if this idea is supported) to make it more complete.
- [ ]  Use separate index storage only for dynamic transform data, because it is the only data which is affected by UpdateGlobalTransforms. Potentially, it will reduce memory usage and improve perfomance.
- [ ] Current implementation controls only transform recalculations, but global bounds and spatials are still calculated for all objects. Technically, we can apply the same idea for them too for better perfomance, yet we need more flags. The "nicest" solution in my opinion is to introduce additional flag member to `TransformData`, so the logic is the following:
_if the value is set in the bitfield then the "visitor" should check `TransformData::flag`, apply actions (commit or refuse the recalculation of new global bounds and transforms) and immediately reset the `TransformData::flag = 0`._ This solution provides flexibility (we can use multiple flags and quickly reset them) and we will still be able to pass "changed" status from parent to its children. Alternatively we can make bitfield just two times larger :)

Prefabs and scenes used for benchmarking:
[Benchmark.zip](https://github.com/user-attachments/files/18925093/Benchmark.zip)


C-Core:
As discussed on discord, this change has a performance implication that we are not willing to pay.

PR #1505, opened 2025-03-09
AngelScript Linux support

Sanakan8472:
# Engine Changes
* `ezAngelScriptEngineSingleton::RegisterPodType` takes another argument with additional flags, needed for value types in AngelScript on Linux.
* Fixed type registration and various function declarations to also work with Linux calling conventions.
* Trim `"` from command line arguments in `ezCommandLineUtils::GetAbsolutePathOption`.
* Extracted `ezAngelScriptEngineSingleton::PreprocessCode` and `ezAngelScriptEngineSingleton::FindCorrectSectionAndLine` to allow reuse in test code.
# Platform Changes
* Disable mold on Android as it does not support `--undefined-glob` parameter on the version 2.30.0 that ships with Ubuntu 24 and is needed to prevent the linker from stripping linked plugins.
# Vulkan Changes
* Fixed use after free when update next frame texture / buffer updates where executed but the target resource was deleted and flushed in the previous frame, e.g. due to a window being destroyed.
  * The calls to `UpdateBufferForNextFramePlatform` and `UpdateTextureForNextFramePlatform` upload to the staging buffer right away, but then store `ezPendingBufferCopyVulkan` or `ezPendingTextureCopyVulkan` to be executed later.
  * At the start of each frame, `ezGALDeviceVulkan::Submit` will call `ezInitContextVulkan::ExecutePendingCopies`.
  * If in the meantime a buffer or texture is destroyed, its entries are removed from the stored pending copies, preventing the use after free.
# Test Fixes
* Enable EditorProcessor tests in GameEngineTests on Linux.
* Make SubstanceTest not assert on Linux.
* `StringsTests.as` and `TypesTest.as` have been moved into the `Types` folder. These files are no longer tested via the messaging system but the scripts are directly executed by the `ezGameEngineTestApplication_AngelScript::RunTestScript` function. This allows for better error output when debugging native calling convention issues.


Code/Engine/Foundation/Utilities/Implementation/CommandLineUtils.cpp
```diff
@@ -205,7 +205,7 @@ ezStringView ezCommandLineUtils::GetStringOption(ezStringView sOption, ezUInt32
 const ezString ezCommandLineUtils::GetAbsolutePathOption(ezStringView sOption, ezUInt32 uiArgument /*= 0*/, ezStringView sDefault /*= {} */, bool bCaseSensitive /*= false*/) const
 {
   ezStringView sPath = GetStringOption(sOption, uiArgument, sDefault, bCaseSensitive);
-
+  sPath.Trim("\"");
```
jankrassnigg: I'm surprised that this should ever happen. GetStringOption should already not contain quotes. Can you double check why this happened for you? And if this is really needed, it should probably already be in GetStringOption.

Code/EnginePlugins/AngelScriptPlugin/TypeRegistration/AsRegisterTypes_String.cpp
```diff
@@ -130,8 +135,8 @@ void ezAngelScriptEngineSingleton::Register_StringView()
   AS_CHECK(m_pEngine->RegisterObjectMethod("ezStringView", "bool TrimWordStart(ezStringView sWord)", asMETHOD(ezStringView, TrimWordStart), asCALL_THISCALL));
   AS_CHECK(m_pEngine->RegisterObjectMethod("ezStringView", "bool TrimWordEnd(ezStringView sWord)", asMETHOD(ezStringView, TrimWordEnd), asCALL_THISCALL));
 
-  AS_CHECK(m_pEngine->RegisterObjectMethod("ezStringView", "bool opEquals(ezStringView) const", asFUNCTIONPR(operator==, (ezStringView, ezStringView), bool), asCALL_CDECL_OBJFIRST));
-  AS_CHECK(m_pEngine->RegisterObjectMethod("ezStringView", "int opCmp(ezStringView) const", asFUNCTIONPR(ezStringView_opCmp, (ezStringView, ezStringView), int), asCALL_CDECL_OBJFIRST));
+  AS_CHECK(m_pEngine->RegisterObjectMethod("ezStringView", "bool opEquals(const ezStringView& in)", asFUNCTIONPR(ezStringView_opEqual, (ezStringView*, const ezStringView&), bool), asCALL_CDECL_OBJFIRST));
```
jankrassnigg: opEquals should probably still be const

Code/EnginePlugins/AngelScriptPlugin/Runtime/AsCompilation.cpp
```diff
@@ -191,8 +195,21 @@ asIScriptModule* ezAngelScriptEngineSingleton::CompileModule(ezStringView sModul
   if (asPP.Process(fullCode).Failed())
   {
     ezLog::Error("Failed to pre-process AngelScript");
```
jankrassnigg: If there is an ezResult return type, then the function shouldn't also log an error. Instead the caller should log on failure.

Code/UnitTests/GameEngineTest/AngelScriptTest/AngelScriptTest.cpp
```diff
@@ -87,4 +98,86 @@ ezTestAppRun ezGameEngineTestApplication_AngelScript::SubTestBasisExec(const cha
...
+    ezStringBuilder sTestCode;
+    ezStringBuilder sProcessedCode;
+    {
+      ezFileReader read;
+      if (read.Open(sScriptPath).Failed())
+      {
+        ezLog::Error("Failed to open file '{}'.", sScriptPath);
+        return;
+      }
+      sTestCode.Clear();
```
jankrassnigg: the clear is redundant

Code/Engine/RendererVulkan/Device/Implementation/DeviceVulkan.cpp
```diff
@@ -1133,6 +1129,11 @@ ezGALBuffer* ezGALDeviceVulkan::CreateBufferPlatform(
 void ezGALDeviceVulkan::DestroyBufferPlatform(ezGALBuffer* pBuffer)
 {
   ezGALBufferVulkan* pVulkanBuffer = static_cast<ezGALBufferVulkan*>(pBuffer);
+  for (ezInt32 i = (ezInt32)m_PendingBufferCopies.GetCount() - 1; i >= 0; --i)
```
C-Core: Since you do a swap with last anyways you don't need to iterate backwards I guess? Swap with last should be fine since the order doesn't really matter, we don't allow overlapping updates.

C-Core (approved):
Vulkan changes look good

Code/EnginePlugins/AngelScriptPlugin/TypeRegistration/AsRegisterTypes_String.cpp
Sanakan8472: Oh right! Thanks for noticing. I'm still getting a hang of angel script function declarations.

Code/EnginePlugins/AngelScriptPlugin/Runtime/AsCompilation.cpp
Sanakan8472: Good point! Fixed.

Code/Engine/RendererVulkan/Device/Implementation/DeviceVulkan.cpp
Sanakan8472: Fixed! Not sure why I used a backwards loop. As you said, order is not important and I was already not maintaining the order anyways.

Code/Engine/Foundation/Utilities/Implementation/CommandLineUtils.cpp
Sanakan8472: So when I enabled the asset processor tests in GameEngineTest under Linux, starting the `ezEditorProcessor` is done via `opt.AddArgument("\"{0}\"", sProjectDir);` which adds the `"`. On Linux, these are verbatim forwarded to the command line of the process so when calling `CreateOrOpenProject(false, pCmd->GetAbsolutePathOption("-project")).IgnoreResult();` inside `ezEditorProcessor` , the `-project` string is escaped with `"`.
I'll move it into `ezCommandLineUtils::GetStringOption`.

Code/ThirdParty/AngelScript/CMakeLists.txt
```diff
@@ -106,7 +106,7 @@ if(${ARCHFLAGS} MATCHES "^arm")
     endif()
 endif()
 
-if(${ARCHFLAGS} MATCHES "^aarch64")
+if(EZ_CMAKE_ARCHITECTURE_ARM AND EZ_CMAKE_ARCHITECTURE_64BIT)
     message(STATUS "AngelScript using AARCH64 ASM")
 
     enable_language(ASM)
```
jankrassnigg, line 112: Regarding the build not running on Android:

It's possible that we have to enable a different assembler on ARM:
https://developer.android.com/ndk/guides/cmake#yasm-cmake

So maybe try
enable_language(ASM_NASM)

jankrassnigg: Also note, the original AS CMakeLists.txt is here:
https://svn.code.sf.net/p/angelscript/code/trunk/sdk/angelscript/projects/cmake/CMakeLists.txt

I don't know whether anything else in there is necessary for building on Linux / Android. For example there is this, but I have no idea what it does:

```
# Fix x64 issues on Linux
if("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "x86_64" AND UNIX AND NOT APPLE)
	target_compile_options(${ANGELSCRIPT_LIBRARY_NAME} PRIVATE -fPIC)
endif()
```

Sanakan8472: Ok so for some reason, CMake requires the ASM language support to be enabled globally not just for one project or it forgets it during generate, even if it is only required for a single project.

jankrassnigg (approved):

PR #1510, opened 2025-03-14
Wait for layers to be loaded before export

### LINKED ISSUE #1507: [BUG] Automatic scene transform fails with layers
When you run ezEditorProcessor.exe on a scene that uses layers, the data from the non-main layers will be missing from the exported file.

1. Go to "Data\Samples\Testing Chambers\AssetCache\Common\Scenes" and delete "Ai.ezBinScene"
2. Run ezEditorProcessor.exe with "-project "C:\ezBla\Data\Samples\Testing Chambers" -transform Default"
3. The newly generated Ai.ezBinScene will be ~7KB when it should be 45KB.

It looks like this is happening:

* In ezAssetCurator::ProcessAsset() the scene is not yet open, so we call ezQtEditorApp::GetSingleton()->OpenDocument()
* bWasOpen == false
* then we call TransformAsset() which does a remote export
* However, the EngineStatus (m_EngineStatus) at that time is still "Initializing" it hasn't reached "Loaded" yet.
* ~~My assumption is, that it takes another IPC channel update to send the layer data to the engine, so althrough the data from the main layer is already there, the scene isn't completly transferred to the engine process yet.~~
* ezAssetDocument::RemoteExport() actually checks for `GetEngineStatus() == ezAssetDocument::EngineStatus::Initializing` and asserts that data gets loaded. So that theory is wrong.
* However, when I insert into `ezSceneDocument::RequestExportScene` a call to `SendGameWorldToEngine();` right before `ezAssetDocument::RemoteExport()` all the data is there. Though that would unnecessarily increase export times.
* Another theory is, that `ezAssetDocument::RemoteExport()` waits for `ezDocumentOpenResponseMsgToEditor` to continue with the export. However, only AFTER that message arrives, will `ezAssetDocument::HandleEngineMessage` actually do the mirroring. So it's possible that we insert the `ezExportDocumentMsgToEngine` into the stream before all the object creation stuff is sent.
* `ezProcessCommunicationChannel::MessageFunc` actually first executes `m_WaitForMessageCallback` and only broadcasts the message afterwards, but if I swap that logic, it doesn't fix the problem (and maybe creates different ones?)

Sanakan8472:
* Overloaded `ezScene2Document::InternalTransformAsset` to wait for layers and scene to be fully loaded before exporting. Fixes #1507.
* Don't timeout in `ezProcessCommunicationChannel::WaitForMessage` if a debugger is attached.

### DIFF
--- a/Code/Editor/EditorFramework/Assets/Implementation/AssetDocument.cpp
+++ b/Code/Editor/EditorFramework/Assets/Implementation/AssetDocument.cpp
@@ -699,9 +699,7 @@ ezStatus ezAssetDocument::RemoteExport(const ezAssetFileHeader& header, const ch
 
   ezLog::Info("Exporting {0} to \"{1}\"", GetDocumentTypeName(), szOutputTarget);
 
-  ezStatus loadResult = WaitForEngineStatusLoaded();
-  if (loadResult.Failed())
-    return loadResult;
+  EZ_SUCCEED_OR_RETURN(WaitForEngineStatusLoaded());
 
   range.BeginNextStep(szOutputTarget);
 
--- a/Code/Editor/EditorFramework/Assets/Implementation/AssetDocument.cpp
+++ b/Code/Editor/EditorFramework/Assets/Implementation/AssetDocument.cpp
@@ -699,18 +699,9 @@ ezStatus ezAssetDocument::RemoteExport(const ezAssetFileHeader& header, const ch
 
   ezLog::Info("Exporting {0} to \"{1}\"", GetDocumentTypeName(), szOutputTarget);
 
-  if (GetEngineStatus() == ezAssetDocument::EngineStatus::Disconnected)
-  {
-    return ezStatus(ezFmt("Exporting {0} to \"{1}\" failed, engine not started or crashed.", GetDocumentTypeName(), szOutputTarget));
-  }
-  else if (GetEngineStatus() == ezAssetDocument::EngineStatus::Initializing)
-  {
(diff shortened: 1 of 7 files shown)


Code/Editor/EditorFramework/Assets/Implementation/AssetDocument.cpp
```diff
@@ -699,18 +699,9 @@ ezStatus ezAssetDocument::RemoteExport(const ezAssetFileHeader& header, const ch
...
-    return ezStatus(ezFmt("Exporting {0} to \"{1}\" failed, engine not started or crashed.", GetDocumentTypeName(), szOutputTarget));
-  }
-  else if (GetEngineStatus() == ezAssetDocument::EngineStatus::Initializing)
-  {
-    if (ezEditorEngineProcessConnection::GetSingleton()->WaitForDocumentMessage(GetGuid(), ezDocumentOpenResponseMsgToEditor::GetStaticRTTI(), ezTime::MakeFromSeconds(10)).Failed())
-    {
-      return ezStatus(ezFmt("Exporting {0} to \"{1}\" failed, document initialization timed out.", GetDocumentTypeName(), szOutputTarget));
-    }
-    EZ_ASSERT_DEV(GetEngineStatus() == ezAssetDocument::EngineStatus::Loaded, "After receiving ezDocumentOpenResponseMsgToEditor, the document should be in loaded state.");
-  }
```
jankrassnigg, line 713: This can also just be EZ_SUCCEED_OR_RETURN

jankrassnigg (approved):

Code/Editor/EditorFramework/Assets/Implementation/AssetDocument.cpp
Sanakan8472: Oh good point! Fixed in 3 places.

PR #1515, opened 2025-03-30
CreateResource is no longer run under the resource manager lock

Sanakan8472:
This allows for the resource's `CreateResource` function to load other resources without creating a deadlock. As the resource creation calls `ezResourceManager::GetOrCreateResource` and `ezResourceManager::CreateResource` are no longer executed in their entirely under the resource manager lock, special care needs to be taken that other threads that acquire the lock in the meantime don't interfere with the resource creation. The most important part is that `ezResourceManager::LoadResourceByType` ignores resources that have the `ezResourceFlags::IsCreatedResource` flag set.

# Misc changes
* Fixed various comments in the `ResourceManager.h`.
* Expose `GetCapacity()` on the `ezIdTable`.
* Fix `ezArrayPtr` lldb custom visualizer.
* Created a base class to `ezRegisterSimpleTestHelper` called `ezRegisterTestHelper`which allow it to be derived from to write custom simple tests like simple renderer tests etc. without needing to rewrite the registration infrastructure.

### DIFF
--- a/Code/Engine/Core/ResourceManager/Implementation/ResourceLoading.cpp
+++ b/Code/Engine/Core/ResourceManager/Implementation/ResourceLoading.cpp
@@ -16,8 +16,8 @@ void ezResourceManager::InternalPreloadResource(ezResource* pResource, bool bHig
   if (s_pState->m_bShutdown)
     return;
 
-  // Runtime created resources are not loaded via tasks but created on the stack directly.
-  if (pResource->GetBaseResourceFlags().IsSet(ezResourceFlags::IsCreatedResource))
+  // Runtime created resources without loaders are not loaded via tasks but created on the stack directly.
+  if (pResource->GetBaseResourceFlags().IsSet(ezResourceFlags::IsCreatedResource) && !pResource->GetBaseResourceFlags().IsSet(ezResourceFlags::HasCustomDataLoader))
     return;
 
   EZ_LOCK(s_ResourceMutex);
--- a/Code/Engine/Core/ResourceManager/Implementation/ResourceLoading.cpp
+++ b/Code/Engine/Core/ResourceManager/Implementation/ResourceLoading.cpp
@@ -27,6 +27,10 @@ void ezResourceManager::InternalPreloadResource(ezResource* pResource, bool bHig
     return;
   }
 
+  // Runtime created resources are not loaded via tasks but created on the stack directly.
+  if (pResource->GetBaseResourceFlags().IsSet(ezResourceFlags::IsCreatedResource))
+    return;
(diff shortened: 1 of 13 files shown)


Code/Engine/Core/ResourceManager/Implementation/ResourceLoading.cpp
```diff
@@ -27,6 +27,10 @@ void ezResourceManager::InternalPreloadResource(ezResource* pResource, bool bHig
     return;
   }
 
+  // Runtime created resources are not loaded via tasks but created on the stack directly.
+  if (pResource->GetBaseResourceFlags().IsSet(ezResourceFlags::IsCreatedResource))
+    return;
+
```
jankrassnigg: You should be able to move this up, before the lock, because this flag depends entirely on the resource, not the resource manager, so thread-safety shouldn't be a concern here.

Code/Engine/Core/ResourceManager/Implementation/ResourceManager_inl.h
```diff
@@ -53,48 +53,64 @@ ezTypedResourceHandle<ResourceType> ezResourceManager::GetExistingResource(ezStr
...
   static_assert(std::is_rvalue_reference<DescriptorType&&>::value, "Please std::move the descriptor into this function");
 
+  ezTypedResourceHandle<ResourceType> hResource;
+  ResourceType* pResource = nullptr;
   EZ_LOG_BLOCK("ezResourceManager::CreateResource", sResourceID);
+  {
+    EZ_LOCK(s_ResourceMutex);
+    // In this locked scope, we decide whether we are creating the resource or waiting for it to be created by someone else.
+    if (bAllowGetFallback)
+      hResource = GetExistingResource<ResourceType>(sResourceID);
```
jankrassnigg, line 72: I don't understand the purpose of this fallback resource here, why and how is this used for creating a resource?

Sanakan8472: So I had to combine the code from `ezResourceManager::GetOrCreateResource` and `ezResourceManager::CreateResource` as they effectively do the same except for this little change which is this bool: `bAllowGetFallback`. If set, it first tries to `get` the resource, i.e. fall back to an already existing resource.

jankrassnigg (approved):

Code/Engine/Core/ResourceManager/Implementation/ResourceLoading.cpp
Sanakan8472: Done!

PR #1516, opened 2025-03-31
Material Refactor [BREAKING CHANGE]

Sanakan8472:
## Material Shader Constants - BREAKING CHANGE
* A new material shader section has been added called `[MATERIALCONSTANTS]`. This replaces the `CONSTANT_BUFFER(ezMaterialConstants, 1)` code that was previously done to define material constants.
* To migrate to the new system, a few steps must be taken:
  * All constants that were in the body of the `ezMaterialConstants` buffer, need to be inlined into the `[MATERIALCONSTANTS]` section of the `.ezShader` file. All constants must use the macros defined in `ConstantBufferMacros.h`, e.g. `FLOAT4`, `BOOL1` etc. Note, it is fine to not have a `[MATERIALCONSTANTS]` section if a material does not expose any constants.
  * The material constants must be the same for all permutations of a shader, i.e. any `#ifdefs` inside the constants sections must be removed. This is necessary to allow materials to be stored in an array.
  * Remove any instance of `#include <Shaders/Materials/DefaultMaterialCB.h>`. Instead, inline its constants into your shader's `[MATERIALCONSTANTS]` section.
  * Accessing constants must be done now via the new `GetMaterialData(ConstantName)` macro.
  * Re-transform all visual shader materials.
* Old material shaders will fail to compile if an `ezMaterialConstants` buffer is detected, prompting the user to migrate to the new system.
* Also, runtime inheritance of materials has been removed as it caused major performance overhead and complicated material management. At transform time, material assets are now flattened into material resources without inheritance. Old materials that use inheritance can still be used by the engine, but will then be flattened at load time.

## Shader Changes
* The `[MATERIALCONSTANTS]` is necessary to allow creating storage for a loaded material independent of the permutation that is used. To define the storage, the layout of the material constants must be well defined and parsable inside the `ezShaderResource`.
* `ezShaderResource` will parse the `[MATERIALCONSTANTS]` sections and create a layout that can be retrieved via the new `ezShaderResource::GetMaterialLayout` function.
* The shader compiler will do the same parsing and fail compilation if a `materialData` buffer is found which has a different layout. Compilation will also fail if a `sMaterialConstants` buffer is found, prompting the user to migrate to the new system.
* Both `ezShaderResource` and shader compiler use the `ezShaderParser::ParseMaterialConstantsSection` function to extract the constants from the `.ezShader` file. As the layout of constants is platform dependent, afterwards `ezShaderParser::LayoutMaterialConstants` has to be called with the expected layout format. This is retrieved from either `ezShaderProgramCompiler::GetMaterialBufferLayout` or via `ezGALDevice::GetDefaultDevice()->GetCapabilities().m_MaterialBufferLayout`.
* The shader compiler needs to generate HLSL code out of the abstract material constants section. This is done by using the template file `MaterialConstants.template` in which the constants are injected. The resulting code is prepended before every shader code before compiling, similar to the `[SHADER]` section.
* The DXC shader compiler has been extended to support extracting layouts from structured buffers on top of the current constant buffer support. However, the DX11 compiler does not support this, so we are continuing to render using constant buffers on this platform.

## Material Resource
* The `ezMaterialResource` class no longer holds any GPU resources. This was a problem before as material updates could happen at any time, even in the middle of a render frame. Now, the `ezMaterialResource` only manages the CPU state and changes are extracted in the extraction phase.
* Whenever a user calls one of the material functions that modify its state, a new dirty flag is set to define what was changed, e.g. `Texture2D`, `PermutationVar`, etc.
* After loading / creation of a material, it contains an `ezMaterialResource::ezMaterialId` that can be retrieved via `ezMaterialResource::GetMaterialId`. This will later be the index of the material in the structured buffer shared by all materials using the same shader. For now, the value can be ignored as every material gets its own buffer until high level renderer refactor takes place.
  * Should the `ezMaterialId` ever change on a material due to reloading with a different shader, the `ezMaterialManager::s_MaterialShaderChangedEvent` event is fired to inform anyone caching the id that it has to be updated for the next render frame.

## Material Manager
* The new `ezMaterialManager` singleton is now responsible for the GPU side of materials. At the end of the extraction phase, it runs `ezMaterialManager::ExtractMaterialUpdates` and extracts dirty materials and at the start of the next render frame runs `ezMaterialManager::ApplyMaterialChanges` to apply the extracted changed to the GPU representation.
* For each `ezShaderResource` the material manager creates a `MaterialShaderConstants` instance that manages all materials that share this shader.
* The materials can be stored in individual constant buffers / structured buffers or one singular structured buffer in preparation for bindless rendering, depending on platform support.
  * DX11 uses individual constant buffers as the DX11 shader compiler can't reflect structured buffers.
  * Vulkan uses individual structured buffers as the high level renderer needs to refactored first to have a place to put the material index.
* Each material calls `ezMaterialManager::MaterialAdded` during `UpdateContent` or `CreateResource`. This can be fired multiple times for a single material due to resource reload. On each call we check whether the material shader has changed and if so re-register the material at the new `MaterialShaderConstants` instance and fire the `ezMaterialManager::s_MaterialShaderChangedEvent` event to inform users that any cached material indices must be invalidated for the next render frame.
* `ezMaterialManager::MaterialAdded` is called by the material in the destructor and `ezMaterialManager::MaterialModified` whenever its dirty flag changes.

## Rendering
* As materials are managed by the material manager, the `ezRenderContext` no longer needs to lazily update material constant buffers on first use.
* For rendering, calling `ezMaterialManager::GetMaterialData` returns a `MaterialData` struct, containing all rendering parameters of a material. Materials that aren't loaded yet will return a nullptr here.
  * **BUG**: Due to extraction, there is a one frame delay before material resources that are technically loaded can be rendered as `ezMaterialManager::GetMaterialData` still returns nullptr until the next frame.

## Misc
* Changing one of the visual shader ddl files now works while the editor is open.
* The visual shader code generator no longer injects the VSE constants into all shader stage sections, instead they are just injected into the new [MATERIALCONSTANTS] section.
* Added support for simple renderer tests using the `EZ_CREATE_SIMPLE_RENDERER_TEST_GROUP` and `EZ_CREATE_SIMPLE_RENDERER_TEST` macros. This is equivalent to the `EZ_CREATE_SIMPLE_TEST` pattern, but also initializes a renderer.
* Removed `ReflectanceValue` from materials as it was not used anywhere.
* Added unit tests for various material buffer layouts.


Code/Engine/RendererFoundation/Shader/ShaderByteCode.h
```diff
@@ -42,7 +42,7 @@ struct EZ_RENDERERFOUNDATION_DLL ezShaderConstant
 
   static ezUInt32 s_TypeSize[Type::ENUM_COUNT];
 
-  void CopyDataFormVariant(ezUInt8* pDest, ezVariant* pValue) const;
+  void CopyDataFormVariant(ezUInt8* pDest, const ezVariant* pValue) const;
```
C-Core: While your at it, can you fix the typo in CopyDataFromVariant ?

Code/Engine/RendererCore/Material/MaterialResource.h
```diff
@@ -108,52 +117,55 @@ class EZ_RENDERERCORE_DLL ezMaterialResource final : public ezResource
...
+    };
+
+    struct Bits
+    {
+      StorageType Parameter : 1;
+      StorageType Texture2D : 1;
+      StorageType TextureCube : 1;
+      StorageType PermutationVar : 1;
+      StorageType ShaderAndId : 1;
+      StorageType FlattenHierarchy : 1;
```
C-Core: Doesn't match the enum above

Code/Engine/RendererCore/Material/Implementation/MaterialManager.cpp
```diff
@@ -0,0 +1,686 @@
...
+  {
+    ezResourceLock<ezMaterialResource> pMaterial(extractedMaterial.m_hMaterial, ezResourceAcquireMode::PointerOnly);
+    auto matIt = m_Materials.FindOrAdd(pMaterial.GetPointer());
+    MaterialData& md = matIt.Value();
+    for (ezMaterialResource::DirtyFlags::Enum flag : extractedMaterial.m_DirtyFlags)
+    {
+      switch (flag)
+      {
+        case ezMaterialResource::DirtyFlags::Parameter:
+          md.m_Parameters = extractedMaterial.m_Parameters;
```
C-Core: Could use move here instead of copy since the extracted material will be dead anyways afterwards, right?

Sanakan8472: Not trivially, because these are using the frame allocator but you are right, these can be moved if I fix the allocator to use for these.

Code/Engine/RendererFoundation/Shader/ShaderByteCode.h
Sanakan8472: Fixed!

Code/Engine/RendererCore/Material/MaterialResource.h
Sanakan8472: Oh thanks for noticing!

Code/Engine/RendererCore/Material/Implementation/MaterialManager.cpp
Sanakan8472: Fixed!

Code/Engine/RendererCore/Material/Implementation/MaterialManager.cpp
```diff
@@ -0,0 +1,694 @@
...
+  extractedMaterial.m_hMaterial = pMaterial->GetResourceHandle();
+  extractedMaterial.m_hShader = pMaterial->m_mDesc.m_hShader;
+  extractedMaterial.m_MaterialId = pMaterial->m_MaterialId;
+  extractedMaterial.m_DirtyFlags = pMaterial->m_DirtyFlags;
+  for (ezMaterialResource::DirtyFlags::Enum flag : pMaterial->m_DirtyFlags)
+  {
+    switch (flag)
+    {
+      case ezMaterialResource::DirtyFlags::Parameter:
+        extractedMaterial.m_Parameters = std::move(pMaterial->m_mDesc.m_Parameters);
```
clkern, line 231: Are you sure about these moves here? The material desc is "dead" afterwards. I think we should copy here but move in ApplyMaterialChanges?

C-Core: Are you sure about these moves here? The material desc is "dead" afterwards. I think we should copy here but move in ApplyMaterialChanges?

C-Core (approved):

Code/Engine/RendererCore/Material/Implementation/MaterialManager.cpp
Sanakan8472: Woops! Thanks for noticing! Fixed and updated the renderer test to detect this.

PR #1517, opened 2025-04-02
MiniAudio: Added support for sound groups

jankrassnigg:
* Added suppport for high level sound groups (to adjust effects, music, etc)
* Improved how sounds get instantiated


Code/EnginePlugins/FmodPlugin/Components/FmodEventComponent.cpp
```diff
@@ -185,6 +185,15 @@ void ezFmodEventComponentManager::UpdateOcclusion(const ezWorldModule::UpdateCon
 
 void ezFmodEventComponentManager::UpdateEvents(const ezWorldModule::UpdateContext& context)
 {
+  const ezTime tNow = ezTime::Now();
```
C-Core: It would be better to distribute the work across multiple frames than limit the framerate which will result in frame time spikes.

jankrassnigg: fixed

clkern (approved):

PR #1526, opened 2025-04-12
Visual Script Compiler Refactoring

C-Core:
This is a large scale refactor of the visual script compiler which had a few major design issues and was way to fragile to extend further:
* The new compiler ensures correct temporary variable assignment by doing a proper topological traversal of the graph
* It will be now much easier to expose compiler errors in the editor graph in the future
* All game object input pins are now auto assigned with the script owner game object
* Added support for out and inout function parameter (reference and pointer)

Other changes:
* Added a flag to the script component to run the update function even if the world simulation is not running
* Added a new flag to game objects to control shape icon visibility independently of the "was created by prefab" flag. Also added a flag to the prefab reference component to show the shape icons of objects inside the prefab.
* New export message that can be used in scripts or custom components to do stuff on export
* Other smaller fixes and cleanups


Code/Engine/GameEngine/Messages/ExportMessage.h
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+#include <Foundation/Communication/Message.h>
+#include <GameEngine/GameEngineDLL.h>
+
+struct EZ_GAMEENGINE_DLL ezMsgExport : public ezMessage
```
jankrassnigg, line 8: Please document what the message is used for.

jankrassnigg (approved):

PR #1528, opened 2025-04-13
Add box and capsule shape queries

AlexandrKedrov:
We discussed this topic in Discord channel. Opening PR in case these changes might be helpful.

Some additional considerations:
- Fixed "BoxExtends" -> "BoxExtents"
- `transform` parameter in `SweepTestBox` and `QueryShapesInBox` is ambiguous. Implementation uses only position and rotation parts. Perhaps it's better to split transform into two params.

### DIFF
--- a/Code/EnginePlugins/JoltPlugin/System/JoltQueries.cpp
+++ b/Code/EnginePlugins/JoltPlugin/System/JoltQueries.cpp
@@ -191,7 +191,7 @@ bool ezJoltWorldModule::SweepTestSphere(ezPhysicsCastResult& out_result, float f
 
   const JPH::SphereShape shape(fSphereRadius);
 
-  return SweepTest(out_result, shape, JPH::Mat44::sTranslation(ezJoltConversionUtils::ToVec3(vStart)), vDir, fDistance, params, collection);
+  return SweepTest(out_result, shape, JPH::Vec3(1., 1., 1.), JPH::Mat44::sTranslation(ezJoltConversionUtils::ToVec3(vStart)), vDir, fDistance, params, collection);
 }
 
 bool ezJoltWorldModule::SweepTestBox(ezPhysicsCastResult& out_result, ezVec3 vBoxExtents, const ezTransform& transform, const ezVec3& vDir, float fDistance, const ezPhysicsQueryParameters& params, ezPhysicsHitCollection collection) const
@@ -199,8 +199,9 @@ bool ezJoltWorldModule::SweepTestBox(ezPhysicsCastResult& out_result, ezVec3 vBo
   const JPH::BoxShape shape(ezJoltConversionUtils::ToVec3(vBoxExtents * 0.5f));
 
   const JPH::Mat44 trans = JPH::Mat44::sRotationTranslation(ezJoltConversionUtils::ToQuat(transform.m_qRotation), ezJoltConversionUtils::ToVec3(transform.m_vPosition));
+  const JPH::Vec3 scale = ezJoltConversionUtils::ToVec3(transform.m_vScale);
 
-  return SweepTest(out_result, shape, trans, vDir, fDistance, params, collection);
+  return SweepTest(out_result, shape, scale, trans, vDir, fDistance, params, collection);
 }
 
 bool ezJoltWorldModule::SweepTestCapsule(ezPhysicsCastResult& out_result, float fCapsuleRadius, float fCapsuleHeight, const ezTransform& transform, const ezVec3& vDir, float fDistance, const ezPhysicsQueryParameters& params, ezPhysicsHitCollection collection) const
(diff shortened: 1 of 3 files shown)


Code/EnginePlugins/JoltPlugin/System/JoltQueries.cpp
```diff
@@ -327,14 +327,56 @@ void ezJoltWorldModule::QueryShapesInSphere(ezPhysicsOverlapResultArray& out_res
...
+}
+
+void ezJoltWorldModule::QueryShapesInBox(ezPhysicsOverlapResultArray& out_results, const ezVec3& vBoxExtents, const ezTransform& transform, const ezPhysicsQueryParameters& params) const
+{
+  out_results.m_Results.Clear();
+
+  if (vBoxExtents.x <= 0.0f || vBoxExtents.y <= 0.0f || vBoxExtents.z <= 0.0f)
+    return;
+
+  const JPH::BoxShape shape(ezJoltConversionUtils::ToVec3(vBoxExtents * 0.5f));
```
jankrassnigg: you can simply multiply the transform scale into the box size here

Code/EnginePlugins/JoltPlugin/System/JoltQueries.cpp
```diff
@@ -327,14 +327,56 @@ void ezJoltWorldModule::QueryShapesInSphere(ezPhysicsOverlapResultArray& out_res
...
+}
+
+void ezJoltWorldModule::QueryShapesInCapsule(ezPhysicsOverlapResultArray& out_results, float fCapsuleRadius, float fCapsuleHeight, const ezTransform& transform, const ezPhysicsQueryParameters& params) const
+{
+  out_results.m_Results.Clear();
+
+  if (fCapsuleRadius <= 0.0f)
+    return;
+
+  const JPH::CapsuleShape shape(fCapsuleHeight * 0.5f, fCapsuleRadius);
```
jankrassnigg: you can multiply transform.m_vScale.z into the capsule height, and ezMath::Max(m_vScale.x, m_vScale.y) into the radius

jankrassnigg: you should probably pass ezMath::Max(0.0f, fCapsuleHeight * 0.5f) here, to ignore negative values

jankrassnigg (approved):
Very nice, thanks a lot!

Code/EnginePlugins/JoltPlugin/System/JoltWorldModule.h
```diff
@@ -74,6 +74,10 @@ class EZ_JOLTPLUGIN_DLL ezJoltWorldModule : public ezPhysicsWorldModuleInterface
 
   virtual void QueryShapesInSphere(ezPhysicsOverlapResultArray& out_results, float fSphereRadius, const ezVec3& vPosition, const ezPhysicsQueryParameters& params) const override;
 
+  virtual void QueryShapesInBox(ezPhysicsOverlapResultArray& out_results, const ezVec3& vBoxExtents, const ezTransform& transform, const ezPhysicsQueryParameters& params) const;
+
+  virtual void QueryShapesInCapsule(ezPhysicsOverlapResultArray& out_results, float fCapsuleRadius, float fCapsuleHeight, const ezTransform& transform, const ezPhysicsQueryParameters& params) const;
```
jankrassnigg: these need to be "override"

Code/EnginePlugins/JoltPlugin/System/JoltQueries.cpp
```diff
@@ -327,14 +327,56 @@ void ezJoltWorldModule::QueryShapesInSphere(ezPhysicsOverlapResultArray& out_res
     return;
 
   const JPH::SphereShape shape(fSphereRadius);
```
jankrassnigg, line 397: we should probably also multiply the radius here by ezMath::Max(vScale.x, vScale.y, vScale.z) to be consistent with the other two functions

FabricioMSantos:
Can you also add cylinder? 
To be used on wheels and barrels

Code/EnginePlugins/JoltPlugin/System/JoltQueries.cpp
AlexandrKedrov: Maybe we should delegate scaling to Jolt? There is a shapeScale param, which seems to handle all cases well.
I have tested it for negative and zero scaling. Otherwise, in our implementation, there should be lots of `if condtions`.

Plus, I suppose, Jolt can handle non-uniform transform of spheres and capsules. Thus it should be possible to query ellipsoid-like objects.

AlexandrKedrov: I have pushed the scaling through Jolt version. If you don't like it I will apply changes you proposed.

AlexandrKedrov:
Please, let me run a few more tests before merging. And perhaps I can try adding support for cylinders as suggested by @FabricioMSantos. Likely, I will let you know tomorrow - today I am on a business trip

AlexandrKedrov:
I ran tests and found out I was wrong about non-uniforms scaling in Jolt. So @jankrassnigg I backed to the changes you proposed.

PR #1545, opened 2025-04-25
Force Loading of Highest Quality Levels during Tests

Sanakan8472:
Decal tests are failing spuriously due to some mipmaps not being loaded yet.
* Added `EnsureResourceCondition` which is a generalized version of `EnsureResourceLoadingState`.
* If `GetForceNoFallbackAcquisition` is > 0, wait for higher quality levels to load during resource acquisition. Should be the same logic as for the first quality level just with the added check `pResource->GetNumQualityLevelsLoadable() == 0`.
* `m_uiForceNoFallbackAcquisition` is now actually counted down during the frame tick. Previously, this state persisted indefinitely.
* Random change: Fixed starting the ezPlayer under Linux and the alternative player executable browser dialog.

### DIFF
--- a/Code/Engine/Core/ResourceManager/Implementation/ResourceManager.cpp
+++ b/Code/Engine/Core/ResourceManager/Implementation/ResourceManager.cpp
@@ -500,6 +500,11 @@ void ezResourceManager::PerFrameUpdate()
   {
     FreeUnusedResources(s_pState->m_AutoFreeUnusedTimeout, s_pState->m_AutoFreeUnusedThreshold);
   }
+
+  if (s_pState->m_uiForceNoFallbackAcquisition > 0)
+  {
+    s_pState->m_uiForceNoFallbackAcquisition--;
+  }
 }
 
 const ezEvent<const ezResourceEvent&, ezMutex>& ezResourceManager::GetResourceEvents()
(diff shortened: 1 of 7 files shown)


Code/Engine/Core/ResourceManager/Implementation/ResourceManager.cpp
```diff
@@ -500,6 +500,11 @@ void ezResourceManager::PerFrameUpdate()
   {
     FreeUnusedResources(s_pState->m_AutoFreeUnusedTimeout, s_pState->m_AutoFreeUnusedThreshold);
   }
+
+  if (s_pState->m_uiForceNoFallbackAcquisition > 0)
+  {
+    s_pState->m_uiForceNoFallbackAcquisition--;
```
jankrassnigg, line 506: Oops.

jankrassnigg (commented):
I don't see where this fixes the test ?

Code/Engine/Core/ResourceManager/Implementation/ResourceManager_inl.h
```diff
@@ -207,7 +207,16 @@ ResourceType* ezResourceManager::BeginAcquireResource(const ezTypedResourceHandl
       // accessing IsQueuedForLoading without a lock here is save because InternalPreloadResource() will lock and early out if necessary
       // and accidentally skipping InternalPreloadResource() is no problem
       if (IsQueuedForLoading(pResource) == false && pResource->GetNumQualityLevelsLoadable() > 0)
+      {
         InternalPreloadResource(pResource, false);
+
+        if (GetForceNoFallbackAcquisition() > 0)
+        {
+          EnsureResourceCondition(pResource, [=]() -> bool
+            { return ((ezInt32)pResource->GetLoadingState() >= (ezInt32)ezResourceState::Loaded && pResource->GetNumQualityLevelsLoadable() == 0) ||
```
jankrassnigg, line 216: Oh this is the fix ?

Sanakan8472: That's the fix!

jankrassnigg (approved):

PR #1551, opened 2025-05-02
Pipeline State Objects

Sanakan8472:
## Pipeline State Objects
* New GAL resource: `ezGALGraphicsPipelineCreationDescription` creates a `ezGALGraphicsPipeline`.
* New GAL resource: `ezGALComputePipelineCreationDescription` creates a `ezGALComputePipeline`.
* `ezGALCommandEncoder::SetGraphicsPipeline` and `SetComputePipeline` replaces these previous functions: `SetShader`, `SetVertexDeclaration`, `SetPrimitiveTopology`, `SetBlendState`, `SetDepthStencilState`, `SetRasterizerState`. Same is true for the `ezGALCommandEncoderCommonPlatformInterface`.
* Creating a pipeline increases the reference count on all valid handles in the descriptor.
* Added `ezGALPipelineCache` which is just a hashmap from descriptor to handle which holds a reference to each pipeline. These are never freed until shutdown. This is just a stopgap solution until the high level interface changes, as `ezRenderContext` interface remains the same for now.
* `SetBlendState`, `SetDepthStencilState`, `SetRasterizerState` have been moved from `ezGALCommandEncoder` to `ezRenderContext`.
  * `SetBlendState` parameter `blendFactor` has been removed as it was never implemented.
  * `SetBlendState` parameter `uiSampleMask` has been removed as it required created a new PSO when changed. Also unused so far.
  * `SetDepthStencilState` parameter `uiStencilRefValue` has been moved into a separate function `ezGALCommandEncoder::SetStencilReference`.

## Vulkan
* With the now existing pipeline resources, most of the code in `ezResourceCacheVulkan` has moved into the new resource types. The only changes are:
  * `vk::DynamicState::eScissor` is now dependent on the `ezGALRasterizerStateCreationDescription::m_bScissorTest` flag.
  * `vk::DynamicState::eStencilReference` was added and is dependent on `ezGALDepthStencilStateCreationDescription::m_bStencilTest`.
* `ezResourceCacheVulkan` now maintains a `vk::PipelineCache` on disk. This follows the best practices outlined here: https://zeux.io/2019/07/17/serializing-pipeline-cache/.
* The `ezGALShaderVulkan` now holds a `vk::PipelineLayout` instead of it being generated on the fly in the `ezGALCommandEncoderImplVulkan`.

## Misc
* Guard against shader compiler generating invalid code by appending code snippets that are not separated via new lines.
* Changed posix `GetUserDataFolder` to point to `~/.local/share` instead of just `~/`. Test output data now lands in here instead of cluttering the home directory. Only downside is that test result html files can't be opened by the snap version of Firefox. Use e.g. the dep version from the official webpage instead.
* Changed posix `GetUserDocumentsFolder` to point to `~/Documents` instead of just `~/`.
* Changed a few more tests to not grab the mouse cursor.
* Changed default value of `ezGALResourceFormat` to `Invalid` instead of `RGBAUByteNormalizedsRGB`.

### DIFF
--- a/Code/Engine/RendererDX11/CommandEncoder/CommandEncoderImplDX11.h
+++ b/Code/Engine/RendererDX11/CommandEncoder/CommandEncoderImplDX11.h
@@ -30,8 +30,6 @@ public:
   // ezGALCommandEncoderCommonPlatformInterface
   // State setting functions
 
-  void SetShaderPlatform(const ezGALShader* pShader);
-
   virtual void SetConstantBufferPlatform(const ezShaderResourceBinding& binding, const ezGALBuffer* pBuffer) override;
   virtual void SetSamplerStatePlatform(const ezShaderResourceBinding& binding, const ezGALSamplerState* pSamplerState) override;
   virtual void SetResourceViewPlatform(const ezShaderResourceBinding& binding, const ezGALTextureResourceView* pResourceView) override;
@@ -116,18 +114,18 @@ public:
   virtual void SetGraphicsPipelinePlatform(const ezGALGraphicsPipeline* pGraphicsPipeline) override;
   virtual void SetComputePipelinePlatform(const ezGALComputePipeline* pComputePipeline) override;
 
-  void SetVertexDeclarationPlatform(const ezGALVertexDeclaration* pVertexDeclaration);
-  void SetPrimitiveTopologyPlatform(ezGALPrimitiveTopology::Enum topology);
-  void SetBlendStatePlatform(const ezGALBlendState* pBlendState, const ezColor& blendFactor = ezColor::White, ezUInt32 uiSampleMask = 0xFFFFFFFFu);
-  void SetDepthStencilStatePlatform(const ezGALDepthStencilState* pDepthStencilState);
-  void SetRasterizerStatePlatform(const ezGALRasterizerState* pRasterizerState);
-
   virtual void SetViewportPlatform(const ezRectFloat& rect, float fMinDepth, float fMaxDepth) override;
   virtual void SetScissorRectPlatform(const ezRectU32& rect) override;
   virtual void SetStencilReferencePlatform(ezUInt8 uiStencilRefValue) override;
 
 private:
   friend class ezGALDeviceDX11;
+  void SetShader(const ezGALShader* pShader);
+  void SetVertexDeclaration(const ezGALVertexDeclaration* pVertexDeclaration);
+  void SetPrimitiveTopology(ezGALPrimitiveTopology::Enum topology);
+  void SetBlendState(const ezGALBlendState* pBlendState, const ezColor& blendFactor = ezColor::White, ezUInt32 uiSampleMask = 0xFFFFFFFFu);
+  void SetDepthStencilState(const ezGALDepthStencilState* pDepthStencilState);
+  void SetRasterizerState(const ezGALRasterizerState* pRasterizerState);
 
   bool UnsetResourceViews(const ezGALResourceBase* pResource);
   bool UnsetUnorderedAccessViews(const ezGALResourceBase* pResource);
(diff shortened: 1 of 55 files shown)


Code/Engine/RendererDX11/CommandEncoder/CommandEncoderImplDX11.h
```diff
@@ -112,15 +112,19 @@ class EZ_RENDERERDX11_DLL ezGALCommandEncoderImplDX11 final : public ezGALComman
...
-  virtual void SetVertexDeclarationPlatform(const ezGALVertexDeclaration* pVertexDeclaration) override;
-  virtual void SetPrimitiveTopologyPlatform(ezGALPrimitiveTopology::Enum topology) override;
 
-  virtual void SetBlendStatePlatform(const ezGALBlendState* pBlendState, const ezColor& blendFactor, ezUInt32 uiSampleMask) override;
-  virtual void SetDepthStencilStatePlatform(const ezGALDepthStencilState* pDepthStencilState, ezUInt8 uiStencilRefValue) override;
-  virtual void SetRasterizerStatePlatform(const ezGALRasterizerState* pRasterizerState) override;
+  virtual void SetGraphicsPipelinePlatform(const ezGALGraphicsPipeline* pGraphicsPipeline) override;
+  virtual void SetComputePipelinePlatform(const ezGALComputePipeline* pComputePipeline) override;
+
+  void SetVertexDeclarationPlatform(const ezGALVertexDeclaration* pVertexDeclaration);
```
C-Core: Can we make these functions private now?

Sanakan8472: Of course! I'll also remove the Platform at the end so it's clear it's no longer part of the interface.

Code/Engine/RendererFoundation/State/Implementation/PipelineCache_inl.h
```diff
@@ -0,0 +1,42 @@
...
+
+template <typename HandleType, typename DescType, typename KeyType>
+EZ_ALWAYS_INLINE ezResult ezGALPipelineCache::TryInsertPipeline(const DescType& description, HandleType hNewPipeline, ezHashTable<KeyType, HandleType, ezGALPipelineCache::CacheKeyHasher>& table)
+{
+  KeyType key;
+  key.m_Desc = description;
+  key.m_uiHash = description.CalculateHash();
+
+  EZ_LOCK(m_Mutex);
+  HandleType* pExistingPipeline = table.GetValue(key);
```
C-Core: Would be better to just call Insert and then compare the value returned by out_pOldValue with hNewPipeline.

C-Core (approved):

Code/Engine/RendererVulkan/Cache/Implementation/ResourceCacheVulkan.cpp
```diff
@@ -431,54 +322,112 @@ vk::DescriptorSetLayout ezResourceCacheVulkan::RequestDescriptorSetLayout(const
...
+  prefixHeader.m_uiDataSize = static_cast<ezUInt32>(dataSize);
+  prefixHeader.m_uiVendorID = deviceProperties.vendorID;
+  prefixHeader.m_uiDeviceID = deviceProperties.deviceID;
+  prefixHeader.m_uiDriverVersion = deviceProperties.driverVersion;
+  prefixHeader.m_uiDriverABI = sizeof(void*);
+  for (ezUInt32 i = 0; i < VK_UUID_SIZE; ++i)
+  {
+    prefixHeader.m_uiUuid[i] = deviceProperties.pipelineCacheUUID[i];
+  }
+  prefixHeader.m_uiDataHash = ezHashingUtils::MurmurHash64(pipelineCacheData.GetData(), dataSize);
```
C-Core: We typically use xxHash for everything

Sanakan8472: Fixed!

Code/Engine/RendererFoundation/State/Implementation/PipelineCache_inl.h
Sanakan8472: Fixed!

PR #1566, opened 2025-05-17
Various Fixes

Sanakan8472:
* Fix Asset Curator Panel not showing newly added transform failures. 
  * Added unit test for some basic model changes affected by this change.
* Fixed a broken preview shader.
* Fix race conditions in ezQtAssetBrowserModel. As ezFileSystemModel events can be fired at any time, event handlers must use weak pointers for the event handler to safely allows for ezQtAssetBrowserModel destruction.
* Fix RmlGui tests in case locale is changed.

### DIFF
--- a/Code/Editor/EditorFramework/Assets/Implementation/AssetBrowserModel.cpp
+++ b/Code/Editor/EditorFramework/Assets/Implementation/AssetBrowserModel.cpp
@@ -108,8 +108,19 @@ ezQtAssetBrowserModel::ezQtAssetBrowserModel(QObject* pParent, ezQtAssetFilter* 
   : QAbstractItemModel(pParent)
   , m_pFilter(pFilter)
 {
-  EZ_ASSERT_DEBUG(pFilter != nullptr, "ezQtAssetBrowserModel requires a valid filter.");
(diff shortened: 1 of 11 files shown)


Code/Editor/EditorFramework/Assets/Implementation/AssetBrowserModel.cpp
```diff
@@ -770,6 +797,8 @@ void ezQtAssetBrowserModel::FileSystemFileEventHandler(const ezFileChangedEvent&
 
   if (bFire)
   {
+    // QTimer::singleShot(500, [this]()
```
jankrassnigg: Remove dead code ?

jankrassnigg (approved):

Code/Editor/EditorFramework/Assets/Implementation/AssetBrowserModel.cpp
Sanakan8472: Woopsi! Thanks for noticing. Is fixed now. I'll update clang tidy in a later commit to detect Qt pointers as pointer types.

PR #1579, opened 2025-05-30
Cookie textures and light materials for spot lights

C-Core:
BREAKING: VSE materials need to be re-saved in the editor otherwise they will fail to transform.

* Added support for cookie texture and light materials for spot lights:
image
image

* This is based on a new dynamic texture atlas that is now also used for the shadow pool.
* Light materials can be created in VSE, they are called decal materials there. The regular decals can be extended to make use of them as well in the future.
image


Data/Base/Materials/Common/DecalCopy.ezMaterialAsset
jankrassnigg, line 1: this file doesn't seem to be used anywhere, what is it for ?

jankrassnigg (commented):
The "Material" and "Cookie" options can be set at the same time, but then only the Material seems to be used. Would it be possible to allow both at the same time? Don't know whether there is a use for having the material for the animation, and the cookie as a static mask?

Also a bit more obvious example would be nice, had to search the history to even find where this might be used.

Data/Base/Materials/Common/DecalCopy.ezMaterialAsset
C-Core: It is used to copy a simple texture (e.g. the cookie texture) to the decal atlas.

C-Core:
Material and Cookie are exclusive. Since the material can be completely user generated it is not clear what a "mask" should do. I don't think there is a good way to combine them both. Also it would complicate the code even further, which is already complicated enough.

Agreed a dedicated sample for all kind of light types and their various features would be nice. Maybe later, I've spend so much time with this feature already.

PR #1583, opened 2025-06-01
SRV / UAV Refactor

Sanakan8472:
The old model was based on DX11 API and did not support buffers that could be viewed as multiple out of Structured-, Texel- and ByteAddressBuffer. As the old model counted elements instead of bytes, the view was ambiguous as soon as multiple kinds of elements could be present in a buffer. To fix this, the following changes have been made: 
## ezGALBufferResourceViewCreationDescription, ezGALBufferUnorderedAccessViewCreationDescription
* Removed `m_uiFirstElement` and `m_uiNumElements`, replacing them with `m_uiByteOffset` and `m_uiByteCount`. If `m_uiByteCount` is 0, it means the rest of the buffer.
* Removed `m_bRawView` and replaced it with `ezEnum<ezGALShaderResourceType> m_ResourceType`. This unambiguously defines what kind of view this is and what fields need to be set.
  * `m_ResourceType` can be `StructuredBuffer`, `TexelBuffer` or `ByteAddressBuffer` for SRVs and the same with `RW` at the end for UAVs. The only exception is for views on buffers which only support one of these flags in their `ezGALBufferUsageFlags` in which case `m_ResourceType` can be left as `Unknown`.
  * Added `ezGALShaderResourceType::ByteAddressBuffer` and `ezGALShaderResourceType::ByteAddressBufferRW` and extended shader compilers to recognize these in reflection. Note that the old Ids have been preserved to not have to patch old serialized data.
* `m_Format` is only valid if `m_ResourceType` is `TexelBuffer(RW)`. Setting it for any other view type result in an error.

## ezGALBufferCreationDescription
* `m_uiStructSize` is only required if `m_BufferFlags` contains `VertexBuffer`, `IndexBuffer` or `StructuredBuffer`, but can be set as a hint for other buffer types as well, which is usef for e.g. `ezGALDynamicBuffer` as an element size hint.
* `m_Format` is required if `m_BufferFlags` contains `TexelBuffer` and will result in an error if not set or if set on a non-TexelBuffer.

## ezGALDevice
* Added lots of validation to view creation. Note that `CreateResourceView` now takes the description by value in order to modify it to make platform code simpler:
  * If `m_ResourceType` is `Unknown` and not ambiguous, the correct type will be patched in.
  * If `m_uiByteCount` is 0, it will be replaced with the correct count to reach the end of the buffer.
* Extracted duplicated view creation / destruction code into templates.
* `GetDefaultResourceView` has an optional parameter: `ezEnum<ezGALShaderResourceType> slotType = ezGALShaderResourceType::Unknown`. If the buffer is not ambiguous, the optional parameter can be ignored. Otherwise it must be set or the call fails. Note that we now create a default resource view for every supported view type by the buffer of `StructuredBuffer`, `TexelBuffer` or `ByteAddressBuffer`.
* Added `m_bSupportsMultipleSRVTypes` to `ezGALDeviceCapabilities`. Defines whether more than of of `ezGALBufferUsageFlags::StructuredBuffer`, `ezGALBufferUsageFlags::TexelBuffer` and `ezGALBufferUsageFlags::ByteAddressBuffer` is supported on a single buffer.

## Misc
* Added `ezGraphicsTest::GetDeviceCapabilities` which caches the caps so that not every test init has to create a temporary GAL device to query these.
* Added `TexelBuffer` and `ByteAddressBuffer` tests. Thanks to the interface changes, the test code can be shared with the `StructuredBuffer` as only the declaration of the buffer and the used shader has to change. The test was also extended with UAV write tests for each type of buffer.


Code/Engine/RendererFoundation/Descriptors/Descriptors.h
```diff
@@ -239,10 +239,10 @@ struct ezGALTextureResourceViewCreationDescription : public ezHashableStruct<ezG
 struct ezGALBufferResourceViewCreationDescription : public ezHashableStruct<ezGALBufferResourceViewCreationDescription>
 {
   ezGALBufferHandle m_hBuffer;
-  ezUInt32 m_uiFirstElement = 0;
-  ezUInt32 m_uiNumElements = 0;
-  ezEnum<ezGALResourceFormat> m_Format = ezGALResourceFormat::Invalid;
-  bool m_bRawView = false;
+  ezUInt32 m_uiByteOffset = 0;                    ///< Start of the view to the buffer. Must be multiple of the element size.
+  ezUInt32 m_uiByteCount = 0;                     ///< If 0, this maps to the rest of the buffer.
```
C-Core: Vulkan has this VK_WHOLE_SIZE which maps to uint(-1) aka ezInvalidIndex, I think this would make more sense than 0 as magic value.

Code/Engine/RendererFoundation/Device/DeviceCapabilities.h
```diff
@@ -71,6 +71,7 @@ struct EZ_RENDERERFOUNDATION_DLL ezGALDeviceCapabilities
   bool m_bSupportsConservativeRasterization = false;
   bool m_bSupportsVSRenderTargetArrayIndex = false;
   bool m_bSupportsTexelBuffer = false; ///< Whether ezGALBufferUsageFlags::TexelBuffer is supported. Hardcoded per platform as it must match SUPPORTS_TEXEL_BUFFER shader define.
+  bool m_bSupportsMultipleSRVTypes = true; ///< Whether more than of of ezGALBufferUsageFlags::StructuredBuffer, ezGALBufferUsageFlags::TexelBuffer and ezGALBufferUsageFlags::ByteAddressBuffer is supported on a buffer.
```
C-Core: typo in comment "of of" should be "one of" I guess?

C-Core (approved):

Code/Engine/RendererFoundation/Descriptors/Descriptors.h
Sanakan8472: Added `EZ_GAL_WHOLE_SIZE`.

Code/Engine/RendererFoundation/Device/DeviceCapabilities.h
Sanakan8472: Thanks!

PR #1584, opened 2025-06-15
New Linux CI Machine

Sanakan8472:
# Linux CI Machine changes
* Moved Linux build to dedicated machine.
* Moved Android-arm64 to the same machine.
* Linux is now testing Editor and GameEngine tests as well.

# Linux Fixes
* Prevent creating a file reader on a directory.
* Various fixes on Posix ezProcess implementation to allow calling various APIs more than once.
* Stack tracer now compiles without unity builds.

# Renderer Fixes
* Shared texture Linux fixes for correct handle lifetime.
* Basics renderer test was using the mising mesh texture which has no mipmaps and caused lots of image compare failures. Now replaced by FullbrightWhite material.
* LoadScene test output looks much different on all platforms now than the reference image to new images have been checked in.
* Added alternative Vulkan reference images for dynamic texture atlas tests due to line rendering differences. 
* Workaround for random NaN splotches on AMD Radeon 680M in SSAO shader.

# Misc Fixes
* Editor Processor no longer opens an editor window on failure to compile.
* Valgrind fixes for access violations in string functions.
* Merged EditorProcessorTest into GenerateCompile test to support Linux as that test almost does the same prerequisites.


Code/Engine/Foundation/Platform/Posix/OSFile_Posix.inl
```diff
@@ -59,6 +59,21 @@ ezResult ezOSFile::InternalOpen(ezStringView sFile, ezFileOpenMode::Enum OpenMod
     return EZ_FAILURE;
   }
 
+  struct stat stats = {};
+  if (fstat(fd, &stats) != 0)
+  {
+    ezLog::Error("Failed to stat file {}, error {}", szFile, ezArgErrno(errno));
```
jankrassnigg: Such low level functions shouldn't write to the log.

Code/UnitTests/GameEngineTest/Basics/Basics.cpp
```diff
@@ -354,7 +354,7 @@ void ezGameEngineTestApplication_Basics::SubTestManyMeshesSetup()
   m_pWorld->Clear();
 
   ezMeshResourceHandle hMesh = ezResourceManager::LoadResource<ezMeshResource>("Meshes/MissingMesh.ezBinMesh");
-
+  ezMaterialResourceHandle hMaterial = ezResourceManager::LoadResource<ezMaterialResource>("{ b93a6e80-5bf7-48f0-b7b3-2097d0133159 }"); // FullbrightWhite
```
jankrassnigg, line 357: Maybe a texture with some structure would be better than just one color?

Sanakan8472: I think this is fine, the purpose of the test is to check whether rendering many meshes is correctly sliced into batches so we only are interested in the presence or absence of a box which a white pixel is sufficient for. There are already many tests that render textures so adding them here would only massively increase the reference image size.

Code/Engine/Foundation/Platform/Posix/OSFile_Posix.inl
Sanakan8472: Good point, I wasn't sure whether I should log here. There are a few log messages in this file and I'm not sure these should be in here either.

jankrassnigg: Probably not. Most file system functions can fail under normal circumstances. They should just return an error and leave the logging to higher level code.

jankrassnigg (approved):

Code/Engine/Foundation/Platform/Posix/OSFile_Posix.inl
Sanakan8472: Ok, I'll remove them then. There also aren't any equivalent log messages in the Windows implementation.

PR #1589, opened 2025-06-25
Resource View Refactor

Sanakan8472:
## Bind Groups

* Removed all `BindX` methods from the `ezRenderContext`. Instead, call `ezRenderContext::GetBindGroup` which returns a `ezBindGroupBuilder` and call the `BindX` methods there. Note that that you can pass any of the constants `EZ_GAL_BIND_GROUP_FRAME`, `EZ_GAL_BIND_GROUP_RENDER_PASS`, `EZ_GAL_BIND_GROUP_MATERIAL`, `EZ_GAL_BIND_GROUP_DRAW_CALL` into the `GetBindGroup` call, but for now all resources reside in the global bind group so best to ignore until further refactorings have taken place.
* `ezBindGroupBuilder` now handles all resource binding calls. SRV / UAV as resources have been removed from the code base and are now implicitly created when necessary by the renderer implementation. You now bind resources directly to the builder. Any view relevant data like `ezEnum<ezGALTextureType>` will be automatically derived once the resource is bound. View extends are defined via these parameters on the bind calls `ezGALBufferRange` and `ezGALTextureRange`. See migration guide for further details.
* `ezRenderContext` will call `ezBindGroupBuilder::CreateBindGroup` if the builder was modified before submitting a draw call. This creates a `ezGALBindGroupCreationDescription` which is then passed into `ezGALCommandEncoder::SetBindGroup`.
* `ezGALBindGroupCreationDescription` consists of the target `ezGALBindGroupLayoutHandle` and a list of `ezGALBindGroupItem` which map one to one to the array of `ezShaderResourceBinding` in the layout.
* `ezGALCommandEncoder::SetBindGroup` replaces all `SetX` functions that accepted a `ezShaderResourceBinding` and a resource before. This function will do all the validation that previously view creation did but only in debug builds.
* Any view creation happens in the backend if neccessary in the Vulkan / DX11 specific buffer / texture classes.

## BREAKING CHANGES / Migration

Remappings from old `ezRenderContext` functions to new `ezBindGroupBuilder` functions:
* `BindSamplerState` -> `BindSampler`
* `BindBuffer`, `BindConstantBuffer`, `BindUAV(buffer)` -> `BindBuffer`
* `BindTexture2D`, `BindTexture3D`, `BindTextureCube` `BindUAV(texture)` -> `BindTexture`
* `ezGALDevice::GetDefaultResourceView` -> remove entirely. Replaced by default function parameters for `ezGALTextureRange` and `ezGALBufferRange`.

```C++
// OLD
ezGALTextureUnorderedAccessViewHandle hFilterOutput;
{
  ezGALTextureUnorderedAccessViewCreationDescription desc;
  desc.m_hTexture = pFilteredSpecularOutput->m_TextureHandle;
  desc.m_uiMipLevelToUse = uiMipMapIndex;
  desc.m_uiFirstArraySlice = m_uiSpecularOutputIndex * 6;
  desc.m_uiArraySize = 6;
  desc.m_OverrideViewType = ezGALTextureType::Texture2DArray;
  hFilterOutput = pDevice->CreateUnorderedAccessView(desc);
}
renderViewContext.m_pRenderContext->BindUAV("ReflectionOutput", hFilterOutput);

// NEW
ezBindGroupBuilder& bindGroup = renderViewContext.m_pRenderContext->GetBindGroup();

ezGALTextureRange textureRange;
textureRange.m_uiBaseMipLevel = uiMipMapIndex;
textureRange.m_uiBaseArraySlice = m_uiSpecularOutputIndex * 6;
textureRange.m_uiArraySlices = 6;
bindGroup.BindTexture("ReflectionOutput", pFilteredSpecularOutput->m_TextureHandle, textureRange);
// Or as a shorter form:
bindGroup.BindTexture("ReflectionOutput", pFilteredSpecularOutput->m_TextureHandle, {m_uiSpecularOutputIndex * 6, 6, uiMipMapIndex, EZ_GAL_ALL_MIP_LEVELS});
```

## Renderer Changes

* Removed all `ClearUnorderedAccessView` functions from the renderer. If needed write a custom compute shader.
* For DX11, `D3D_FEATURE_LEVEL_11_1` is now the minimum feature level as the engine was already no longer running on 11.0 due to reliance on the Overwrite buffer upload flag. Hence, the `m_bSupportsNoOverwriteBufferUpdate` capability has been removed.
* Proxy textures now store a handle to their parent texture and their target slice. This is needed to replace all proxy handles with the actual textures during `ezBindGroupBuilder::CreateBindGroup` which makes renderer implementations easier.

## Misc

* Fixed ComputeShaderHistogram to no longer rely on `CopyTexture` with incompatible formats or `ClearUnorderedAccessView`.
* Smaller fixes to allow DXVK to compile.


Code/Engine/Foundation/Platform/Linux/FormatString_Linux.cpp
```diff
@@ -14,4 +14,10 @@ ezStringView BuildString(char* szTmp, ezUInt32 uiLength, const ezArgErrno& arg)
   ezStringUtils::snprintf(szTmp, uiLength, "%i (\"%s\")", arg.m_iErrno, szErrorMsg);
   return ezStringView(szTmp);
 }
+
+ezStringView BuildString(char* szTmp, ezUInt32 uiLength, const ezArgErrorCode& arg)
+{
+  ezStringUtils::snprintf(szTmp, uiLength, "%i", arg.m_ErrorCode);
```
jankrassnigg: since m_ErrorCode is a uint32, this has to be "%u"

Code/Engine/RendererCore/RenderContext/BindGroupBuilder.h
```diff
@@ -0,0 +1,66 @@
...
+{
+public:
+  ezBindGroupBuilder();
+  /// \brief Returns whether a Bind* call modified in internal state since the last CreateBindGroup call.
+  bool IsModified() const { return m_bModified; }
+
+  /// Binds a sampler to this bind group
+  /// @param sSlotName The slot under which the sampler is to be bound.
+  /// @param hSampler If valid, it will be bound. If not, bind group item under this slot will be removed and replaced with a fallback resource if required.
+  void BindSampler(const ezTempHashedString& sSlotName, ezGALSamplerStateHandle hSampler);
```
jankrassnigg: ezTempHashedString is just a uint32, so can be passed by value

Code/Engine/RendererCore/RenderContext/Implementation/RenderContext.cpp
```diff
@@ -475,7 +277,8 @@ void ezRenderContext::SetPushConstants(const ezTempHashedString& sSlotName, ezAr
     {
       ezArrayPtr<ezUInt8> targetStorage = pStorage->GetRawDataForWriting();
       ezMemoryUtils::Copy(targetStorage.GetPtr(), data.GetPtr(), data.GetCount());
-      BindConstantBuffer(sSlotName, m_hPushConstantsStorage);
+      // #TODO: which bind group?
```
jankrassnigg: todo

Code/Engine/RendererFoundation/Descriptors/Implementation/Enumerations_inl.h
```diff
@@ -41,3 +41,40 @@ inline bool ezGALShaderTextureType::IsArray(ezGALShaderTextureType::Enum format)
...
+    case ezGALShaderTextureType::Texture2DArray:
+    case ezGALShaderTextureType::Texture2DMSArray:
+      return ezGALTextureType::Texture2DArray;
+    case ezGALShaderTextureType::Texture3D:
+      return ezGALTextureType::Texture3D;
+    case ezGALShaderTextureType::TextureCube:
+      return ezGALTextureType::TextureCube;
+    case ezGALShaderTextureType::TextureCubeArray:
+      return ezGALTextureType::TextureCubeArray;
+    default:
```
jankrassnigg: default should be the last case

Code/Engine/RendererFoundation/RendererFoundationDLL.h
```diff
@@ -425,13 +424,32 @@ struct ezGALTextureSubresource
...
 
+
 struct ezGALSystemMemoryDescription
 {
   ezConstByteBlobPtr m_pData;
   ezUInt32 m_uiRowPitch = 0;
   ezUInt32 m_uiSlicePitch = 0;
 };
 
+struct ezGALTextureRange
```
jankrassnigg, line 437: doc comment for what the struct is

Code/Engine/RendererFoundation/RendererFoundationDLL.h
```diff
@@ -425,13 +424,32 @@ struct ezGALTextureSubresource
...
+  {
+    return {0, 1, uiBaseMipLevel, uiMipLevels};
+  }
+  ezUInt16 m_uiBaseArraySlice = 0;                    ///< Index of the first array slice to be used.
+  ezUInt16 m_uiArraySlices = EZ_GAL_ALL_ARRAY_SLICES; ///< Number of array slices to be used. If set to EZ_GAL_ALL_ARRAY_SLICES, the maximum number of allowed slices is used dependent on texture size and binding contraints.
+  ezUInt8 m_uiBaseMipLevel = 0;                       ///< The first mip level to be used.
+  ezUInt8 m_uiMipLevels = EZ_GAL_ALL_MIP_LEVELS;      ///< Number of mip levels to be used. Ignored for UAVs. If set to EZ_GAL_ALL_MIP_LEVELS, the maximum number of allowed mip maps is used dependent on texture size.
+};
+
+struct ezGALBufferRange
```
jankrassnigg, line 452: struct doc comment

Code/Engine/RendererFoundation/RendererFoundationDLL.h
```diff
@@ -604,6 +594,12 @@ class ezGALPipelineLayoutHandle
   friend class ezGALDevice;
 };
 
+class ezGALBindGroupHandle
```
jankrassnigg, line 605: class doc comment

Code/Engine/RendererFoundation/Shader/Implementation/BindGroup_inl.h
```diff
@@ -0,0 +1,22 @@
+
+ezGALBindGroupItem::ezGALBindGroupItem()
+{
+  m_Dummy[0] = {};
```
jankrassnigg: what's the dummy for and why is only the first element initialized ?

Code/Engine/RendererFoundation/Shader/Implementation/BindGroup_inl.h
```diff
@@ -0,0 +1,22 @@
...
+}
+ezGALBindGroupItem::ezGALBindGroupItem(const ezGALBindGroupItem& rhs)
+{
+  m_Flags = rhs.m_Flags;
+  m_Dummy[0] = rhs.m_Dummy[0];
+  m_Dummy[1] = rhs.m_Dummy[1];
+  m_Dummy[2] = rhs.m_Dummy[2];
+  m_Dummy[3] = rhs.m_Dummy[3];
```
jankrassnigg: can be implement with *this = rhs

Data/Samples/ComputeShaderHistogram/Shaders/histogramclear.ezShader
jankrassnigg, line 1: Filename is a bit ugly with the all lower case name. Same for "clearParams.h" and "screenCopy.ezShader", which is not consistent with our typical file naming

Data/Samples/ComputeShaderHistogram/Shaders/screenCopy.ezShader
jankrassnigg, line 1: can this be a built-in feature, at some point?
having to write a dedicated shader for this feels wrong

Code/Engine/RendererVulkan/CommandEncoder/Implementation/CommandEncoderImplVulkan.cpp
```diff
@@ -1325,64 +1137,108 @@ ezResult ezGALCommandEncoderImplVulkan::FlushDeferredStateChanges()
...
-  }
-  return pUAV;
-}
+    const ezShaderResourceBinding& binding = bindings[i];
+    const ezGALBindGroupItem& item = m_BindGroups[uiBindGroup].m_Desc.m_BindGroupItems[i];
+
+    vk::WriteDescriptorSet& write = m_DescriptorWrites.ExpandAndGetRef();
+    write.dstArrayElement = 0;
+    write.descriptorType = ezConversionUtilsVulkan::GetDescriptorType(binding.m_ResourceType);
+    write.dstBinding = binding.m_iSlot; // #TODO_VULKAN this should be i + arrayIndex or something?
```
jankrassnigg: todo

Code/Engine/RendererFoundation/CommandEncoder/Implementation/CommandEncoder.cpp
```diff
@@ -6,127 +9,172 @@
...
-        EZ_ASSERT_DEBUG((binding.m_TextureType == ezGALShaderTextureType::Texture2DArray && !bMSAA) || (binding.m_TextureType == ezGALShaderTextureType::Texture2DMSArray && bMSAA), "Mismatch between shader resource and bound view.");
+        case ezGALShaderResourceType::Texture:
+        case ezGALShaderResourceType::TextureRW:
+        case ezGALShaderResourceType::TextureAndSampler:
+        {
+          EZ_ASSERT_ALWAYS(item.m_Flags.IsSet(ezGALBindGroupItemFlags::Texture), "Item type does not match binding");
+          const ezGALTexture* pTexture = m_Device.GetTexture(item.m_Texture.m_hTexture);
+          EZ_ASSERT_ALWAYS(pTexture != nullptr, "Invalid texture");
+          const auto& textureDesc = pTexture->GetDescription();
+          // TODO item.m_Texture.m_OverrideViewFormat
```
jankrassnigg: todo

Code/Engine/RendererFoundation/CommandEncoder/Implementation/CommandEncoder.cpp
jankrassnigg, line 1: should all those asserts in this file really be EZ_ASSERT_ALWAYS ?

Code/Engine/RendererCore/RenderContext/Implementation/BindGroupBuilder.cpp
```diff
@@ -0,0 +1,316 @@
...
+#include <RendererFoundation/Resources/ProxyTexture.h>
+#include <RendererFoundation/Resources/RendererFallbackResources.h>
+#include <RendererFoundation/Shader/BindGroupLayout.h>
+
+ezUInt32 ezBindGroupBuilder::s_uiWrites = 0;
+ezUInt32 ezBindGroupBuilder::s_uiReads = 0;
+
+ezBindGroupBuilder::ezBindGroupBuilder()
+{
+  // Platforms that do not support immutable samples like DX11 still need them to be bound manually, so they are bound here.
```
jankrassnigg: samplers, not samples

Code/Engine/RendererCore/RenderContext/Implementation/BindGroupBuilder.cpp
```diff
@@ -0,0 +1,316 @@
...
+  }
+}
+
+ezResult ezBindGroupBuilder::CreateBindGroup(const ezGALDevice* pDevice, ezGALBindGroupLayoutHandle hBindGroupLayout, ezGALBindGroupCreationDescription& out_BindGroup)
+{
+  EZ_LOG_BLOCK("CreateBindGroup");
+  const ezGALBindGroupLayout* pLayout = pDevice->GetBindGroupLayout(hBindGroupLayout);
+  if (pLayout == nullptr)
+  {
+    ezLog::Error("Bind group layout is null.");
```
jankrassnigg: maybe this should just be an assert ?
generally I wonder whether this function's error handling is the right thing, it both logs and returns a status code
but then, is this even allowed to fail at runtime, or would that always be a programmer's error ?

Code/Engine/RendererCore/RenderContext/Implementation/BindGroupBuilder.cpp
```diff
@@ -0,0 +1,316 @@
...
+          if (ezGALBufferHandle hBuffer = ezGALRendererFallbackResources::GetFallbackBuffer(binding.m_ResourceType); !hBuffer.IsInvalidated())
+          {
+            item.m_Flags = ezGALBindGroupItemFlags::Buffer | ezGALBindGroupItemFlags::Fallback;
+            item.m_Buffer.m_hBuffer = hBuffer;
+            item.m_Buffer.m_BufferRange = {};
+            item.m_Buffer.m_OverrideViewFormat = {};
+          }
+          else
+          {
+            ezLog::Error("No buffer bound for binding '{0}'", binding.m_sName);
```
jankrassnigg: error or assert ?

Code/Engine/RendererCore/RenderContext/BindGroupBuilder.h
```diff
@@ -0,0 +1,66 @@
...
+
+  /// Binds a buffer to this bind group
+  /// @param sSlotName The slot under which the buffer is to be bound.
+  /// @param hBuffer If valid, it will be bound. If not, bind group item under this slot will be removed and replaced with a fallback resource if required.
+  /// @param bufferRange What part of the buffer should be bound. Default is entire buffer.
+  /// @param overrideViewFormat Texel Buffers only! Sets the format of the texel buffer. If invalid, default format of the buffer will be used.
```
jankrassnigg: could name the parameter "texelBufferOverrideFormat" then

Sanakan8472: Good idea, also to differentiate it from the equally named `overrideViewFormat` on textures which can apply to all textures.

Code/Engine/RendererCore/RenderContext/Implementation/BindGroupBuilder.cpp
Sanakan8472: So I've been struggling a bit with error handling here. 
Previously, error handling wa done during UAV / SRV creation on every build type. This was done up front but now everything is deferred a bit.

I think I'll move all resource dependent error handling from the `CreateBindGroup` to the respective `BindX` calls as these should also be called less often. This would then assert at the right place, when you try to bind the resource. This would be mainly dead handle and out of bounds checks. 

`CreateBindGroup` will be called a lot so it should be as cheap as possible so I'll make it not do any validation except for asserting on nullptrs.
Due to the use of fallback resources, the function really can't fail unless we forgot to add a fallback resource to the `ezGALRendererFallbackResources` class.

The `ezGALBindGroupItem` to `ezShaderResourceBinding` validation happens in `ezGALCommandEncoder::SetBindGroup`. This will check whether the bound resouce is compatible with the binding slot. If this fails, this will always be a user error (I think?), not something that can fail at runtime due to e.g. resources not being loaded yet. E.g. a resource is missing a capability like UAV but is bound to a UAV slot or a texture range of 2 slices is bound to a cube map that requires exactly 6 slices. These checks are very expensive so they will only be run in debug builds for now. Not 100% sure whether we should enable it in dev for now until we find some corner cases maybe? I put all the checks in a huge `#if EZ_ENABLED(EZ_COMPILE_FOR_DEBUG)` block so we can quickly enable them for dev as well by changing one line.

Code/Engine/RendererFoundation/CommandEncoder/Implementation/CommandEncoder.cpp
Sanakan8472: So as I wasn't sure whether we want to enable this validation in dev builds as well (technically, these were mostly part of the SRV / UAV creation, so they were previously run on all builds). So I added a big `#if EZ_ENABLED(EZ_COMPILE_FOR_DEBUG)` at the top to be able to quicly flip flop on which builds we want to run these, that's why the individual checks are `EZ_ASSERT_ALWAYS` so only the huge ifdev at the top decides on when to run these.

Code/Engine/RendererVulkan/CommandEncoder/Implementation/CommandEncoderImplVulkan.cpp
Sanakan8472: Oh I think I copied his over from the old code. The current impl is correct. If we used array indices (which we don't support in shaders so we would never get that far) it would be put into the `dstArrayElement` field.

Data/Samples/ComputeShaderHistogram/Shaders/histogramclear.ezShader
Sanakan8472: So whole sample doesn't follow any naming schemas and I was torn between keep it as is or tear everything apart. I can make another pass over it to clean it up.

Code/Engine/RendererFoundation/Shader/Implementation/BindGroup_inl.h
Sanakan8472: Ok so the moment I added the union, all the base class operators and ctors were marked as deleted by some C++ quirk. I think I added the dummy to make clang compile. Really I just want to base class to null out the memory and that's it. Let me see if I can convince it to compile without the dummy. I think I'm forced to init at least one of the union members but setting any one them would make the struct non-zeroed out which I don't want.

Code/Engine/RendererCore/RenderContext/Implementation/RenderContext.cpp
Sanakan8472: Ah yes, that's a problem for future me when I rewrite all shaders to use bind groups. I've changed the comment to make it clearer what we need to do in the future, basically:
"Once all shaders are migrated to use multiple bind groups, we need to decide where to put the fallback buffer for platforms that don't support push constants, e.g. DX11. Right now, everything ends up in slot 0, so this code is correct for now but it is also the worst place to put the push constants as that group should change the least and push constants should change the most."

Code/Engine/Foundation/Platform/Linux/FormatString_Linux.cpp
Sanakan8472: Good catch! This was actually wrong in the windows impl as well. Note I only added these so that I can compile DX11 code on Linux using DXVK.

Code/Engine/RendererCore/RenderContext/Implementation/BindGroupBuilder.cpp
Sanakan8472: Assert it is! See big post on the builder code above.

Code/Engine/RendererFoundation/CommandEncoder/Implementation/CommandEncoder.cpp
Sanakan8472: Oh right! We never validated that before. Add checks now that `GetBitsPerElement` and `GetChannelCount` match between formats.

Data/Samples/ComputeShaderHistogram/Shaders/screenCopy.ezShader
Sanakan8472: Yeah well the CopyTexture call that the sample was using didn't work on vulkan as the formats do not match at all. DX11 does not care though and the sample only ever ran on DX11 before. For now I just wanted to get it to run at all, even if not pretty. Too much already in this PR. I'll have it on my todo list to make the CopyTexture API support any formats by falling back to rendering if it can't be solved with a memcpy on the GPU.

Data/Samples/ComputeShaderHistogram/Shaders/histogramclear.ezShader
jankrassnigg: You don't need to make it perfect, but at least new stuff should follow the typical names.

Sanakan8472: Gave all the sample shaders proper names.

Code/Engine/RendererFoundation/Shader/Implementation/BindGroup_inl.h
Sanakan8472: Seems to build without it now!

jankrassnigg (approved):

Code/Engine/RendererCore/RenderContext/Implementation/BindGroupBuilder.cpp
```diff
@@ -0,0 +1,302 @@
...
+  }
+  m_bModified = false;
+}
+
+void ezBindGroupBuilder::RemoveItem(ezTempHashedString sSlotName, ezHashTable<ezUInt64, ezGALBindGroupItem>& ref_Container)
+{
+  s_uiReads++;
+  if (ref_Container.Contains(sSlotName.GetHash()))
+  {
+    ref_Container.Remove(sSlotName.GetHash());
```
C-Core: Remove should return whether is has actually removed something. The extra contains means an extra hash slot lookup, in case of collisions this can mean linear search etc. Better just call Remove directly.

Code/Engine/RendererCore/RenderContext/Implementation/BindGroupBuilder.cpp
```diff
@@ -0,0 +1,302 @@
...
+    m_bModified = true;
+    s_uiWrites++;
+  }
+}
+
+void ezBindGroupBuilder::InsertItem(ezTempHashedString sSlotName, const ezGALBindGroupItem& item, ezHashTable<ezUInt64, ezGALBindGroupItem>& ref_Container)
+{
+  s_uiReads++;
+  ezGALBindGroupItem* pOldItem = nullptr;
+  if (ref_Container.TryGetValue(sSlotName.GetHash(), pOldItem))
```
C-Core: Insert can return the old value, I would use that and do the comparison afterwards. Same as with Remove, you are currently doing the hash slot lookup twice.

C-Core (approved):

Code/Engine/RendererCore/RenderContext/Implementation/BindGroupBuilder.cpp
Sanakan8472: Oh right! Thanks for pointing that out!

Code/Engine/RendererCore/RenderContext/Implementation/BindGroupBuilder.cpp
Sanakan8472: Thanks! I'm clearly not thinking in hash sets yet.

PR #1590, opened 2025-06-27
Fixed #723: Added ezJoltBreakableSlabComponent

### LINKED ISSUE #723: [FEATURE] Jolt: Migrate and/or rewrite breakable sheet component
The PhysX breakable sheet component has not been migrated to Jolt yet. Could either attempt to migrate its feature set to Jolt or rewrite it.

However, I would probably at least rewrite the way the physics bodies are set up. Potentially use a Jolt ragdoll for the set of bodies (not sure whether that makes sense). 

Another option would be to build something new that is more dynamic, E.g. for breakable glass windows, it would be very cool to dynamically compute the break pattern and maybe have additional state (jointed bodies) to be able to further break it apart. Probably quite a lot of work.

This is not a trivial task, but for someone who wants to do a bit of physics programming with Jolt this is a nice project.

jankrassnigg:
Basically the same as the breakable sheet component that we had a while ago, but with additional features.
There's a new scene "Hall" to try things out.


Code/Engine/Core/Interfaces/PhysicsWorldModule.h
```diff
@@ -148,6 +148,36 @@ struct EZ_CORE_DLL ezMsgReleaseObjectGrab : public ezMessage
...
+/// Only sent for certain physics object combinations, e.g. debris doesn't trigger this.
+/// The reported contact position and normal is an average of the contact manifold.
+/// This is mainly meant for less important interactions, like breaking decorative things.
+struct EZ_CORE_DLL ezMsgPhysicContact : public ezMessage
+{
+  EZ_DECLARE_MESSAGE_TYPE(ezMsgPhysicContact, ezMessage);
+
+  ezVec3 m_vGlobalPosition;
+  ezVec3 m_vNormal;
+  float m_fImpactSqr;
```
Sanakan8472, line 177: Why is the impact here squared but not on the character version?

Code/EnginePlugins/JoltPlugin/Components/JoltBreakableSlabComponent.h
```diff
@@ -0,0 +1,265 @@
...
+namespace JPH
+{
+  class BodyInterface;
+}
+
+/// \brief Flags that define which edges of a breakable slab are fixed/anchored in the world.
+/// A fixed edge means that shards adjacent to that edge will remain stationary and not fall under gravity.
+struct EZ_JOLTPLUGIN_DLL ezJoltBreakableSlabFlags
+{
+  using StorageType = ezUInt16;
```
Sanakan8472: ezUInt8 would suffice?

Code/EnginePlugins/JoltPlugin/Components/Implementation/JoltBreakableSlabComponent.cpp
```diff
@@ -0,0 +1,1348 @@
...
+//////////////////////////////////////////////////////////////////////////
+
+// clang-format off
+EZ_BEGIN_COMPONENT_TYPE(ezJoltBreakableSlabComponent, 1, ezComponentMode::Dynamic)
+{
+  EZ_BEGIN_PROPERTIES
+  {
+    EZ_ACCESSOR_PROPERTY("Width", GetWidth, SetWidth)->AddAttributes(new ezDefaultValueAttribute(1.0f), new ezClampValueAttribute(0.1f, 8.0f), new ezSuffixAttribute(" m")),
+    EZ_ACCESSOR_PROPERTY("Height", GetHeight, SetHeight)->AddAttributes(new ezDefaultValueAttribute(1.0f), new ezClampValueAttribute(0.1f, 8.0f), new ezSuffixAttribute(" m")),
+    EZ_ACCESSOR_PROPERTY("Thickness", GetThickness, SetThickness)->AddAttributes(new ezDefaultValueAttribute(0.02f), new ezClampValueAttribute(0.005f, 1.0f), new ezSuffixAttribute(" m")),
```
Sanakan8472, line 153: m_fThickness default value differs for editor and engine.

Code/EnginePlugins/JoltPlugin/Components/Implementation/JoltBreakableSlabComponent.cpp
```diff
@@ -0,0 +1,1348 @@
...
+{
+  m_fContactReportForceThreshold = 0.0f;
+
+  if (m_ShatterPoints.IsEmpty())
+  {
+    ((ezJoltBreakableSlabComponentManager*)GetOwningManager())->m_RequireBreakage.Insert(GetHandle());
+  }
+
+  auto& pt = m_ShatterPoints.ExpandAndGetRef();
+  pt.m_vGlobalPosition.Set(1024, 1024, 1024);
```
Sanakan8472, line 471: Can we make this a constant like `ezInvalidIndex`? E.g. `ezShatterALl` ?

jankrassnigg: yeah, but this is only internal, and only used in 2 places, so not that important atm

Code/Engine/Core/Interfaces/PhysicsWorldModule.h
jankrassnigg: Because that's what I get from different Jolt code paths. I could return the squared value in both cases, or the non-squared value, but in both cases I would then need to do extra calculations, which may be wasted.
It's not pretty, I know.

PR #1592, opened 2025-06-29
Resource Bind Group Assignments

Sanakan8472:
# Resource Bind Group Assignments

Most resources are now assigned to individual bind groups. Previously, as everything was using the default frame bind group, every `ezBindGroupBuilder` in use was called like this:
```C++
ezBindGroupBuilder& bindGroup = ezRenderContext::GetDefaultInstance()->GetBindGroup();
```
Now, the bind group builders have been renamed as follows:
```C++
// The `EZ_GAL_BIND_GROUP_FRAME` bind group still relies on the default parameter of GetBindGroup. This allows to quickly look at all bindings that have not been migrated yet.
ezBindGroupBuilder& bindGroup = ezRenderContext::GetDefaultInstance()->GetBindGroup();
// Migrated code to the new bind groups:
ezBindGroupBuilder& bindGroupRenderPass = ezRenderContext::GetDefaultInstance()->GetBindGroup(EZ_GAL_BIND_GROUP_RENDER_PASS);
ezBindGroupBuilder& bindGroupMaterial = ezRenderContext::GetDefaultInstance()->GetBindGroup(EZ_GAL_BIND_GROUP_MATERIAL);
ezBindGroupBuilder& bindGroupDraw = ezRenderContext::GetDefaultInstance()->GetBindGroup(EZ_GAL_BIND_GROUP_DRAW_CALL);
// Samples:
ezBindGroupBuilder& bindGroupSample = ezRenderContext::GetDefaultInstance()->GetBindGroup();
// Unit tests:
ezBindGroupBuilder& bindGroupTest = ezRenderContext::GetDefaultInstance()->GetBindGroup();
```
By searching for e.g. `bindGroupRenderPass.Bind` you can quickly look up all bindings to that bind group. There is still some code that has not been touched. Mainly render passes that run very small shaders like copy shaders which still reside in the default `EZ_GAL_BIND_GROUP_FRAME` bind group. These generally don't share any resources with other code so there isn't any gain from moving these (I think?).

## Materials
* Everything from `ezClusteredDataGPU` and `ezSimplifiedDataGPU` resides in `EZ_GAL_BIND_GROUP_FRAME`. E.g. "SkyIrradianceTexture", "perLightDataBuffer" and "ezGlobalConstants".
* The `EZ_GAL_BIND_GROUP_RENDER_PASS` contains render pass specific resources, e.g. "SSAOTexture" for the `ezOpaqueForwardRenderPass` or "SceneColor" for the `ezTransparentForwardRenderPass`.
* The `EZ_GAL_BIND_GROUP_MATERIAL` contains the "materialData" constant / structured buffer as well as all material resources. The shader compiler will force the exposed resources (those in the `[MATERIALPARAMETER]` section) into this group for now for easier migration.
* `EZ_GAL_BIND_GROUP_DRAW_CALL` contains mostly "perInstanceData" and "ezObjectConstants".

## Particles
These are very similiar to materials. The particle data arrays are in the `EZ_GAL_BIND_GROUP_DRAW_CALL` group. `EZ_GAL_BIND_GROUP_MATERIAL` contains "ParticleTexture" and "ezParticleSystemConstants".

## Push Constants
On platforms that don't support push constants, the fallback constant buffer is now located in the `EZ_GAL_BIND_GROUP_DRAW_CALL` bind group.

## Simple Renderers
Smaller renderers like debug renderer or lens flare renderer put all their resources into the `EZ_GAL_BIND_GROUP_RENDER_PASS` bind group to not pollute the frame bind group which these also access.

# Renderer Changes
* `MeshDecalMaterial.ezShader` now works on Vulkan.
* DX11 will now emulate bind groups. This is done by the shader compiler by simply overwriting the bind group index after shader reflection which the values that `ezShaderParser::ParseShaderResources` extracted.
* Cleaned up renderer code by renaming 'sets' into 'bind groups' in older parts of the codebase.


Code/Engine/ShaderCompilerHLSL/ShaderCompilerHLSL.cpp
```diff
@@ -346,6 +346,39 @@ ezSharedPtr<ezShaderConstantBufferLayout> ezShaderCompilerHLSL::ReflectConstantB
   return pLayout;
 }
 
+ezResult ezShaderCompilerHLSL::AddFakeSetAssignments(ezShaderProgramData& inout_Data, ezGALShaderStage::Enum Stage, ezLogInterface* pLog)
```
C-Core: Should be called bind group and not set

C-Core (approved):

Code/Engine/ShaderCompilerHLSL/ShaderCompilerHLSL.cpp
Sanakan8472: Oh right, 'set' really is an inconvenient word to search for. Fixed!

PR #1600, opened 2025-07-10
Sped up BreakableSlab setup in editor

jankrassnigg:
It was recreating meshes many times during scene setup. Now this is done only once and potentially on a separate thread. Also optimized the rectangle shape creation.
Clarified and renamed the component update function description.

### DIFF
--- a/Code/Engine/Core/World/WorldModule.h
+++ b/Code/Engine/Core/World/WorldModule.h
@@ -70,8 +70,8 @@ protected:
                                                   ///< called after all its dependencies have been called.
     ezEnum<ezWorldUpdatePhase> m_Phase;           ///< The update phase in which this update function should be called. See ezWorld for a description on the different phases.
     bool m_bOnlyUpdateWhenSimulating = false;     ///< The update function is only called when the world simulation is enabled.
-    ezUInt16 m_uiAsyncPhaseBatchSize = 0;         ///< 0 means all components are updated one after another, but still in parallel with other component managers.
-                                                  ///< >0 means m_Function will be called multiple times with batches of roughly this size.
+    ezUInt16 m_uiAsyncPhaseBatchSize = 0;         ///< 0 means m_Function is called once per frame, to update all components, but still in parallel with other world modules.
+                                                  ///< >0 means m_Function is called multiple times (in parallel) with batches of roughly this size.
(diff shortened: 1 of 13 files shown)


Code/Engine/Core/World/WorldModule.h
```diff
@@ -70,8 +70,8 @@ class EZ_CORE_DLL ezWorldModule : public ezReflectedClass
                                                   ///< called after all its dependencies have been called.
     ezEnum<ezWorldUpdatePhase> m_Phase;           ///< The update phase in which this update function should be called. See ezWorld for a description on the different phases.
     bool m_bOnlyUpdateWhenSimulating = false;     ///< The update function is only called when the world simulation is enabled.
-    ezUInt16 m_uiGranularity = 0;                 ///< The granularity in which batch updates should happen during the asynchronous phase. Has to be 0 for
-                                                  ///< synchronous functions.
+    ezUInt16 m_uiAsyncPhaseBatchSize = 0;         ///< 0 means all components are updated one after another, but still in parallel with other component managers.
```
C-Core: should say "in parallel with other world modules", they don't need to be component managers

Code/Engine/GameEngine/AI/Implementation/SensorComponent.cpp
```diff
@@ -621,18 +621,19 @@ void ezSensorWorldModule::Initialize()
   SUPER::Initialize();
 
   {
-    auto updateDesc = EZ_CREATE_MODULE_UPDATE_FUNCTION_DESC(ezSensorWorldModule::UpdateSensors, this);
-    updateDesc.m_Phase = ezWorldUpdatePhase::Async;
-    updateDesc.m_bOnlyUpdateWhenSimulating = true;
+    auto desc = EZ_CREATE_MODULE_UPDATE_FUNCTION_DESC(ezSensorWorldModule::UpdateSensors, this);
+    desc.m_Phase = ezWorldUpdatePhase::Async;
+    desc.m_bOnlyUpdateWhenSimulating = true;
+    desc.m_uiAsyncPhaseBatchSize = 16;
```
C-Core: This is wrong. The sensor component uses its own scheduler to distribute sensor update across multiple frames. It must not be called multiple times within one frame.

C-Core (approved):

PR #1606, opened 2025-07-11
Shader Compiler Multi-Treading Support and Renderer Fixes

### LINKED ISSUE #1573: [BUG] Shader compilation errors
Use a Dev build.
Run `ezShaderCompiler.exe -project "abs-ez/Data\Samples\Testing Chambers" -ignoreErrors`

This will compile all shaders. The following errors show up in the log:

```console
Error: Could not compile shader 'AssetCache/Common/Materials/Common/Pattern.autogen.ezShader' for profile 'ps_5_0'
Error: Shaders/Materials/MaterialPixelShader.h(298,32-43): error X3004: undeclared identifier 'gameObjectId'
Shaders/Materials/MaterialPixelShader.h(298,18-44): error X3013: 'RGBA8ToFloat4': no matching 1 parameter function
```

I think the declaration and usage of `gameObjectId` uses different preprocessor conditions.

```console
Error: Could not compile shader 'Shaders/Debug/DebugPrimitive.ezShader' for profile 'vs_5_0'
Error: Shaders/Debug/DebugPrimitive.vs(32,11-24): warning X3578: Output value 'main' is not completely initialized
Shaders/Debug/DebugPrimitive.vs(32,11-24): error X4580: Output variable main contains a system-interpreted value (SV_RenderTargetArrayIndex) which must be written in every execution path of the shader.  Unconditional initialization may help.
```

I assume this is about the stereo pipeline, and we don't always write the index for the render target.
The same happens in `GizmoHandle.ezShader`.

```console
Error: Could not compile shader 'Shaders/RmlUi.ezShader' for profile 'vs_5_0'
Error: Shaders/RmlUi.vs(67,1-6): error X3000: unrecognized identifier 'VS_OUT'
```

Don't know what this is about.

### LINKED ISSUE #1604: [BUG] TextureSample flickers heavily
The TextureSample renders only every second frame, it seems.

Sanakan8472:
# Shader Compiler
* Various static caches have been extended with locks.
* Added a `ezTokenizedFileCache` as a parameter to `ezShaderCompiler::CompileShaderPermutationForPlatforms`. This allows to drastically cut down on parsing cost.
* `ezRTTI::ForEachDerivedType` hold a lock, so the shader compiler was never using more than one thread as all compile calls were under the `ezRTTI` lock. Separated the RTTI discovery loop from the actual shader compile loop so the compilation happens outside the `ezRTTI` lock.
* Fixed shader compiler tool to handle both relative as well as absolute files and directories. The problem was that a relative folder can't be made absolute via `ezFileSystem::ResolvePath` as that relies and file readers which don't apply to folders.
* Fixed all shaders that failed compilation. Fixes #1573.

# Renderer Fixes
* `ezGameApplication::IsGameUpdateEnabled` will now call `ezRenderWorld::IsRenderingScheduled` which in turns checks for active views and submitted render pipelines. The old code was not checking for render pipelines which caused extracted render pipelines to be rendered much later than planned, e.g. when resources they extracted where already destroyed.
* Separated layout change code from the `ezGALCommandEncoderImplVulkan::CreateDescriptorSet`into a new function `EnsureBindGroupTextureLayout`. If a descriptor was cached before, the layout change code would be skipped, causing Vulkan validation issues.
* The texture sample was caching the render target view to a single texture in the swap chain and reusing it every frame, causing flicker. Fixes #1604.
* Fixed a bug that caused VSE materials to not work reliably on skinned meshes. The material output node was exposing `Permutation VERTEX_SKINNING;` as a material property which overwrote the global state of the skinned mesh renderer.

### DIFF
--- a/Data/Base/Shaders/Debug/DebugPrimitive.ezShader
+++ b/Data/Base/Shaders/Debug/DebugPrimitive.ezShader
@@ -31,11 +31,13 @@ DepthTest = true
 
 VS_OUT main(VS_IN Input)
 {
-#if CAMERA_MODE == CAMERA_MODE_STEREO
-  s_ActiveCameraEyeIndex = Input.InstanceID % 2;
-#endif
-
   VS_OUT RetVal;
+  #if defined(CAMERA_MODE) && CAMERA_MODE == CAMERA_MODE_STEREO
+    s_ActiveCameraEyeIndex = Input.InstanceID % 2;
+    #if VERTEX_SHADER_RENDER_TARGET_ARRAY_INDEX == TRUE
+      RetVal.RenderTargetArrayIndex = Input.InstanceID % 2;
+    #endif
+  #endif
 
   #if PRE_TRANSFORMED_VERTICES
     float2 screenPosition = (Input.Position.xy * ViewportSize.zw) * float2(2.0, -2.0) + float2(-1.0, 1.0);
--- a/Data/Samples/Testing Chambers/Materials/MeshDecalMaterial.ezShader
+++ b/Data/Samples/Testing Chambers/Materials/MeshDecalMaterial.ezShader
@@ -10,6 +10,7 @@ SHADING_QUALITY = SHADING_QUALITY_NORMAL
 TWO_SIDED = FALSE
 FLIP_WINDING
 FORWARD_PASS_WRITE_DEPTH
+VERTEX_SHADER_RENDER_TARGET_ARRAY_INDEX
 MSAA
 CAMERA_MODE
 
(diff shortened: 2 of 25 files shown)


Data/Base/Shaders/Debug/DebugPrimitive.ezShader
```diff
@@ -31,11 +31,13 @@ DepthTest = true
 
 VS_OUT main(VS_IN Input)
 {
-#if CAMERA_MODE == CAMERA_MODE_STEREO
-  s_ActiveCameraEyeIndex = Input.InstanceID % 2;
-#endif
-
   VS_OUT RetVal;
+  #if defined(CAMERA_MODE) && CAMERA_MODE == CAMERA_MODE_STEREO
+    s_ActiveCameraEyeIndex = Input.InstanceID % 2;
```
C-Core, line 36: Technically the s_ActiveCameraEyeIndex assignment is not needed since the variable is not used anywhere in this vertex shader but doesn't hurt.

C-Core (approved):

Data/Samples/Testing Chambers/Materials/MeshDecalMaterial.ezShader
```diff
@@ -62,8 +63,17 @@ float GetVertexDepthBias()
 
 VS_OUT main(VS_IN Input)
 {
+#if CAMERA_MODE == CAMERA_MODE_STEREO
```
C-Core: This extra stereo code is not needed since it is all already done inside FillVertexData()

Data/Base/Shaders/Debug/DebugPrimitive.ezShader
Sanakan8472: It's used by `GetWorldToScreenMatrix` below.

Data/Samples/Testing Chambers/Materials/MeshDecalMaterial.ezShader
Sanakan8472: Thanks! Didn't notice that!

PR #1615, opened 2025-07-25
Scripting and other improvements

C-Core:
* Various fixes in the visual script compiler, e.g. in some cases data nodes where executed in the wrong order or in use registers have been overwritten.
* Implicit conversion for float -> vec3 and vec3 -> transform in visual script
* Added PushBackRange function to visual script
* Extended spawn prefab script functions to have more options and also return the created root objects
* Exposed stable random functions to script (ezSimdRandom in C++)
* Fixed how var args were determined from reflection in angel script
* Fixed and extended a few simd vector functions
* Transform properties have a widget now
* Lowered default light penumbra size for sharper shadows and less artifacts with default settings


Code/EnginePlugins/AngelScriptPlugin/TypeRegistration/AsRegister_Reflected.cpp
```diff
@@ -5,17 +5,18 @@
...
 #include <Foundation/Reflection/ReflectionUtils.h>
 
-bool ezAngelScriptEngineSingleton::AppendType(ezStringBuilder& decl, const ezRTTI* pRtti, const ezScriptableFunctionAttribute* pFuncAttr, ezUInt32 uiArg, bool& inout_VarArgs)
+bool ezAngelScriptEngineSingleton::AppendType(ezStringBuilder& decl, const ezRTTI* pRtti, const ezScriptableFunctionAttribute* pFuncAttr, ezUInt32 uiArg)
 {
   const bool bIsReturnValue = uiArg == ezInvalidIndex;
+  const auto argType = pFuncAttr ? pFuncAttr->GetArgumentType(uiArg) : ezScriptableFunctionAttribute::ArgType::In;
 
-  if (pRtti == nullptr)
+  if (pRtti == nullptr || pRtti == ezGetStaticRTTI<ezVariantArray>())
```
jankrassnigg, line 13: If bIsReturnValue == false, this shouldn't append void, but probably do nothing or assert.
I guess we don't have that case, because ezVariantArray is (so far) only used as a last vararg parameter.

Code/EnginePlugins/AngelScriptPlugin/TypeRegistration/AsRegister_Reflected.cpp
```diff
@@ -395,10 +388,16 @@ void ezAngelScriptEngineSingleton::RegisterGenericFunction(const char* szTypeNam
 
 void ezAngelScriptEngineSingleton::RegisterSingleGenericFunction(const char* szFuncName, const char* szTypeName, const ezAbstractFunctionProperty* const pFunc, const ezScriptableFunctionAttribute* pFuncAttr, bool bIsInherited, const ezRTTI* pReturnType)
 {
-  ezStringBuilder decl;
   bool bVarArgs = false;
+  if (const ezDynamicPinAttribute* pVarArgsAttr = pFunc->GetAttributeByType<ezDynamicPinAttribute>())
```
jankrassnigg, line 392: Is this ezDynamicPinAttribute something that we can rely on? Also should we maybe assert, that it is set for vararg params?

Code/EnginePlugins/AngelScriptPlugin/TypeRegistration/AsRegister_Reflected.cpp
```diff
@@ -491,13 +491,8 @@ void ezAngelScriptEngineSingleton::RegisterSingleGenericFunction(const char* szF
...
         if (defaultValue.IsValid())
         {
-          if (pArgType->GetVariantType() == ezVariantType::UInt32)
-          {
-            // special case to make the default value unsigned
-            defaultValue = defaultValue.ConvertTo<ezUInt32>();
-          }
-
-          decl.Append(ezAngelScriptUtils::DefaultValueToString(defaultValue));
+          ezVariantType::Enum expectedType = bIsEnum ? ezVariantType::Int64 : pArgType->GetVariantType();
```
jankrassnigg, line 494: no idea what this change is about

Code/EnginePlugins/AngelScriptPlugin/TypeRegistration/AsRegister_Reflected.cpp
C-Core: The dynamic pin attribute is the only thing that indicates var args, a variant array alone is not enough.

Code/EnginePlugins/AngelScriptPlugin/TypeRegistration/AsRegister_Reflected.cpp
C-Core: The value in ezDefaultValueAttribute may not have the correct type for a function parameter. Thus we pass the type of the function parameter and convert the value to that type. Otherwise you can end up with cases like bool b = 1 in the function declaration und angel script does not like that.

Code/EnginePlugins/AngelScriptPlugin/TypeRegistration/AsRegister_Reflected.cpp
C-Core: Can change that in my next PR

Issue #1624, opened 2025-08-06
[BUG] ezWorldModule::DeregisterUpdateFunction crashes engine when called from Update


AlexandrKedrov:
**Bug description:**
If one of the so-called "Update" functions calls ezWorldModule::DeregisterUpdateFunction, it crashes the engine

**How to Reproduce:**
A dummy example to reproduce
```cpp
// DummyComponent.h
#include <Core/World/ComponentManager.h>
#include <Core/World/Component.h>


class DummyComponentManager : public ezComponentManager<class DummyComponent, ezBlockStorageType::FreeList>
{
  using SUPER = ezComponentManager<class DummyComponent, ezBlockStorageType::FreeList>;

public:
  DummyComponentManager(ezWorld* pWorld);
  void OnSimulationStarted() override;
  void Update(const ezWorldModule::UpdateContext& context);

private:
  ezWorldModule::UpdateFunctionDesc getUpdateDescription();
};

class DummyComponent : public ezComponent
{
   EZ_DECLARE_COMPONENT_TYPE(DummyComponent, ezComponent, DummyComponentManager);
};
```

```cpp
// DummyComponent.cpp
#include "DummyComponent.h"


// clang-format off
EZ_BEGIN_COMPONENT_TYPE(DummyComponent, 1, ezComponentMode::Static)
EZ_END_DYNAMIC_REFLECTED_TYPE;
// clang-format on


DummyComponentManager::DummyComponentManager(ezWorld* pWorld)
  : SUPER(pWorld)
{}

void DummyComponentManager::OnSimulationStarted()
{
  SUPER::OnSimulationStarted();

  auto desc = getUpdateDescription();
  RegisterUpdateFunction(desc);
}

void DummyComponentManager::Update(const ezWorldModule::UpdateContext& context)
{
  auto desc = getUpdateDescription();
  DeregisterUpdateFunction(desc);
}

ezWorldModule::UpdateFunctionDesc DummyComponentManager::getUpdateDescription()
{
  auto desc = EZ_CREATE_MODULE_UPDATE_FUNCTION_DESC(DummyComponentManager::Update, this);
  desc.m_Phase = ezWorldUpdatePhase::PreAsync;
  desc.m_bOnlyUpdateWhenSimulating = true;

  return desc;
}
```

PR #1625, opened 2025-08-06
Trivial fix for #1624

AlexandrKedrov:
Here is a trivial solution for #1624. 

But there is one consideration I would like to highlight: 
Most importantly it doesn't take dependencies (`UpdateFunctionDesc::m_DependsOn`) into account  before deletion. Perhaps a better approach is to mark all the deletion methods for internal usage only.

Current changes:

1. Added `WorldData::m_UpdateFunctionsToDeregister` to store update functions marked for deletion
2. `WorldData::DeregisterUpdateFunction` is renamed to `DeregisterUpdateFunctionInternal`
3. `WorldData::DeregisterUpdateFunctions` is renamed to `DeregisterUpdateFunctionsInternal` for constistency
4. Now `WorldData::DeregisterUpdateFunction` appends functiond description to  `WorldData::m_UpdateFunctionsToDeregister` array
5. All marked for deletion update functions are deleted by `ezWorld::ProcessUpdateFunctionsToDeregister` by the end of `ezWorld::Update()`

### DIFF
--- a/Code/Engine/Core/World/Implementation/World.cpp
+++ b/Code/Engine/Core/World/Implementation/World.cpp
@@ -1348,9 +1348,6 @@ void ezWorld::ProcessUpdateFunctionsToDeregister()
 {
   CheckForWriteAccess();
 
-  if (m_Data.m_UpdateFunctionsToDeregister.IsEmpty())
-    return;
-
   for (const ezWorldModule::UpdateFunctionDesc& updateFunction : m_Data.m_UpdateFunctionsToDeregister)
   {
     DeregisterUpdateFunctionInternal(updateFunction);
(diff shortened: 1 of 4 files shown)


jankrassnigg:
That one failing test often fails, unclear why, but not due to this change.

Code/Engine/Core/World/Implementation/World.cpp
```diff
@@ -548,6 +548,7 @@ void ezWorld::Update()
     EZ_PROFILE_SCOPE("Delete Dead Objects");
     DeleteDeadObjects();
     DeleteDeadComponents();
+    ProcessUpdateFunctionsToDeregister();
```
C-Core: I would maybe put this call directly before the ProcessUpdateFunctionsToRegister call. This place seems a bit odd.

C-Core (approved):

PR #1635, opened 2025-08-21
Added new sample project: Crazy Ball

jankrassnigg:
Sample for different input handling methods and a bit of physics. Fixed the behavior of the MoveTo component.
Changed that global keys will now overwrite an existing reference, to allow easier hand-off from one object to another within a frame.


C-Core:
Looks good, the only thing I would add is some kind of goal visual.

C-Core (approved):

PR #1637, opened 2025-08-31
Better foot normal detection

### LINKED ISSUE #1636: [BUG] Incorrect normal calculation when "jumping on narrow high object"
**Bug description:**
If the character "jump on narrow high object" contact collector will gather normals incorrectly.
```cpp
void ezJoltCharacterControllerComponent::CollectContacts(ezDynamicArray<ContactPoint>& out_Contacts, const JPH::Shape* pShape, const ezVec3& vQueryPosition, const ezQuat& qQueryRotation, float fCollisionTolerance) const
{
  out_Contacts.Clear();

  class ContactCollector : public JPH::CollideShapeCollector
  {
  public:
    ezDynamicArray<ContactPoint>* m_pContacts = nullptr;
    const JPH::BodyLockInterface* m_pLockInterface = nullptr;

    virtual void AddHit(const JPH::CollideShapeResult& result) override
    {
      ...
      // contact.m_vContactNormal equals ezVec3(0, -1, 0), which results in incorrect angle calculation with ezVec3::MakeAxisZ();
      contact.m_vContactNormal = ezJoltConversionUtils::ToVec3(-result.mPenetrationAxis.Normalized());
      ...
    }
    ...
  };
  ...
}
```
**How to Reproduce:**
1. Set up high narrow object (barrel in Testing Chambers might work)
2. Jump on the object without pressing any keys afterwards
3. Try to jump again -> no jump allowed

**Video:**

https://github.com/user-attachments/assets/0bfa9306-8214-4682-8497-74efc15c92c6

AlexandrKedrov:
Resolves #1636

**Main idea**
1. Keep a foot shape collider on the same height as the character.
2. For collision detection use `mMaxSeparationDistance = m_fMaxStepDown` which is described in Jolt as 
```cpp
/// When > 0 contacts in the vicinity of the query shape can be found. All nearest contacts that are not further away than this distance will be found.
/// Note that in this case CollideShapeResult::mPenetrationDepth can become negative to indicate that objects are not overlapping. (unit: meter)"
```
3. Therefore we can detect shapes which are below our character by m_fMaxStepDown. Also we can use mPenetrationDepth to filter out false collisions (outside capsule radius)

**Screenshots**
Before changes: 
image
After changes:
image

**Other changes**:
I have replaced `fCollisionTolerance` with `fMaxSeparationDistance`. The later seems to be more suitable choice for most of the cases + it provides additional info via `mPenetrationDepth`

I haven't found any Jolt's CC features to implement custom foot size. Actually, since #753 most of the stair stepping logic is handled by Jolt. 

Additionally, I would be thankful if someone could run some tests. I will do some more tests on my own, but I don't have any experience developing CC and might miss something important

### DIFF
--- a/Code/EnginePlugins/JoltPlugin/Character/Implementation/JoltCharacterControllerComponent.cpp
+++ b/Code/EnginePlugins/JoltPlugin/Character/Implementation/JoltCharacterControllerComponent.cpp
@@ -248,6 +248,7 @@ void ezJoltCharacterControllerComponent::CollectCastContacts(ezDynamicArray<Cont
       contact.m_vContactNormal = ezJoltConversionUtils::ToVec3(-result.mPenetrationAxis.Normalized());
       contact.m_BodyID = result.mBodyID2;
       contact.m_fCastFraction = result.mFraction;
+      contact.m_fPenetrationDepth = result.mPenetrationDepth;
       contact.m_SubShapeID = result.mSubShapeID2;
 
       JPH::BodyLockRead lock(*m_pLockInterface, contact.m_BodyID);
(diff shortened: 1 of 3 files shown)


Sanakan8472:
Thanks @AlexandrKedrov for your contribution! Unfortunately, I don't have much insight into the physics code so I'll hand over to @jankrassnigg to have a look. In the meantime, I can at least trigger the pipelines.

jankrassnigg:
Tested it a bit, and it seems to work quite well.

jankrassnigg (approved):
Thanks for the contribution! As far as I can tell it even made the stair stepping a bit more robust.

PR #1640, opened 2025-09-02
Change Location of "Add Entry" button

jankrassnigg:
This is an attempt to improve the usability of large arrays in the UI. I moved the "Add Entry" button to the bottom, such that you don't have to scroll up when adding a new item to a long list.

grafik

To make it clearer what it belongs to, the button text includes the property name of the array. I've experimented with using the typename instead (of the entry type), but that doesn't work with assets (then it would say "Add String" instead of "Add Material").

I'm not sure about those extra buttons in the exposed parameters list. I didn't get them to right align. Not sure whether center alignment is good (probably not). Or maybe whether to move them somewhere entirely different.


PlasmaDev5:
small UI/UX feedback note
may look a little nicer if the button filled the same width as the item list above it. may also have to change height to have a good aspect ratio 

Mockup:
image

jankrassnigg:
> small UI/UX feedback note may look a little nicer if the button filled the same width as the item list above it. may also have to change height to have a good aspect ratio
> 
> Mockup: image

Maybe. I'm generally wondering whether there are more ways to make it fit better visually.

PR #1645, opened 2025-09-05
Added VSE node documentation comments

jankrassnigg:
These show up as tooltips on the nodes.


Data/Tools/ezEditor/VisualShader/Utils.ddl
jankrassnigg, line 1: Some comments in this file were AI-generated, since that was faster than Googling the details.

@C-Core maybe you can double-check that they are not completely wrong.

PR #1653, opened 2025-09-14
lowered velocity of a grabbed object released with motion

veksha:
(press E to grab object then move mouse and press E to release)

before and after:

https://github.com/user-attachments/assets/1d8bd675-b189-47a2-814d-e9baedbbeeb6


jankrassnigg:
Well this is just a hack. I would want to have a more general solution that also works consistently for different objects.

In the throw function you'll see that we use the weight categories to determine which force to apply. From the mass of the object and this impulse we can calculate the maximum velocity: 

https://chatgpt.com/s/t_68c6c4dfea00819181ce369cfe785669

We can then use that to check whether the current velocity is larger than that and clamp it to the maximum throw velocity. 

Then this would be consistent with the throw behavior and people could tweak it for every object category.

veksha:
I improved the code. I think this is going somewhere. but there is a hard-coded part. i don't know to get impulse from outside without refactoring / breaking changes. please, check.

veksha:
nice. now basketball games are possible with ezEngine! :)

PR #1657, opened 2025-09-17
TonemapPass: added WhitePoint property

veksha:
image


Data/Base/Shaders/Pipeline/Tonemap.ezShader
```diff
@@ -94,7 +94,7 @@ float4 main(PS_IN Input) : SV_Target
   // Tonemapping
   float3 color = FilmicTonemap(sceneColor.rgb);
 
-  float3 whiteScale = 1.0 / FilmicTonemap(11.2);
+  float3 whiteScale = WhitePoint / FilmicTonemap(11.2);
```
C-Core: Almost correct, the WhitePoint is the 11.2 here, so this line should be: 
float3 whiteScale = 1.0 / FilmicTonemap(WhitePoint);
It defines which value should be white after the tonemapping. Also don't forget to set the default value to 11.2 on the C++ so that the default behavior stays the same.
Otherwise this PR looks very good :)

jankrassnigg: and don't forget to adjust the property clamp range and default value

veksha: okay. i think 0..50 is ok range.

Code/Engine/RendererCore/Pipeline/Implementation/Passes/TonemapPass.cpp
```diff
@@ -52,6 +53,7 @@ ezTonemapPass::ezTonemapPass()
   m_fContrast = 1.0f;
   m_fLut1Strength = 1.0f;
   m_fLut2Strength = 0.0f;
+  m_fWhitePoint = 1.0f;
```
C-Core: Should also be 11.2, same default as in the reflection above

veksha: well spotted!

C-Core (approved):

PR #1667, opened 2025-09-22
Cleaned up Console implementation

jankrassnigg:
* Disentangled the ezConsole and ezQuakeConsole implementation.
* Moved ezQuakeConsole and ezCommandInterpreterLua into GameEngine.
* Fixed that switching the mouse cursor visibility at runtime didn't work. Turns out calling ShowCursor from another thread has no effect.
* Fixed that RenderFps and RenderConsole where called on another thread than the rest of the rendering.


veksha:
> * Fixed that switching the mouse cursor visibility at runtime didn't work. Turns out calling ShowCursor from another thread has no effect.

Hi. feedback from me about this "ShowCursor" change.
It works, but... It seems to change cursor visibility only on next WM_INPUT message. Cursor doesn't hide immediately. User must move mouse to make it disappear.

i have different approach.
what i do in my game is the following. i create task for this and it works ok:

```cpp
void Game1GameState::ShowCursor(bool bShow)
{
  // SetShowMouseCursor only works from Main Thread, so create task for this.
  auto pTask = EZ_DEFAULT_NEW(ezDelegateTask<void>, "HideCursor", ezTaskNesting::Never, [bShow]()
    {
                  auto pWindow = that->GetWindow();
                  auto pInputDevice = pWindow->GetInputDevice();
                  pInputDevice->SetShowMouseCursor(bShow); });
  pTask->ConfigureTask("HideCursor", ezTaskNesting::Maybe);
  ezTaskSystem::StartSingleTask(pTask, ezTaskPriority::ThisFrameMainThread);
}
```

get window:
```cpp
ezWindowPlatformShared* Game1GameState::GetWindow()
{
  ezDynamicArray<ezActor*> actors;
  ezActorManager::GetSingleton()->GetAllActors(actors);
  ezWindowBase* pWnd = actors[actors.GetCount() - 1]->GetPlugin<ezActorPluginWindowOwner>()->m_pWindow.Borrow();
  ezWindowPlatformShared* pWnd2 = (ezWindowPlatformShared*)pWnd;
  return pWnd2;
}
```

You can consider a trick like this. it hides cursor immediately.

PS: also, may i ask, why "while" loops are needed for your approach? they look scary for me:)

jankrassnigg:
Hm, I had the impression that the cursor shows up immediately. But I can take another look. The task trick would work, but it is quite a heavy thing to do :D

The while loop is indeed not supposed to be there, but in theory if some other code also calls ShowCursor, you can't guarantee that calling ShowCursor(true) once actually does show the cursor. Windows counts how many times it was called internally.
Maybe I'll remove it again, though.

PR #1668, opened 2025-09-23
Replaced Quake-style console with an ImGui based console

jankrassnigg:
This console is much more user-friendly.
It allows to tweak CVars from a window OR a command widget. It shows the log messages in a scroll area and allows for filtering. There are basic stats (time + memory).
The stats and log window can be pinned, which means they also appear as an overlay when the console is closed. When open, the mouse cursor is shown, and it isn't clipped, so opening the console can be used to click outside the window.

image

### DIFF
--- a/Code/Engine/GameEngine/DearImgui/Implementation/DearImgui.cpp
+++ b/Code/Engine/GameEngine/DearImgui/Implementation/DearImgui.cpp
@@ -314,7 +314,7 @@ void ezImgui::BeginFrame(const ezViewHandle& hView)
 
   ImGuiIO& cfg = ImGui::GetIO();
 
-  if (auto* pCVar = (ezCVarFloat*)ezCVar::FindCVarByName("Debug.TextScale"))
+  if (auto* pCVar = (ezCVarFloat*)ezCVar::FindCVarByName("App.TextScale"))
   {
     cfg.FontGlobalScale = *pCVar;
   }
--- a/Code/Engine/GameEngine/DearImgui/Implementation/DearImgui.cpp
+++ b/Code/Engine/GameEngine/DearImgui/Implementation/DearImgui.cpp
@@ -285,6 +285,7 @@ ImGuiContext* ezImgui::CreateContext()
   ImGui::SetCurrentContext(nullptr);
   ImGuiContext* context = ImGui::CreateContext(m_pSharedFontAtlas.Borrow());
 
+  m_pTextScaleCVar = (ezCVarFloat*)ezCVar::FindCVarByName("App.TextScale");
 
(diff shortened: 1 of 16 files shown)


Code/Engine/GameEngine/DearImgui/Implementation/DearImgui.cpp
```diff
@@ -314,7 +314,7 @@ void ezImgui::BeginFrame(const ezViewHandle& hView)
 
   ImGuiIO& cfg = ImGui::GetIO();
 
-  if (auto* pCVar = (ezCVarFloat*)ezCVar::FindCVarByName("Debug.TextScale"))
+  if (auto* pCVar = (ezCVarFloat*)ezCVar::FindCVarByName("App.TextScale"))
```
C-Core: The cvar should be cached somewhere. FindCVarByName does a linear search over all cvars which is not a good idea to do every frame.

Code/Engine/RendererCore/Debug/Implementation/DebugRenderer.cpp
```diff
@@ -16,7 +16,7 @@
 #include <RendererFoundation/Resources/BufferPool.h>
 #include <RendererFoundation/Shader/Types.h>
 
-ezCVarFloat cvar_DebugTextScale("Debug.TextScale", 1.0f, ezCVarFlags::Save, "Global scale for debug text");
+ezCVarFloat cvar_AppTextScale("App.TextScale", 1.0f, ezCVarFlags::Save, "Global scale for debug text");
```
C-Core, line 19: Should we move this cvar somewhere else since it is now used for more than just debug text?

Code/Engine/GameEngine/GameApplication/Implementation/GameApplication.cpp
```diff
@@ -28,6 +28,13 @@
 #include <Texture/Image/Formats/TgaFileFormat.h>
 #include <Texture/Image/Image.h>
 
+#ifdef BUILDSYSTEM_ENABLE_IMGUI_SUPPORT
+#  include <GameEngine/Console/ImGuiConsole.h>
+#  include <GameEngine/DearImgui/DearImgui.h>
+#endif
+
+#define USE_IMGUI_CONSOLE 1
```
C-Core: This define should be inside the BUILDSYSTEM_ENABLE_IMGUI_SUPPORT block above

Code/Engine/RendererCore/Debug/Implementation/DebugRenderer.cpp
jankrassnigg: was wondering the same, but don't have a good idea

PR #1672, opened 2025-09-26
Replace actor system by window manager

jankrassnigg:
Gets rid of the actor system, which had a confusing name and completely over-engineered functionality when it ultimately did something very trivial.

The replacement is called ezWindowManager, and while still being somewhat more complex than strictly necessary, it should be much more straight-forward to understand and use.


Code/Engine/Core/System/Implementation/WindowManager.cpp
```diff
@@ -0,0 +1,200 @@
...
+    }
+  }
+
+  for (const ezRegisteredWndHandle& id : toClose)
+  {
+    Close(id);
+  }
+}
+
+bool ezWindowManager::IsValid(ezRegisteredWndHandle id) const
```
Sanakan8472: This one is missing a ` EZ_LOCK(m_Mutex);`

Code/Engine/Core/System/Implementation/WindowManager.cpp
```diff
@@ -0,0 +1,200 @@
...
+  ezUniquePtr<Data>* pDataPtr = nullptr;
+  if (!m_Data.TryGetValue(id.GetInternalID(), pDataPtr))
+    return;
+
+  (*pDataPtr)->m_OnDestroy = onDestroyCallback;
+}
+
+ezStringView ezWindowManager::GetName(ezRegisteredWndHandle id) const
+{
+  if (!m_Data.Contains(id.GetInternalID()))
```
Sanakan8472: All the getter functions should also get locked or they might crash on concurrent modifications.

Code/Engine/Core/System/Implementation/WindowManager.cpp
```diff
@@ -0,0 +1,200 @@
...
+}
+
+ezWindowManager::~ezWindowManager()
+{
+  CloseAll(nullptr);
+}
+
+void ezWindowManager::Update()
+{
+  EZ_LOCK(m_Mutex);
```
Sanakan8472: This creates the potential for deadlocks as we are holding the lock while executing arbitrary callbacks. On the other hand, not holding the lock opens the door to the window getting closed while we are iterating on the copy. Not sure what the more likely scenario is for the current set of ProcessWindowMessages functions.
Maybe we should make `ezWindowBase` ref counted? It already has AddReference and RemoveReference functions but these are only used right now for ensuring that windows are not destroyed before rendering to them has stopped.

jankrassnigg: I actually ran into those ref counts on ezWindow recently and would have preferred to remove them...

I'm not entirely sure about the locking in general. We could also just remove this, because it is questionable, whether modifying this in a multithreaded way is really a good thing. I would also not expect this to happen in practice.

I think I only did this, because the actor system was doing it.

PR #1676, opened 2025-09-29
Refactored mesh vertex stream declaration

C-Core:
* Previously the vertex stream layout for meshes was completely flexible and allow for tons of different combinations. This made things unnecessarly complex and furthermore impossible to prebake PSOs in the future since the vertex stream declaration is part of it.
* Now meshes use a fixed layout of vertex streams where vertex data is split into multiple buffer instead of using one interleaved buffer for everything:
`Buffer 0: Position,
Buffer 1: Normal, Tangent and TexCoord0,
Buffer 2: TexCoord1,
Buffer 3: Color0,
Buffer 4: Color1,
Buffer 5: SkinningData`
This reduces the possible combinations and makes it possible to specify the stream layout with a simple bitmask in the shader in the future.
* The various precision options have been reduced to one high precision flag which makes every attribute high precision. This wastes some memory but is rarely used in practice anyways.
* Some cleanups in related rendering code

NOTE: All meshes have to be re-transformed since the runtime serialization changed quite a bit.


Code/Engine/RendererCore/Meshes/Implementation/MeshBufferUtils.cpp
```diff
@@ -45,11 +16,11 @@ EZ_BEGIN_STATIC_REFLECTED_ENUM(ezMeshTexCoordPrecision, 1)
   EZ_ENUM_CONSTANT(ezMeshTexCoordPrecision::_32Bit),
 EZ_END_STATIC_REFLECTED_ENUM;
 
-EZ_BEGIN_STATIC_REFLECTED_ENUM(ezMeshBoneWeigthPrecision, 1)
-  EZ_ENUM_CONSTANT(ezMeshBoneWeigthPrecision::_8Bit),
-  EZ_ENUM_CONSTANT(ezMeshBoneWeigthPrecision::_10Bit),
-  EZ_ENUM_CONSTANT(ezMeshBoneWeigthPrecision::_16Bit),
-  EZ_ENUM_CONSTANT(ezMeshBoneWeigthPrecision::_32Bit),
+EZ_BEGIN_STATIC_REFLECTED_ENUM(ezMeshBoneWeightPrecision, 1)
```
jankrassnigg: is this enum still used, at all?

Code/Engine/RendererCore/Meshes/Implementation/MeshBufferUtils_inl.h
```diff
@@ -63,29 +63,9 @@ EZ_ALWAYS_INLINE ezResult ezMeshBufferUtils::EncodeTexCoord(const ezVec2& vTexCo
 }
 
 // static
-EZ_ALWAYS_INLINE ezResult ezMeshBufferUtils::EncodeBoneWeights(const ezVec4& vWeights, ezArrayPtr<ezUInt8> dest, ezMeshBoneWeigthPrecision::Enum precision)
+EZ_ALWAYS_INLINE ezResult ezMeshBufferUtils::EncodeBoneWeights(const ezVec4& vWeights, ezArrayPtr<ezUInt8> dest, ezMeshBoneWeightPrecision::Enum precision)
```
jankrassnigg: still needed ?

jankrassnigg (approved):

Code/Engine/RendererCore/Meshes/Implementation/MeshBufferUtils.cpp
C-Core: It was still used in internally but I got rid of it now

Code/Engine/RendererCore/Meshes/Implementation/MeshBufferUtils_inl.h
C-Core: Removed

Sanakan8472 (approved):

PR #1690, opened 2025-10-17
Refactored dynamic mesh buffer

C-Core:
* The dynamic mesh buffer now uses the same vertex stream layout as regular meshes. 
* It now automatically uploads its data at the end of the extraction phase instead of an explicit update call.
* The default mesh renderer can now also render custom mesh render data (containing a dynamic mesh buffer). 
* Refactored the cloth sheet and jolt cloth sheet components to also use custom mesh render data.
* Removed custom mesh and cloth renderer since they are no longer needed.


Code/Engine/RendererCore/Meshes/DynamicMeshBufferResource.h
```diff
@@ -1,51 +1,54 @@
...
   {
     // store in [0; 1] range
-    m_vEncodedTangent.x = vTangent.x * 0.5f + 0.5f;
-    m_vEncodedTangent.y = vTangent.y * 0.5f + 0.5f;
-    m_vEncodedTangent.z = vTangent.z * 0.5f + 0.5f;
-    m_vEncodedTangent.w = fBitangentSign < 0.0f ? 0.0f : 1.0f;
+    m_vEncodedTangent.x = ezMath::ColorFloatToShort(vTangent.x * 0.5f + 0.5f);
+    m_vEncodedTangent.y = ezMath::ColorFloatToShort(vTangent.y * 0.5f + 0.5f);
+    m_vEncodedTangent.z = ezMath::ColorFloatToShort(vTangent.z * 0.5f + 0.5f);
+    m_vEncodedTangent.w = ezMath::ColorFloatToShort(fBitangentSign < 0.0f ? 0.0f : 1.0f);
```
jankrassnigg, line 48: why are these treated as colors?
also why would alpha value (sign bit) need to be treated as a color?

Code/EnginePlugins/GameComponentsPlugin/Physics/Implementation/ClothSheetComponent.cpp
```diff
@@ -259,87 +144,35 @@ ezResult ezClothSheetComponent::GetLocalBounds(ezBoundingBoxSphere& ref_bounds,
...
 
   if (m_hMaterial.IsValid())
   {
     ezResourceLock<ezMaterialResource> pMaterial(m_hMaterial, ezResourceAcquireMode::AllowLoadingFallback);
     category = pMaterial->GetRenderDataCategory();
+    bDontCacheYet = pMaterial.GetAcquireResult() == ezResourceAcquireResult::LoadingFallback;
   }
 
-  msg.AddRenderData(pRenderData, category, ezRenderData::Caching::Never);
+  msg.AddRenderData(pRenderData, category, bDontCacheYet ? ezRenderData::Caching::Never : ezRenderData::Caching::IfStatic);
```
jankrassnigg, line 175: doesn't this break rendering? the cloth is dynamic, bounding boxes change all the time

Code/EnginePlugins/JoltPlugin/Components/Implementation/JoltClothSheetComponent.cpp
```diff
@@ -84,25 +58,13 @@ EZ_END_DYNAMIC_REFLECTED_TYPE;
...
-}
-
 void ezJoltClothSheetComponent::SerializeComponent(ezWorldWriter& inout_stream) const
 {
   SUPER::SerializeComponent(inout_stream);
   auto& s = inout_stream.GetStream();
 
   s << m_vSize;
-  s << m_vSegments;
+  s << m_vNumVertices;
```
jankrassnigg, line 67: just a rename ?

jankrassnigg (approved):

Code/Engine/RendererCore/Meshes/DynamicMeshBufferResource.h
C-Core: Its just the name of the conversion function, has nothing to do with colors in this case.

Code/EnginePlugins/JoltPlugin/Components/Implementation/JoltClothSheetComponent.cpp
C-Core: Yes only rename. Segments was confusing since it meant number of vertices in this case while it really meant number of segments for the regular (not jolt) cloth sheet component.

Code/EnginePlugins/GameComponentsPlugin/Physics/Implementation/ClothSheetComponent.cpp
C-Core: No, the bounding box can still change, doesn't matter for the render data.

PR #1704, opened 2025-10-29
World space RML UI

x0r04rg:
Adds a component that displays RML UI on a 3d mesh. It is possible to interact with the UI by sending it any input via RaytraceInput and ReceiveInput methods.

If there is anything you would do differently just let me know and I'll rewrite it, especially the rendering part since I'm not that familiar with rendering in EZ.


jankrassnigg:
Very cool! We'll need some time to try it out and go through the code. Currently the component isn't used in any scene where one could try it out. Would you be able to add a minor example somewhere? Doesn't need to actually make sense, but I am wondering how the input integration needs to be triggered, because as far as I can tell this has to be done from the outside, right?

C-Core:
First of all thanks a lot for this contribution and very impressive that you managed to pull that off on your own!
I looked through the code and there are a bunch of things to improve and cleanup:
* The two RmlUiCanvas components share a lot of code, I think it would make sense to move all common functionality to a RmlUiCanvasComponentBase class
* The new UpdateInput function in the RmlUiContext is almost identical to the old one. I would delete the old one and add an input provider to the 2D component as well, which is then filled with the keyboard input, just like it was done in the old UpdateInput function.
* We already have ezIntersectionUtils::RayTriangleIntersection, please make use of that instead of writing your own ray triangle intersection code

Now to the rendering part, the 3D canvas component should not do any mesh rendering itself:
* The 3D canvas component should have a material, a material slot index and a texture slot name. When its texture changes it creates a new, unique, material with the given material as base and sets its texture on the given texture slot. 
You can take a look at the skybox component which does something similar. Then it will send a ezMsgSetMeshMaterial to its owner game object (this can be extended to an arbitray target object in the future, but not necessary right away). 
This way the rendering can be done by any other component that handles the ezMsgSetMeshMaterial message and allows for custom shaders etc.
* The OnMsgExtractRenderData is still needed to trigger the rendering of the RmlUi to the texture. GetLocalBounds and OnMsgExtractGeometry are not needed anymore.
* The 3D canvas can still have a mesh for input ray casting, which then can be a different, maybe simplified mesh, to the render geometry. It would be nice if there was a fallback to a simple plane (would need a normal direction (ezBasisAxis?) and extents properties).

Let me know if you have any questions or run into issues (feel free to ping me on discord).

x0r04rg:
Thanks for the feedback, guys! I'll start by adding an example to the Testing Chambers and then will gradually implement the changes proposed by @C-Core.

x0r04rg:
@jankrassnigg The example is ready, let me know if there is something missing

jankrassnigg:
Nice!

image

x0r04rg:
Seems like it's all done. There is also a minor issue with how material assignment works - I'll bring it up in discord.

Code/EnginePlugins/RmlUiPlugin/Implementation/RmlUiContext.cpp
```diff
@@ -74,61 +61,48 @@ void ezRmlUiContext::HideDocument()
...
     }
   }
 
   // Keyboard
   {
-    ezUInt32 uiLastChar = ezInputManager::RetrieveLastCharacter(false);
-
-    if (uiLastChar >= 32 || uiLastChar == '\n') // >= space (+ enter/return)
+    ezUInt32 uiLastChar = input.m_uiLastCharacter;
+    if (uiLastChar >= 32) // >= space
```
C-Core: Any reason why the enter/return check has been removed?

Code/EnginePlugins/RmlUiPlugin/Components/Implementation/RmlUiCanvas3DComponent.cpp
```diff
@@ -0,0 +1,447 @@
...
+    msg.m_hMaterial = m_hMaterial;
+    msg.m_uiMaterialSlot = m_uiMaterialIndex;
+    GetOwner()->SendMessage(msg);
+
+    m_bRebindTexture = true;
+  }
+
+  if (m_bRebindTexture)
+  {
+    ezResourceLock<ezMaterialResource> pMaterial(m_hMaterial, ezResourceAcquireMode::AllowLoadingFallback);
```
C-Core: You could also just use BlockTillLoaded since the material is created in code anyways and never loaded from disk.

C-Core (approved):
Looks very good now! Again, thanks a lot for putting in all the work!

Code/EnginePlugins/RmlUiPlugin/Implementation/RmlUiContext.cpp
jankrassnigg: That one is needed for text boxes to work correctly.

x0r04rg: Good catch, it was added after I'd began working on this and I somehow missed it during the merge or something. I'll fix it

x0r04rg:
done

PR #1712, opened 2025-11-16
Color gradients can now be loaded as textures

jankrassnigg:
Just reference them in a material, the texture loader automatically creates a 1K 1D texture for them. Also fixed minor inconveniences in RML UI and cleared the particle event queue (might have been a bug, but not sure).

### DIFF
--- a/Code/Engine/RendererCore/Textures/TextureLoader.cpp
+++ b/Code/Engine/RendererCore/Textures/TextureLoader.cpp
@@ -1,5 +1,6 @@
 #include <RendererCore/RendererCorePCH.h>
 
+#include <Core/Curves/ColorGradientResource.h>
 #include <Foundation/Configuration/CVar.h>
 #include <Foundation/Configuration/Startup.h>
 #include <Foundation/IO/FileSystem/FileReader.h>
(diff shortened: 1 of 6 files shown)


Code/Engine/RendererCore/Textures/TextureLoader.cpp
```diff
@@ -101,43 +102,105 @@ ezResourceLoadData ezTextureResourceLoader::OpenDataStream(const ezResource* pRe
...
         return res;
+
+      ezColorGradientResourceDescriptor desc;
+      desc.Load(File);
+
+      pData->m_TexFormat.m_bSRGB = true;
+
+      const ezUInt32 uiHeight = 1;
+
+      // set up a 1024 * 4 uncompressed RGBA texture
```
C-Core: Comment ist outdated, I think a 512x1 texture should be sufficient.

Code/Engine/RendererCore/Textures/TextureLoader.cpp
```diff
@@ -101,43 +102,105 @@ ezResourceLoadData ezTextureResourceLoader::OpenDataStream(const ezResource* pRe
...
+      ezImageHeader header;
+      header.SetWidth(1024);
+      header.SetHeight(uiHeight);
+      header.SetDepth(1);
+      header.SetImageFormat(ezImageFormat::R8G8B8A8_UNORM_SRGB);
```
C-Core: What about HDR gradients?

jankrassnigg: I wasn't sure about those, so I left that problem up to future me.

jankrassnigg: Made it always HDR, shouldn't be a problem for a 512x1 sized texture.

PR #1720, opened 2025-11-23
Try to find the preload collection for the startup scene

### LINKED ISSUE #1648: [BUG] Pink outlines are visible during world/model loading
**Bug description:**

Object outlines are visible during model loading.
For example, in my game it happens on first weapon switch.

**How to Reproduce:**

1. open Testing Chambers, main scene.
2. export and run
3. see how models are having pink outlines while they load.

(this is with "Shipping" build type)

**Screenshots:**

Image

C-Core:
Note that if the preload collection has the same name as the scene it will be used automatically to preload all assets.

Fixes #1648


jankrassnigg (approved):
Does this also need fixes for the RTS sample or any other sample that uses a game state?

PR #1726, opened 2025-11-28
Mesh particles now use sub-meshes for variation

jankrassnigg:
If a mesh particle renderer references a mesh that has multiple sub-meshes, the sub-meshes are randomly selected for the particle. This way you can put multiple meshes for, e.g. broken pieces, into a mesh to have more variety in the effect.

Also added a scale parameter to the mesh renderer, to adjust the size of the mesh independently of the particle size.


Code/EnginePlugins/ParticlePlugin/Type/Mesh/ParticleTypeMesh.cpp
```diff
@@ -219,32 +265,68 @@ void ezParticleTypeMesh::ExtractTypeRenderData(ezMsgExtractRenderData& ref_msg,
...
+        pRenderData->m_vGlobalPosition = GetOwnerSystem()->GetTransform().m_vPosition;
+        pRenderData->Fill(m_InstanceDataOffset, hInstanceDataBuffer, m_hMaterial, m_hMesh, 0, 0, numParticles);
 
-      ref_msg.AddRenderData(pRenderData, m_RenderCategory, ezRenderData::Caching::Never);
+        ref_msg.AddRenderData(pRenderData, m_RenderCategory, ezRenderData::Caching::Never);
+      }
+      else
+      {
+        // Multi-submesh path: need to create render data per particle with correct submesh
+        for (ezUInt32 p = 0; p < numParticles; ++p)
```
C-Core: Shouldn't that be the same logic as in the non opaque path above? Can we de-duplicate it?

jankrassnigg: done

PR #1727, opened 2025-12-01
Add ezSimdDouble class

TrevorCash:
Adds ezSimdDouble class with AVX implementation.

Tested on x86_64 Linux with a corresponding unit test. 

Needs feedback on the Sqrt and Recipricol Related correctness.


C-Core:
To be honest, I don't see the point in why we should have a simd double implementation. It is not something commonly used in games. And even if we want to add this, it should be an SSE level, (we are currently on SSE 4.1, so something like SSE AVX) and not orthogonal to SSE. Also you cannot simply set AVX compiler flags because that will make AVX a minimum requirement which means it will crash on a non-AVX cpu.

jankrassnigg:
I do think that having a double version could be useful, since we may want to support double precision for game object positions at some point. So having this is generally fine IMO. But I agree that we can't just enable the AVX requirement, that would break the engine for many people.

TrevorCash:
All The intrinsics are AVX (1).  So that should be about a 95% support rate according to steam.  I added EZ_REQUESTED_SIMD_LEVEL for control during the cmake process (with higher numbers generally meaning more SIMD capability).  Defaulting to the current SSE implementation.  I'd like to see Double precision support for Transformations.

C-Core:
@TrevorCash I opened a PR with the CMake setup how I envisioned it: https://github.com/ezEngine/ezEngine/pull/1731
This also allows us to go even lower than SSE 4.1 if needed since we also had complains about that sometimes.
I've added a stub of your SimdDouble class so you can rebase this PR on top of it once it is in. 

Note that ezSimdDouble should always be double precision no matter which SSE level is selected, we can fallback to FPU if you don't want to deal with implementing the 2 __m128d variant (which I can totally understand, I wouldn't want to do it either ;) ).

TrevorCash:
for EZ_SIMD_IMPLEMENTATION = AVX, ezSimdDouble uses generally 1 AVX instruction per function. other classes use SSE implementation files unless AVX needed (ezSimdDouble, and future vec4d and similar types) . 

for EZ_SIMD_IMPLEMENTATION = SSE, ezSimdDouble uses generally 2 SSE instructions per function.

Code/Engine/Foundation/Basics/AllDefinesOff.h
```diff
@@ -92,15 +92,19 @@
 #define EZ_SIMD_IMPLEMENTATION_FPU 1
 #define EZ_SIMD_IMPLEMENTATION_SSE 2
 #define EZ_SIMD_IMPLEMENTATION_NEON 3
+#define EZ_SIMD_IMPLEMENTATION_AVX 4
```
C-Core, line 95: Please revert these changes, we don't want to treat AVX as a separate simd implementation but as an advanced version of SSE (as discussed in my other PR).

Code/Engine/Foundation/SimdMath/Implementation/SSE/SSEDouble_inl.h
```diff
@@ -1,3 +1,393 @@
...
+  m_v.zw = _mm_set1_pd(d);
+}
+
+EZ_ALWAYS_INLINE ezSimdDouble::ezSimdDouble(ezInt32 i)
+{
+  EZ_CHECK_SIMD_ALIGNMENT(this);
+
+  __m128i packedInt32 = _mm_set1_epi32(i);
+  m_v.xy = _mm_cvtepi32_pd(packedInt32);
+  m_v.zw = _mm_cvtepi32_pd(packedInt32);
```
C-Core, line 38: m_v.zw = m_v.xy; no need to convert the same thing twice. This goes for all operations here. For simddouble we can assume that x,y,z and w all contain the same value.

Code/Engine/Foundation/SimdMath/Implementation/SSE/SSEVec4f_inl.h
```diff
@@ -578,41 +578,25 @@ EZ_ALWAYS_INLINE ezSimdVec4f ezSimdVec4f::CrossRH(const ezSimdVec4f& v) const
 // static
 EZ_ALWAYS_INLINE ezSimdVec4f ezSimdVec4f::MulAdd(const ezSimdVec4f& a, const ezSimdVec4f& b, const ezSimdVec4f& c)
 {
-#if EZ_SSE_LEVEL >= EZ_SSE_AVX2
```
C-Core, line 581: This should not be deleted.

C-Core (changes requested):

TrevorCash:
I implemented your suggestion for removing duplicate operations thanks for that.  

I Also Reverted the changes to SSE/SSEVec4f_inl.h (I removed the comments because the possibility of adding AVX/AVXVec4f_inl.h)

Removing #define EZ_SIMD_IMPLEMENTATION_AVX 4 really doesnt make sense to me - as it lives alongside NEON and SSE as another equal instrinic technology.

TrevorCash:
Sorry I closed the thread while I was re-organising things.  I re-opened the PR here: https://github.com/ezEngine/ezEngine/pull/1733. Sorry to change threads.

Code/Engine/Foundation/SimdMath/Implementation/SSE/SSEDouble_inl.h
TrevorCash: duplicate commands should be removed in https://github.com/ezEngine/ezEngine/pull/1733

Code/Engine/Foundation/SimdMath/Implementation/SSE/SSEVec4f_inl.h
TrevorCash: if a AVX/AVXVec4f_inl.h file were created should the #if EZ_SSE_LEVEL >= EZ_SSE_AVX2 be removed from this file?  For EZ_MIN_REQUIRED_SSE_LEVEL = SSE4.1 the instructions in the ifdef should not evaluate anyway.

PR #1729, opened 2025-12-01
Fixed #1642: Removed "default value" for Visual Shader texture and parameter nodes

### LINKED ISSUE #1642: [BUG] (Visual Shader Editor) updating texture in "2D Texture" node properties should update it in Shader Properties
**Bug description:**
In Visual Shader Editor when you update texture in "2D Texture" node properties - it's value isn't updated in Shader Properties (which are shown when you click on empty space near the nodes).

**How to Reproduce:**

https://github.com/user-attachments/assets/2616321c-b02c-4e16-b2ee-8381c9d27eaa

**When fixed it should look like:**

https://github.com/user-attachments/assets/b3192861-f156-4db1-b9d2-dbb815145eaf

jankrassnigg:
Gets rid of a redundant place to modify the default values.
I hope this doesn't break anything.

### DIFF
--- a/Data/Tools/ezEditor/VisualShader/Parameters.ddl
+++ b/Data/Tools/ezEditor/VisualShader/Parameters.ddl
@@ -4,19 +4,13 @@ Node %Parameter1f
   string %Color { "Red" }
   string %Docs { "Outputs a 1-component value that can be configured on the material." }
 
-  string %CodeMaterialParams { "float $prop0 @Default($prop1);" }
+  string %CodeMaterialParams { "float $prop0;" }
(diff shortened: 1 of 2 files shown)


C-Core:
Does it still work to create a new parameter node and it shows up correctly in the material properties? I would be surprised since without the default value you don't have a type anymore.

jankrassnigg:
> Does it still work to create a new parameter node and it shows up correctly in the material properties? I would be surprised since without the default value you don't have a type anymore.

Works:

image

There is a "NodeType" property in the visual shader description, so this doesn't seem to be necessary for the material to know the type.

PR #1731, opened 2025-12-02
CMake setup for SSE levels and Simd Double stub

C-Core:


jankrassnigg (approved):

TrevorCash:
suggestion: EZ_MIN_REQUIRED_SSE_LEVEL -> EZ_MIN_REQUIRED_SIMD_LEVEL because technically AVX is not part of SSE.  NEON is also not technically part of SSE.  SIMD encapsulates all intrinsics technologies.

C-Core:
Technically AVX is not part of SSE yes, but it is the next logical step on x86. The cmake setting is specific for x86 so that we can precicely control the min requirements from the outside.
If we ever need something like that for NEON we will introduce another cmake variable for that.

PR #1736, opened 2025-12-08
Custom shader support for particles

jankrassnigg:
Particles can now be rendered with custom shaders. In the particle "quad" or "trail" type, select "Custom Material". For the custom particle material, create a custom shader, either through the shader template, or via visual shaders (VSE). Important: On the material asset set the "Asset Filter Tag" to "QuadParticle" or "TrailParticle", to be able to pick the material in the asset browser as a custom particle material. Using a custom shader you can create more advanced rendering effects. In VSE there are a particle specific nodes for input and output. VSE currently only supports quad particles (no trails), but that would be trivial to add, as well.

Also fixed a few issues in VSE. The same texture can now be sampled multiple times with different UVs by using the texture 2D node with the same name multiple times. Also removed the hard coded "BaseTexture" and other nodes. Also added a "ScreenCoord" node for sampling the screen textures.

Currently there is only a single sample, the "CustomParticleShader" material in the Testing Chambers project.

Custom shader:

pfx shader

In particle editor:

image

In use:

image


Data/Tools/ezEditor/ShaderTemplates/Particle.ezShaderTemplate
```diff
@@ -0,0 +1,122 @@
...
+// See http://ezengine.net/pages/docs/graphics/shaders/shader-permutation-variables.html
+
+%if TEMPLATE_TYPE == TEMPLATE_TYPE_QUAD
+PARTICLE_QUAD_MODE
+%elif TEMPLATE_TYPE == TEMPLATE_TYPE_TRAIL
+PARTICLE_TRAIL_POINTS
+%endif
+
+PARTICLE_RENDER_MODE
+PARTICLE_QUAD_MODE
```
C-Core: PARTICLE_QUAD_MODE is already listed above

PR #1743, opened 2025-12-15
changed turret sound to more appropriate and less annoying one

veksha:
what is changed:
in Turret.ezPrefab - changed AttachAsChild to "true"
in Turret.as - instead of firing every Update tick I check for child with empty name and rename it to "ProjectileFired" and do triggering of the sound.

also sound of firing is changed to "Pulse" which is more appropriate.

### DIFF
--- a/Data/Samples/Testing Chambers/Prefabs/Turret.ezPrefab
+++ b/Data/Samples/Testing Chambers/Prefabs/Turret.ezPrefab
@@ -21,7 +21,7 @@ o
 		s %AssetType{"Prefab"}
 		VarArray %Dependencies{}
 		Uuid %DocumentID{u4{1604195568484542123,4944665773883242456}}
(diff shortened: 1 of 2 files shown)


jankrassnigg:
You can't use "AttachAsChild" for projectiles, because then when the turret rotates, the projectiles also move around it. Stuff that moves on its own shouldn't be a child object of a dynamic parent.

There is a return value on the spawn function that tells you whether the object actually got spawned (it only spawns every few milliseconds). You can just use that to trigger the sound.

veksha:
ScheduleSpawn doesn't return bool, but I can use Blackboard.

veksha:
oh, you mean i can replace ScheduleSpawn with TriggerManualSpawn ? i'll try

veksha:
done. please, check.

PR #1745, opened 2025-12-17
Added fog plane example material

jankrassnigg:
The 'Main' scene in the Testing Chamber example contains a basic example material for how to make a ground fog effect.

image

Also fixed a bug in a visual shader node and added a 'pixel depth' node (though not used for this example).


veksha:
nice! but I found that the border is not smooth when looking outside while standing in the fog.

image

jankrassnigg:
> nice! but I found that the border is not smooth when looking outside while standing in the fog.
> image

Yes, placing the object properly in that terrain is a bit tricky. I've changed it a bit.
And yes, I know that the decals from the exploding barrels project onto that plane. Don't have a solution for that atm.

PR #1758, opened 2026-01-04
Skip particles with not-yet-loaded custom materials

### LINKED ISSUE #1747: [BUG] Custom particle materials trigger DX11 error
**Bug description:**
A custom particle material will trigger a DX11 error in exported scenes since it tries to render quad particles with the default loading material.

D3D11 ERROR: ID3D11DeviceContext::Draw: The Vertex Shader expects application provided input data (which is to say data other than hardware auto-generated values such as VertexID or InstanceID). Therefore an Input Assembler object is expected, but none is bound. [ EXECUTION ERROR #349: DEVICE_DRAW_INPUTLAYOUT_NOT_SET]
D3D11: **BREAK** enabled for the previous message, which was: [ ERROR EXECUTION #349: DEVICE_DRAW_INPUTLAYOUT_NOT_SET ]

**How to Reproduce:**
1. Use the Testing chambers project and place "Effects\CustomParticleShader.ezParticleEffectAsset" in a scene other than the main scene. (The main scene uses a preload collection which circumvents the problem)
2. Export the scene
3. Run it with a debug build of ezPlayer and the above error should occur

**Dev-branch commit hash:**
882b6293671f9128bcc625ce7911d867e7aaf092

jankrassnigg:
Fixes #1747 (hopefully)

Attempts to prevent rendering errors, since fallback materials would use the wrong shader. I couldn't reproduce the original issue, though, potentially due to different loading speeds on my machine.

However, this is only a band-aid, since doing this in the particle code doesn't guarantee that the shader is loaded (material could be loaded, but shader isn't).

Generally, we may need to rethink the fallback approach for materials. It might be better to bind a material and provide a fallback material that could be used. If none is available, the renderer should skip rendering. This can only be decided at the drawcall when material and shader references get resolved.


C-Core (approved):
Works for me now with this fix

PR #1764, opened 2026-01-06
Collision mesh writer and other fixes

C-Core:
* Added a new collision mesh writer singleton that keeps all collision mesh serialization code in one place and also enables the jolt plugin to write collision meshes in the Editor without linking against the jolt cooking dll. Will be used soon by the spline mesh feature.
* Changed the jolt cooking code to only allow triangles instead of arbitrary polygons. We used triangles in most code paths anyways.
* Fixed a few issues with procedural vertex colors
* Fixed that directional light shadows could have objects popping in and out depending on the view direction for larger scenes
* Other small fixes


Code/EditorPlugins/Jolt/EnginePluginJolt/CollisionMeshAsset/JoltCollisionMeshWriter.h
```diff
@@ -0,0 +1,98 @@
...
+#include <EnginePluginJolt/EnginePluginJoltDLL.h>
+
+#include <Foundation/Configuration/Singleton.h>
+#include <Foundation/IO/ChunkStream.h>
+#include <Foundation/IO/CompressedStreamZstd.h>
+#include <Foundation/Utilities/AssetFileHeader.h>
+#include <JoltCooking/JoltCooking.h>
+#include <JoltPlugin/Resources/JoltMeshResourceWriter.h>
+
+class ezJoltCollisionMeshWriter : public ezJoltMeshResourceWriterInterface
```
jankrassnigg: this file could use some comments

Code/EnginePlugins/JoltPlugin/Resources/JoltMeshResourceWriter.h
```diff
@@ -0,0 +1,37 @@
...
+  ezUInt32 m_uiMaxConvexPieces = 1;
+
+  ezDynamicArray<ezVec3> m_Vertices;
+  ezDynamicArray<ezUInt32> m_TriangleIndices;
+  ezDynamicArray<ezUInt16> m_TriangleSurfaceID;
+
+  ezDynamicArray<ezString> m_Surfaces;
+};
+
+/// \brief Interface for writing jolt mesh resources. Used internally by ezJoltMeshResourceUtils::WriteMeshResource.
```
jankrassnigg: I don't get why this is an interface and why that would still be Jolt specific.
If the code is Jolt specific anyway, why not just access the real class right away?
If it is an interface, why would it still be Jolt specific?

Also could use some more comments.

C-Core: I need to be able to write collision meshes from the jolt plugin. But that is only possible with the jolt cooking dll. The jolt plugin should not have cooking as dependency so the interface is implemented in the jolt engine plugin which already links against cooking.

jankrassnigg: Ok, but then the expectation is, that the collision meshes only get generated in editor scenarios (export etc), right?

The JoltCooking library is a bit of an unfortunate remnant from the PhysX days. There physx cooking was a separate (large) library. In Jolt that's not the case, all collision mesh preparation code is in Jolt anyway (which we also use for the breakable stuff).

So maybe it's actually easier to just get rid fo the Jolt cooking library and just put that very small amount of code in the general Jolt plugin?

C-Core: Moved all the "cooking" code into the ezJoltMeshResourceWriter and removed the cooking dll.

jankrassnigg (approved):

PR #1767, opened 2026-01-08
Added StencilRef value and Visual Shader State Nodes

### LINKED ISSUE #1749: No way to specify the stencil reference value
In the shader render state section there is no way to configure the stencil reference value. Generally in the code, `ezGALCommandEncoder::SetStencilReference()` isn't called from anywhere.

Not sure whether this state needs a dedicated PSO, so not sure whether it should be set from the shader, or whether there should be other ways to dynamically modify it's value (using the same shader), but I wouldn't know how to best expose this.

jankrassnigg:
Visual Shaders can now set the render state through state nodes. Supports depth state, stencil state, blend state and rasterizer states. Added that one can set the StencilRef value from a shader (fixes #1749).

### DIFF
--- a/Code/Engine/RendererCore/RenderContext/Implementation/RenderContext.cpp
+++ b/Code/Engine/RendererCore/RenderContext/Implementation/RenderContext.cpp
@@ -302,6 +302,12 @@ void ezRenderContext::SetRasterizerState(ezGALRasterizerStateHandle hRasterizerS
   m_StateFlags.Add(ezRenderContextFlags::PipelineChanged);
 }
 
+void ezRenderContext::SetStencilRefValue(ezUInt8 uiStencilRefValue)
+{
+  m_uiUserStencilRefValue = uiStencilRefValue;
+  m_StateFlags.Add(ezRenderContextFlags::NonPipelineStateChanged);
+}
+
 void ezRenderContext::BindMeshBuffer(const ezMeshBufferResourceHandle& hMeshBuffer, ezGALBufferHandle hDataOffsetsBuffer /*= {}*/, ezUInt32 uiFirstDataOffset /*= 0*/)
 {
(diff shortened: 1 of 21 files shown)


Code/Engine/RendererCore/RenderContext/Implementation/RenderContext.cpp
```diff
@@ -576,9 +576,16 @@ ezResult ezRenderContext::ApplyContextStates(bool bForce)
   if (bForce || m_StateFlags.IsSet(ezRenderContextFlags::PipelineChanged))
   {
     if (m_bRendering)
+    {
       m_pGALCommandEncoder->SetGraphicsPipeline(ezGALPipelineCache::GetPipeline(m_GraphicsPipeline));
+
+      // Set stencil reference value from shader
+      m_pGALCommandEncoder->SetStencilReference(m_uiStencilRefValue);
```
C-Core: This call should only be done if the shader actually wants to set a stencilrefvalue. Otherwise this would always overwrite any ref value set in code before the drawcall.

C-Core (changes requested):

Code/Engine/RendererCore/RenderContext/Implementation/RenderContext.cpp
jankrassnigg: The problem is, how do we even want to handle setting the stencil ref value manually?
Should the shader keep whatever previous value was set untouched, if the user doesn't set it in the shader?
I'm not sure, whether that's a good solution, but I also wouldn't have a better idea right now.
Maybe use 0 as default, UNLESS the user set it to -1 or so, indicating, that an external value should be used?

jankrassnigg: See new commit for changes to make this possible.

C-Core (approved):

PR #1769, opened 2026-01-11
AssetProcessor timeline widget

Sanakan8472:
# Curator Panel Upgrade

 * The Asset Curator panel now has a new widget on top that shows the timeline of processed assets in the `ezAssetProcessor`. The widget shows a timeline view with Y-axis representing each `ezEditorProcessor` process and X-axis representing time. Each asset being processed is displayed as a bar on the timeline. The timeline is compacted in that every length of time in which no asset processing happens is reduced to one second in the timeline and a diagonal pattern is drawn to indicate that the timeline cuts off there.
 * To support this, the `ezAssetProcessor` now has an event that is fired when processing of an asset starts of stops.

image

 # Bug Fixes

 * Fixed a bug in which the `ezAssetProcessor` would repeatedly try to update the same asset. This was caused by some old code in the `ezAssetCurator::AddAssetHash` function that was not properly migrated in the past. The code falsely incorporated package hashes into the thumbnail hash. This meant that package hash changes would affect the thumbnail hash. As there is no inverse dependency tracking of package hashes, changes to the package hash would not propagate upwords to dependent assets and invalidat those. However, as the `ezEditorProcessor` forces reevaluation of the transitive hull of dependencies it would update the thumbnail hash du to changes to the packaging and result in generating an asset for a different hash than the editor requested. The fix is simply to not add package hashes to Thumbnail hashes. Each hash type should be independent of the others. I.e. only assets that are in e.g. the transitive hull of thumbnail depndencies should affect the thumbnail hash. Transform or package dependencies should not affect the thumbnail hash. If we made this assumption implicitly in the past, we can change the hash function again but it should not incorporate the packaging dependencies into any other category.

 # Misc

 * Renamed various classes around the `ezAssetProcessor` to be more appropriately named.
 * Some code was written to detect hash missmatches between the `ezEditorProcessor` and the editor. As this can be useful if another bug in the code pops up, the code was left in but does not have any affect on the transform outcome. See `ezEditorProcessorProcess::HandleHashMissmatch`. We could make the transform fail in case a hash missmatch is registered but I assume that this can happen in the real world if the user is changing assets while processing is in flight. This is fine as long as the state of both editor and `ezEditorProcessor` is eventually consistent and does not run in circles like the bug above.

### DIFF
--- a/Code/Editor/EditorEngineProcessFramework/IPC/ProcessCommunicationChannel.cpp
+++ b/Code/Editor/EditorEngineProcessFramework/IPC/ProcessCommunicationChannel.cpp
@@ -189,5 +189,8 @@ ezResult ezProcessCommunicationChannel::WaitForConnection(ezTime timeout)
 
 bool ezProcessCommunicationChannel::IsConnected() const
 {
+  if (!m_pChannel)
+    return false;
+
   return m_pChannel->IsConnected();
 }
--- a/Code/Editor/EditorFramework/Assets/AssetCurator.h
+++ b/Code/Editor/EditorFramework/Assets/AssetCurator.h
@@ -353,7 +353,7 @@ public:
   /// along with all their dependencies to the destination folder. Files are copied preserving
   /// their relative paths from the data directories. Each file is copied only once, even if
   /// it appears in multiple dependency trees.
-  ExportResult ExportAssets(ezArrayPtr<ezString> sources, ezStringView sDestinationFolder, bool bIncludeTransformDeps = true, bool bIncludeThumbnailDeps = true, bool bIncludePackageDeps = true) const;
+  ExportResult ExportAssets(ezArrayPtr<ezString> sources, ezStringView sDestinationFolder, ezBitflags<ezDependencyFlags> includeDependencyTypes = ezDependencyFlags::Transform | ezDependencyFlags::Thumbnail | ezDependencyFlags::Package) const;
 
   ///@}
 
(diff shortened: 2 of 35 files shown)


jankrassnigg:
The widget needs a scrollbar, once the horizontal space is full, I don't see any updates.

Sanakan8472:
> The widget needs a scrollbar, once the horizontal space is full, I don't see any updates.

You can drag the view with the RMB and zoom in/out with mouse wheel. It's basically a modified property animation timeline. But I'll add a scrollbar to make it easier to understand.

jankrassnigg:
> > The widget needs a scrollbar, once the horizontal space is full, I don't see any updates.
> 
> You can drag the view with the RMB and zoom in/out with mouse wheel. It's basically a modified property animation timeline. But I'll add a scrollbar to make it easier to understand.

Yeah I found that a bit later, but it's not obvious. Also by default it should just keep scrolling to the right, as elements get added.

jankrassnigg:
Maybe out of scope for this PR, but I found a bit weird asset processing behavior. I cleared all the caches for the test project and let it process all assets. Something failed to process (which it shouldn't but that's only problem 1). Then the next 10 or so transforms that were supposed to run on that processor also failed, with a timeout. So it looks like assets are queued on a processor, but if it crashes, the editor keeps waiting for those transforms to finish and just assumes they timed out after a while, when in fact the processor wasn't even running anymore.

jankrassnigg:
I would also change the icon for an inactive processor to something dim grey or so, the yellow dot icon looks still very "active". Also I'm wondering whether the coloring of the tasks should indicate whether this was a transform or a thumbnail generation (though as far as I understand it, sometimes it's both).

And finally, the flamegraph and the log now take up a lot of screenspace below each other, maybe they should be tabbed ? Because with the flamegraph open, I usually don't also care about the text log.

Sanakan8472:
@jankrassnigg I can confirm that a crashing ezEditorProcessor is not picked up by the system. For some reason the check on the QProcess does not work. I'll look into that.

Regarding the coloring: Blue is "transform needed" and green is "thumbnail needed", i.e. the state of the asset before we started working on it. It's the same colors used in the ezQtAssetProcessorProgressWidget. However, a "transform needed" asset will also create a thumbnail of course if the asset type support that. We can check the asset type whether it supports thumbnails to improve the coloring and tooltip but for more detailed graphs more info needs to be passed between the processes besides start and finish processing.

Regarding the log, I think I'll remove it as it is kind of redundant and only outputs start and stop infos. I just need to have a look if it provides any other benefits right now besides start / stop messages and errors.

I'll make a pass over these issues in a second PR after this one.

jankrassnigg (approved):

PR #1771, opened 2026-01-12
Spline Meshes

C-Core:
Added a new component to generate arbitrary meshes along a spline. It can have different start and end parts and multiple middle parts which are chosen at random or by best fit depending on the selected distribution mode.
Screenshot 2025-12-18 213549

Also added a component that can generate the corresponding collision mesh by specifying a mapping between render and collision meshes.


Code/Engine/RendererCore/Components/SplineComponent.h
```diff
@@ -190,6 +205,8 @@ class EZ_RENDERERCORE_DLL ezSplineComponent : public ezComponent
   ezBitflags<ezSplineComponentFlags> m_SplineFlags; // [ property ]
   ezUInt8 m_uiDummy = 0;
   mutable ezUInt16 m_uiExtractedFrame = 0;
+
+  ezUuid m_Uuid;
```
jankrassnigg, line 209: what is this needed for ?

jankrassnigg (approved):

Code/Engine/RendererCore/Components/SplineComponent.h
C-Core: It is used to generate a unique name for e.g. the generated mesh resource which is cached to disk. If both the spline mesh component and the spline itself are in a scene this would not be needed. But if you put the spine mesh component (maybe along with the generate collision component) into a prefab to re-use the setup the id on the spline mesh component would always be the same. Thus we also need an id for the spline itself to generate a stable unique name.

veksha:
Hi. Got error after changing Middle Part from Pipe to Box.

image


```bat
*** Assertion ***: File: "C:\git\ezEngine\Code\Engine\RendererCore\Meshes\Implementation\MeshBufferResource.cpp", Line: "234", Function: "ezMeshBufferResourceDescriptor::AllocateStreams", Expression: "m_VertexStreamConfig.m_uiTypesMask != 0", Message: "You have to add streams via 'AddStream' before calling this function"
C:\git\ezEngine\Code\Editor\EditorEngineProcess\EngineProcGameApp.cpp(132):'ezEngineProcessGameApplication::EditorAssertHandler'
C:\git\ezEngine\Code\EnginePlugins\InspectorPlugin\InspectorApp.cpp(34):'TelemetryAssertHandler'
C:\git\ezEngine\Code\Engine\RendererCore\Meshes\Implementation\MeshBufferResource.cpp(234):'ezMeshBufferResourceDescriptor::AllocateStreams'
C:\git\ezEngine\Code\Engine\RendererCore\Meshes\Implementation\SplineMeshComponent.cpp(381):'ezSplineMeshComponent::GenerateSplineMeshDesc'
C:\git\ezEngine\Code\EnginePlugins\JoltPlugin\Components\Implementation\JoltGenerateCollisionComponent.cpp(49):'SplineCollisionGenerationTask::Execute'
C:\git\ezEngine\Code\Engine\Foundation\Threading\Implementation\Task.cpp(55):'ezTask::Run'
C:\git\ezEngine\Code\Engine\Foundation\Threading\Implementation\TaskSystemTasks.cpp(141):'ezTaskSystem::ExecuteTask'
C:\git\ezEngine\Code\Engine\Foundation\Threading\Implementation\TaskWorkerThread.cpp(99):'ezTaskWorkerThread::Run'
*** Assertion ***      Expression: "m_VertexStreamConfig.m_uiTypesMask != 0"     Function: "ezMeshBufferResourceDescriptor::AllocateStreams"     File: "C:\git\ezEngine\Code\Engine\RendererCore\Meshes\Implementation\MeshBufferResource.cpp"     Line: 234     Message: "You have to add streams via 'AddStream' before calling this function"
```

C-Core:
Fix: https://github.com/ezEngine/ezEngine/commit/3c4cd614043bedae9060199fb1a6d79f76831462

PR #1772, opened 2026-01-14
Added asset export option to asset browser

jankrassnigg:
Right click assets (or folders) in the asset browser and choose "Export with Dependnecies". Then select a folder. This will copy all selected assets and their dependencies (other assets, input files) to that folder. Excludes all files from the "Base" directory, as those should exist in all projects anyway.

Can be used to easily copy an asset and everything that's needed to make it work, to another project.

image

### DIFF
--- a/Code/Editor/EditorFramework/Actions/AssetActions.h
+++ b/Code/Editor/EditorFramework/Actions/AssetActions.h
@@ -20,6 +20,7 @@ public:
   static ezActionDescriptorHandle s_hCheckFileSystem;
   static ezActionDescriptorHandle s_hWriteDependencyDGML;
   static ezActionDescriptorHandle s_hCopyAssetGuid;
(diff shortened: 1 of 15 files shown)


veksha:
Thank you for this PR.
Tested it with wooden table from Store project.

one file was not copied along:
EzEngine\Store\PolyHaven\round_wooden_table_01_2k\round_wooden_table_01.bin

jankrassnigg:
> Thank you for this PR. Tested it with wooden table from Store project.
> 
> one file was not copied along: EzEngine\Store\PolyHaven\round_wooden_table_01_2k\round_wooden_table_01.bin

When you save a mesh, it now correctly adds a depenency to the GLTF bin file (if necessary). I've resaved all assets in the Store sample, to fix the behavior there. You may need to do the same for your files (there is a "Resave assets in folder" option in the asset browser).

PR #1785, opened 2026-01-20
Update (Templatize) Primative Type Tests to test both float and double versions.

TrevorCash:
Step toward more fully testing the primitive types.  The Mat4Test.cpp only updated as a first step.

At first I tried using multiple files with moving the cpp to a inline header etc the decided using a simple template for the test functions seemed to be the simplest approach.

The FoundationTestPCH.h using declaration are not used for Mat4Test.cpp and the using definitions could potentially could be removed from the file once all other Tests are updated. 

Tests


Code/UnitTests/FoundationTest/Math/Mat4Test.cpp
```diff
@@ -535,394 +539,401 @@ EZ_CREATE_SIMPLE_TEST(Math, Mat4)
...
+        m.m_fElementsCM[i] = ezMath::NaN<Type>();
 
         EZ_TEST_BOOL(m.IsNaN());
       }
     }
   }
 }
+
+
+EZ_CREATE_SIMPLE_TEST(Math, Mat4)
```
jankrassnigg: Could be two tests, Mat4f and Mat4d, so that when something fails, it is clearer right away, which one failed. A lot of math failures are because of precision and epsilon values not being large enough.

jankrassnigg (commented):
This looks like a good solution !

TrevorCash:
Please Re-Review - Basic Primative Tests updated with templatization.  All tests passing on my linux machine.

Code/UnitTests/FoundationTest/Math/PlaneTest.cpp
```diff
@@ -5,185 +5,193 @@
 #include <Foundation/Math/Plane.h>
 #include <Foundation/Math/Random.h>
 
-EZ_CREATE_SIMPLE_TEST(Math, Plane)
+template <typename Type>
+void TestPlane()
```
jankrassnigg, line 9: One problem that I often ran into, was that I needed to adjust the epsilon values. Now we won't run into problems with the double type, so I guess it can stay as it is. But in theory we could pass an expected epsilon value into these functions. You could even pass two or three (small, medium, large epsilon) and pick the smallest one that works per operation. That would allow us to have more control over the checks between float and double.

However, that's just an idea I wanted to share, I don't think this is necessary to do, actually.

Code/UnitTests/FoundationTest/Math/PlaneTest.cpp
```diff
@@ -192,85 +200,85 @@ EZ_CREATE_SIMPLE_TEST(Math, Plane)
...
-    EZ_TEST_BOOL(p.GetObjectPosition(ezBoundingSphereT::MakeFromCenterAndRadius(ezVec3T(5, 2, 3), 3.0f)) == ezPositionOnPlane::Back);
-    EZ_TEST_BOOL(p.GetObjectPosition(ezBoundingSphereT::MakeFromCenterAndRadius(ezVec3T(15, 2, 4.999f), 3.0f)) == ezPositionOnPlane::Front);
-    EZ_TEST_BOOL(p.GetObjectPosition(ezBoundingSphereT::MakeFromCenterAndRadius(ezVec3T(5, 2, 3), 4.999f)) == ezPositionOnPlane::Back);
-    EZ_TEST_BOOL(p.GetObjectPosition(ezBoundingSphereT::MakeFromCenterAndRadius(ezVec3T(8, 2, 3), 3.0f)) == ezPositionOnPlane::Spanning);
-    EZ_TEST_BOOL(p.GetObjectPosition(ezBoundingSphereT::MakeFromCenterAndRadius(ezVec3T(12, 2, 3), 3.0f)) == ezPositionOnPlane::Spanning);
+    EZ_TEST_BOOL(p.GetObjectPosition(ezBoundingSphereType::MakeFromCenterAndRadius(ezVec3Type(15, 2, 3), (Type)3.0)) == ezPositionOnPlane::Front);
+    EZ_TEST_BOOL(p.GetObjectPosition(ezBoundingSphereType::MakeFromCenterAndRadius(ezVec3Type(5, 2, 3), (Type)3.0)) == ezPositionOnPlane::Back);
+    EZ_TEST_BOOL(p.GetObjectPosition(ezBoundingSphereType::MakeFromCenterAndRadius(ezVec3Type(15, 2, 4.999), (Type)3.0)) == ezPositionOnPlane::Front);
```
jankrassnigg: looks like there are a few missing (Type) casts for the double values here

see https://dev.azure.com/ezEngineCI/ezEngine/_build/results?buildId=31908&view=logs&j=b9e62f99-1a98-5ed7-01d2-f4794231ed79&t=63edae08-351e-541d-7ec2-087c0ff5cbc1&l=5778

jankrassnigg (approved):
Very nice, thank you for this improvement :)

Code/UnitTests/FoundationTest/Math/PlaneTest.cpp
TrevorCash: Tightening down the epsilon in the tests will potentially help show potential math bugs (and then loosening back up if it makes sense).    Its good to know the math functions are working relatively precisely.  I replaced the hard coded 0.0**1 in the Mat4Test.cpp to ezMath::***Epsilon<Type>().  I  first used ezMath::SmallEpsilon<Type>() as a excercize to see where the tests started failing and then loosened back up to DefaultEpsilon to make the Mat4f case pass again.  see https://github.com/ezEngine/ezEngine/pull/1785/commits/7b89470cf4ed6b201a108e1c9779ece8dcec0d2b .

The ezMath::***Epsilon<Type>() function for Type = float are the same values.

TrevorCash:
Please Review - Most Primative Tests have been templatized.  some tests were left to use floating point versions of epsilons. The ezAngle class was also converted into a template class to support working with double math from other classes. ezAngle resolves to ezAngleTemplate<float> by default and ezAngled resolves to ezAngleTemplate<double>.  Some Tests like color tests still use ez***T classes and could be templatized.  Frustum and Intersections Tests also could be templatized.

Code/Engine/Foundation/Math/Math.h
```diff
@@ -459,6 +467,8 @@ namespace ezMath
 
 } // namespace ezMath
 
+#include <Foundation/Math/Angle.h>
```
jankrassnigg: That's a really bad cyclic include dependency. Angle.h includes Math.h and Math.h now also includes Angle.h. That shouldn't be.

Angle.h should not include Math.h, but only Declarations.h as before.

Math.h can continue to include Angle.h.

However, the Angle_inl.h, where the actual implementation lives, probably can't be included by Angle.h anymore, but Math.h can include it at the very end.

Angle.h may need to include Math.h at its end (!) to cover for cases where a file only includes Angle.h without Math.h first.

Code/Engine/Foundation/Strings/Implementation/FormatStringArgs.h
```diff
@@ -4,9 +4,10 @@
 #  error "Please don't include FormatStringArgs.h directly, but instead include Foundation/Basics.h"
 #endif
 
+#include <Foundation/Math/Declarations.h>
```
jankrassnigg: This file and its include dependencies was notoriously hard to get right. I would prefer to just copy the ezAngle declaration here.

Code/Engine/Foundation/Math/Implementation/MathFloat_inl.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <algorithm>
+#include <type_traits>
```
jankrassnigg: what is this needed for?

Code/Engine/Foundation/Math/Implementation/Math_inl.h
```diff
@@ -472,11 +472,7 @@ namespace ezMath
   }
 } // namespace ezMath
 
-constexpr EZ_FORCE_INLINE ezAngle ezAngle::AngleBetween(ezAngle a, ezAngle b)
```
jankrassnigg, line 475: This was here because it's an inline function that needs to access ezMath. By putting it here, Angle.h did not need to include Math.h

Code/Engine/Foundation/Math/Implementation/Math.cpp
```diff
@@ -223,26 +223,6 @@ size_t ezMath::SafeConvertToSizeT(ezUInt64 uiValue)
 }
 #endif
 
-void ezAngle::NormalizeRange()
```
jankrassnigg, line 226: Same here, since this function needs Math.h and Angle.h (and is not inline), it was put here.

Code/Engine/Foundation/Math/Implementation/Angle_inl.h
```diff
@@ -1,149 +1,203 @@
...
+  return ezAngleTemplate<Type>::MakeFromRadian(a.GetRadian() / f);
 }
 
-constexpr inline float operator/(const ezAngle& a, const ezAngle& b)
+template <typename Type>
+constexpr inline Type operator/(const ezAngleTemplate<Type>& a, const ezAngleTemplate<Type>& b)
 {
   return a.GetRadian() / b.GetRadian();
 }
+
```
jankrassnigg, line 172: These need to be in Math.h where they were before, to break include dependencies.

jankrassnigg (commented):
Overall pretty good, though I'm not sure about the templatization of ezAngle.

The basic types like vec3, mat4, etc. need to support float and double types to make it possible to use them with very large values. However, angles are always in range 2*pi. There is no benefit from using doubles here. That is the reason why they were never made into a template.

The same is true for things like ezColor and probably a few others. Classes should not be templatized just to make them compatible with classes that use double. Otherwise that would ripple through everything and as you've already noticed, that creates quite some headaches with include dependencies. Also it increases compile times and binary size.

So TBH I'd prefer for ezAngle to stay a pure float implementation unless there is a very strong argument for its need.

TrevorCash:
- For ezAngle Its more of a concern of precision with trigonometry than of value range.  so while ezAngleTemplate<float> and ezAngleTemplate<double> both have a range of 0 to 2PI ,  double has more precision within that range to support basic trigonometry with triangles over large distances.  (think astronomy - using angle numbers to store star locations)

- So over the distance of a large gameword that double would support - double precision angles might be needed when working with long distances.  (IE constructing positions at large distances from angles in trig computations) I can easily imagine making a space game and having an angle as a tragectory to a distance planet)

- It also seems good to have the basic math system all be able to work with doubles and not have casts from double->float->double 

I did some basic comparison of trig distances and the effects of float vs double in angles with a claude chat here: https://claude.ai/share/ff8de938-4358-4a87-8285-ab25aa95b330 

Comparing the precision limits:
Earth orbit (ISS): ±80 meters error (float)
Earth orbit (Moon): micrometers error (double)

I would definitly use ezAngled in game code for math with larger distances

TrevorCash:
> The same is true for things like ezColor and probably a few others. Classes should not be templatized just to make them compatible with classes that use double. Otherwise that would ripple through everything and as you've already noticed, that creates quite some headaches with include dependencies. Also it increases compile times and binary size.

Got It - This makes sense. Its easy to over-do the templatization and I was kindof on a momentum roll.  

See my comments on ezAngle though - because I think precision in a double game world may be a concern.

jankrassnigg:
In practice that is only a concern, if you indeed do planetary simulations. I can't think of any other case where such precision would be needed. And I would suspect that this simulation code is tiny and could just use special purpose code. Once you have the planet positions, the rest of the systems won't work with angles any further.

So while true for that use case, I personally don't see much need for double precision angles. But as long as it doesn't disrupt the codebase, we can support this.

Code/Engine/Foundation/Math/Math.h
TrevorCash: removed in commit 0582e67034b74d403e9afb1c1a5b224b27964fb5

Code/Engine/Foundation/Strings/Implementation/FormatStringArgs.h
TrevorCash: removed and replaced with only ezAngle declarations in commit: 27fedea668090b459ef46104c319a8cb7435de7a

Code/Engine/Foundation/Math/Implementation/MathFloat_inl.h
TrevorCash: removed.  commit d7f584a08df44eec64f7e931b6be5aac67448a29

Code/Engine/Foundation/Math/Implementation/Math_inl.h
TrevorCash: in 8f2a9c738dcf70d1414fd98930ff26b7c404e906 I am getting compiles without including Math.h in Angle.h. (verify but I think this does not need to be added back in?)

Code/Engine/Foundation/Math/Implementation/Math.cpp
TrevorCash: see 8f2a9c738dcf70d1414fd98930ff26b7c404e906 (verify but I think this does not need to be added back in?)

TrevorCash:
Please Re-review the include structure -  I am getting full compiles on linux. 

I think ezAngleTemplate<double> may come in handy from time to time - I might use it it in gameside calculations for other things too.

TrevorCash:
building without errors for me on linux and windows.  Please Re-Review

Code/Engine/Foundation/Math/Math.h
TrevorCash: 7fe5cce9aacdb4cce0cf8a98b5e59facd5ee6fe6

jankrassnigg:
Note that there are various remaining errors.

MathTest.cpp + PlaneTest.cpp

https://dev.azure.com/ezEngineCI/ezEngine/_build/results?buildId=32029&view=logs&j=b9e62f99-1a98-5ed7-01d2-f4794231ed79&t=63edae08-351e-541d-7ec2-087c0ff5cbc1&l=5963

Epsilon in SimdDoubleTest.cpp

https://dev.azure.com/ezEngineCI/ezEngine/_build/results?buildId=32028&view=logs&j=df3f7399-4c57-569d-f4a8-12cad43d96da&t=83d6aeca-49c2-5392-495d-27b71162a4c3&l=927

TrevorCash:
All foundation Math and SIMD tests passing on my local windows.  please re-review

TrevorCash:
Sorry - pushed another commit and stopped the CICD.  please re-review

TrevorCash:
please re-review

TrevorCash:
please re-review

TrevorCash:
changes above for local linux compilation. please re-review

TrevorCash:
Please Re-Review

jankrassnigg:
Very nice, thank you for the hard work on this!

TrevorCash:
Cheers!

PR #1803, opened 2026-01-30
GDB pretty printers

Sanakan8472:
Also some basic copilot instructions for easier automation.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -22,21 +22,21 @@ Always run CMake configure before building after adding new source files.
 ### Linux
 
 ```shell
-cmake --preset linux-gcc-debug && cmake --build Workspace/linux-gcc-debug
+./RunCMake.sh --target linux-gcc-debug --solution-name Copilot --workspace-dir copilot && cmake --build Workspace/copilot
 ```
 
-Build outputs: `Output/Bin/LinuxNinjaGccDebug64/` and `Output/Lib/LinuxNinjaGccDebug64/`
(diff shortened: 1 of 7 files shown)


.github/copilot-instructions.md
```diff
@@ -0,0 +1,85 @@
...
+
+Always run CMake configure before building after adding new source files.
+
+### Linux
+
+```shell
+cmake --preset linux-gcc-debug && cmake --build Workspace/linux-gcc-debug
+```
+
+Build outputs: `Output/Bin/LinuxNinjaGccDebug64/` and `Output/Lib/LinuxNinjaGccDebug64/`
```
jankrassnigg: Make sure that ai builds and the output files go into a different folder than what the regular builds do. Otherwise they overlap and you run into problems when the agent modifies your build. 

I think you've done that partially below, but but for all steps.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,85 @@
...
+
+Available presets: `linux-gcc-debug`, `linux-gcc-dev`, `linux-clang-debug`, `linux-clang-dev`
```
jankrassnigg, line 30: Also it's useful to give the ai output for a clear name (Copilot), so that people understand where it comes from

.github/copilot-instructions.md
Sanakan8472: Done! I took the opportunity to align the linux `RunCMake.sh` with the Windows `RunCMake.ps1` so the command lines are similar.

.github/copilot-instructions.md
Sanakan8472: Changed it to 'copilot' for both Windows and Linux. I let it test its own instructions, seems to work. I assume Windows works too, as I copied those from claude.md.

jankrassnigg (approved):

PR #1806, opened 2026-02-01
VSE: Linearized code generation and type system improvements

Sanakan8472:
## Code Generation
* Fix code explosion from recursive inlining - Replaced recursive inline expansion with linearized code generation using topological sorting. Main node inputs now generate helper functions with local variables instead of inlining expressions repeatedly. Output is now almost readable:
image

## Type System
* Add "auto" type support - Pins can now use Type { "auto" } to dynamically compute their output type dimension from the maximum of their connected inputs. This prevents unnecessary type casts and allows HLSL code to be generated that has the minimum required number of components.
* Input pins also support "auto" type (Add, Subtract, Multiply, Min, Max, Lerp, etc.). This allows anything to be stuck in there. The gui always asumes this is a float for the default value logic because that's what the type was set to previously for all actually 'auto' types.
* Fix MergeFloat2/3/4 output types - Changed output types from float to the correct float2/float3/float4.

## DDL Structure
* Separate vertex/pixel shader sections in DDL - Split CodeVertexShader into CodeVertexDefines, CodeVertexIncludes, and CodeVertexBody for proper handling of vertex shader sections (consistent with pixel shader organization).

### DIFF
--- a/Code/EditorPlugins/Assets/EditorPluginAssets/VisualShader/VsCodeGenerator.cpp
+++ b/Code/EditorPlugins/Assets/EditorPluginAssets/VisualShader/VsCodeGenerator.cpp
@@ -15,13 +15,18 @@ namespace
       return 3;
     if (pType == ezGetStaticRTTI<ezVec4>() || pType == ezGetStaticRTTI<ezColor>())
       return 4;
+    // We treat 'auto' as float as all the default values for auto fields are float.
+    if (pType == nullptr)
+      return 1;
+
+    EZ_REPORT_FAILURE("Unknown RTTI type found in VSE type");
     return 0;
   }
 
   // Returns the HLSL type string for a given dimension
-  const char* DimensionToHlslType(ezUInt8 dimension)
+  const char* DimensionToHlslType(ezUInt8 uiDimension)
   {
-    switch (dimension)
+    switch (uiDimension)
     {
       case 1:
         return "float";
@@ -30,7 +35,9 @@ namespace
       case 3:
         return "float3";
       case 4:
+        return "float4";
       default:
+        EZ_REPORT_FAILURE("Unknown vector dimension found in HLSL type");
         return "float4";
     }
   }
@@ -188,6 +195,7 @@ ezString ezVisualShaderCodeGenerator::GetInputPinDefaultValue(const ezDocumentOb
 
 ezResult ezVisualShaderCodeGenerator::CollectNodesInTopologicalOrder(const ezDocumentObject* pRootNode, ezDynamicArray<const ezDocumentObject*>& out_Sorted) const
 {
+  // We need the reachable nodes to ignore connections that go out of this subtree as they are not relevant.
   ezHashSet<const ezDocumentObject*> reachableNodes;
   CollectReachableNodes(pRootNode, reachableNodes);
 
@@ -198,13 +206,13 @@ ezResult ezVisualShaderCodeGenerator::CollectNodesInTopologicalOrder(const ezDoc
 
   while (!nodeStack.IsEmpty())
   {
-    // Find next node with all execution dependencies visited
-    const ezDocumentObject* pAstNode = nullptr;
+    // Find the next node in which all outgoing connections are visited nodes or nodes outside the reachable nodes.
+    const ezDocumentObject* pNode = nullptr;
     for (ezUInt32 i = nodeStack.GetCount(); i-- > 0;)
     {
-      const ezDocumentObject* pCandidateAstNode = nodeStack[i];
+      const ezDocumentObject* pCandidateNode = nodeStack[i];
       bool bAllVisited = true;
-      ezArrayPtr<const ezUniquePtr<const ezPin>> outputPins = m_pNodeManager->GetOutputPins(pCandidateAstNode);
+      ezArrayPtr<const ezUniquePtr<const ezPin>> outputPins = m_pNodeManager->GetOutputPins(pCandidateNode);
       for (auto& pOutputPin : outputPins)
       {
         ezArrayPtr<const ezConnection* const> connections = m_pNodeManager->GetConnections(*pOutputPin);
@@ -223,22 +231,23 @@ ezResult ezVisualShaderCodeGenerator::CollectNodesInTopologicalOrder(const ezDoc
 
       if (bAllVisited)
       {
-        pAstNode = pCandidateAstNode;
+        pNode = pCandidateNode;
         nodeStack.RemoveAtAndCopy(i);
         break;
       }
     }
 
-    if (pAstNode == nullptr)
+    if (pNode == nullptr)
     {
       EZ_REPORT_FAILURE("Execution connection corrupted or loop detected");
       return EZ_FAILURE;
     }
 
-    EZ_VERIFY(visitedNodes.Insert(pAstNode) == false, "");
-    out_Sorted.PushBack(pAstNode);
+    EZ_VERIFY(visitedNodes.Insert(pNode) == false, "Every node should only be visited once");
+    out_Sorted.PushBack(pNode);
 
-    ezArrayPtr<const ezUniquePtr<const ezPin>> inputPins = m_pNodeManager->GetInputPins(pAstNode);
+    // Add all incoming connections of the node to the nodeStack.
+    ezArrayPtr<const ezUniquePtr<const ezPin>> inputPins = m_pNodeManager->GetInputPins(pNode);
     for (ezUInt32 i = 0; i < inputPins.GetCount(); ++i)
     {
       const ezUniquePtr<const ezPin>& pInputPin = inputPins[i];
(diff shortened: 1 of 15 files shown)


Code/EditorPlugins/Assets/EditorPluginAssets/VisualShader/VsCodeGenerator.cpp
```diff
@@ -100,6 +134,301 @@ void ezVisualShaderCodeGenerator::DetermineConfigFileDependencies(const ezDocume
...
+        pDefinesOut->Append("#endif\n");
+      }
+    }
+    return pinDesc.m_sDefaultValue;
+  }
+}
+
+ezResult ezVisualShaderCodeGenerator::CollectNodesInTopologicalOrder(const ezDocumentObject* pRootNode, ezDynamicArray<const ezDocumentObject*>& out_Sorted) const
+{
+  ezHashSet<const ezDocumentObject*> reachableNodes;
```
C-Core, line 199: Comments and variable names in this function are still visual shader specific. E.g. "astNode" and "execution connection" doesn't really apply to VSE

C-Core (approved):

Code/EditorPlugins/Assets/EditorPluginAssets/VisualShader/VsCodeGenerator.cpp
```diff
@@ -2,76 +2,110 @@
...
+  {
+    if (pType == ezGetStaticRTTI<float>())
+      return 1;
+    if (pType == ezGetStaticRTTI<ezVec2>())
+      return 2;
+    if (pType == ezGetStaticRTTI<ezVec3>())
+      return 3;
+    if (pType == ezGetStaticRTTI<ezVec4>() || pType == ezGetStaticRTTI<ezColor>())
+      return 4;
+    return 0;
```
jankrassnigg, line 23: 0 should be an assert not implemented

Code/EditorPlugins/Assets/EditorPluginAssets/VisualShader/VsCodeGenerator.cpp
```diff
@@ -2,76 +2,110 @@
...
-    case ezVariantType::Vector4:
+    switch (value.GetType())
     {
-      ezVec4 v = value.Get<ezVec4>();
-      temp.SetFormat("float4({0}, {1}, {2}, {3})", v.x, v.y, v.z, v.w);
-    }
-    break;
+      case ezVariantType::String:
+      {
+        temp = value.Get<ezString>();
```
jankrassnigg, line 53: What's the purpose for the string type here?
The only thing I could think of, is for adding a custom HLSL node, or so, but not sure whether that wouldn't require additional support for integration.
Either remove, or add a comment.

Code/EditorPlugins/Assets/EditorPluginAssets/VisualShader/VsCodeGenerator.cpp
```diff
@@ -2,76 +2,110 @@
...
+  // Returns the dimension of a type: 1 for float, 2 for ezVec2, 3 for ezVec3, 4 for ezVec4/ezColor, 0 for unknown
+  ezUInt8 GetTypeDimension(const ezRTTI* pType)
+  {
+    if (pType == ezGetStaticRTTI<float>())
+      return 1;
+    if (pType == ezGetStaticRTTI<ezVec2>())
+      return 2;
+    if (pType == ezGetStaticRTTI<ezVec3>())
+      return 3;
+    if (pType == ezGetStaticRTTI<ezVec4>() || pType == ezGetStaticRTTI<ezColor>())
```
jankrassnigg, line 16: Not sure for which types you need this, but ToShaderString() below also uses ezColorGammaUB, int, bool, time and angle.

Code/EditorPlugins/Assets/EditorPluginAssets/VisualShader/VsCodeGenerator.cpp
```diff
@@ -100,6 +134,301 @@ void ezVisualShaderCodeGenerator::DetermineConfigFileDependencies(const ezDocume
...
+    {
+      const ezUniquePtr<const ezPin>& pInputPin = inputPins[i];
+      ezArrayPtr<const ezConnection* const> connections = m_pNodeManager->GetConnections(*pInputPin);
+      for (const auto* pConnection : connections)
+      {
+        const ezDocumentObject* pParentNode = pConnection->GetSourcePin().GetParent();
+        if (out_Nodes.Contains(pParentNode))
+          continue;
+
+        if (!nodeStack.Contains(pParentNode))
```
jankrassnigg, line 167: Shouldn't this check be redundant, because it's already covered by the out_Nodes.Contains check ?

Code/EditorPlugins/Assets/EditorPluginAssets/VisualShader/VsCodeGenerator.cpp
```diff
@@ -100,6 +134,301 @@ void ezVisualShaderCodeGenerator::DetermineConfigFileDependencies(const ezDocume
...
+    {
+      const ezUniquePtr<const ezPin>& pInputPin = inputPins[i];
+      ezArrayPtr<const ezConnection* const> connections = m_pNodeManager->GetConnections(*pInputPin);
+      for (const auto* pConnection : connections)
+      {
+        const ezDocumentObject* pParentNode = pConnection->GetSourcePin().GetParent();
+        if (visitedNodes.Contains(pParentNode))
+          continue;
+
+        if (nodeStack.Contains(pParentNode) == false)
```
jankrassnigg, line 258: same question here

Code/EditorPlugins/Assets/EditorPluginAssets/VisualShader/VsCodeGenerator.cpp
```diff
@@ -100,6 +134,301 @@ void ezVisualShaderCodeGenerator::DetermineConfigFileDependencies(const ezDocume
...
+          sExpr.ReplaceAll(sPinPlaceholder, pinToVarName[pConnectedPin].GetData());
+        }
+      }
+
+      // Replace property placeholders
+      InsertPropertyValues(pNode, pDesc, sExpr).IgnoreResult();
+
+      // Generate local variable with descriptive name including node type
+      ezStringBuilder sVarName;
+      sVarName.SetFormat("_{0}_{1}_{2}", pDesc->m_sName, pOutPin->GetName(), nodeState.m_uiNodeId);
```
jankrassnigg: I would swap the last two arguments. So that it is for instance "_Texture_12_red", "_Texture_12_green", etc, when you have multiple outputs of the same node in use

jankrassnigg (approved):
Impressive work.

Code/EditorPlugins/Assets/EditorPluginAssets/VisualShader/VsCodeGenerator.cpp
Sanakan8472: Oh I didn't write this. I only moved the function into an anon namespace but github seems to not have an option to ignore whitespace so it's now all jumbled up. I put a breakpoint in there and it is indeed used in the `CheckPropertyValues` function for example. It goes through exposed parameters in the RTTI system as far as I can tell.

Code/EditorPlugins/Assets/EditorPluginAssets/VisualShader/VsCodeGenerator.cpp
Sanakan8472: Well, `out_Nodes` is the set of nodes that have been visited already. `nodeStack` is the set of candidate nodes that we haven't looked at yet but are reachable from one of the nodes in `out_Nodes`. As the graph is a DAG, and this function doesn't traverse in topological order and just greedily eats all edges, we can encounter nodes both in the already reached set as well as in the `nodeStack`.

Code/EditorPlugins/Assets/EditorPluginAssets/VisualShader/VsCodeGenerator.cpp
Sanakan8472: Ok in this case you are absolutely right, we are traversing in topological order, so `visitedNodes` will never contain the source of the connection's node as that would mean all of its outgoing edges are already visited which contradicts the fact that we are currently handling one of these outgoing edges.

 You always need to check `nodeStack` though as you can have for example 4 edges between the same two nodes and you wouldn't want to add the nod 4 times for each edge.

Code/EditorPlugins/Assets/EditorPluginAssets/VisualShader/VsCodeGenerator.cpp
Sanakan8472: Ah yes good point, will switch those two around.

Code/EditorPlugins/Assets/EditorPluginAssets/VisualShader/VsCodeGenerator.cpp
Sanakan8472: Ups, I love copy and paste too much!

Code/EditorPlugins/Assets/EditorPluginAssets/VisualShader/VsCodeGenerator.cpp
Sanakan8472: Good point! Fixed!

Code/EditorPlugins/Assets/EditorPluginAssets/VisualShader/VsCodeGenerator.cpp
Sanakan8472: All of this is just code I moved into an anonymous namespace, I actually don't know what exactly it does.

PR #1812, opened 2026-02-07
Don't block editor when engine is stuck

Sanakan8472:
## Editor Engine Process Communication

* Added `ProcessStuck` and `ProcessUnstuck` events to detect unresponsive engine process.
* Changed frame synchronization to non-blocking: editor no longer waits for engine redraw confirmation.
* Added failed redraw counter with 5-sec threshold before triggering stuck state.

## Engine View Widget

* Added visual "Viewport Stalled" indicator when engine process stops responding.
* Indicator displays warning icon and engine process ID for debugging.
  image

## UI Services / Tick Event System:

* Added `BeforeFrame` tick event type with frame request/cancel mechanism.
* Document windows now always request frames via tick event instead of direct requests.
* Fixed `IsEquivalentQtKey` scan codes for Linux: added +8 offset for X11 (xcb).

## Document Window

* Refactored redraw logic to use new tick event frame request system.
* `TriggerRedraw()` now sets a flag instead of immediately rendering.

### DIFF
--- a/Code/Tools/Libs/GuiFoundation/UIServices/Implementation/UIServices.cpp
+++ b/Code/Tools/Libs/GuiFoundation/UIServices/Implementation/UIServices.cpp
@@ -288,17 +288,27 @@ void ezQtUiServices::TickEventHandler()
 
   EZ_ASSERT_DEV(!m_bIsDrawingATM, "Implementation error");
   ezTime startTime = ezTime::Now();
-
-  m_bIsDrawingATM = true;
   s_LastTickEvent.m_uiFrame++;
   s_LastTickEvent.m_Time = startTime;
-  s_LastTickEvent.m_Type = TickEvent::Type::StartFrame;
-  s_TickEvent.Broadcast(s_LastTickEvent);
 
-  s_LastTickEvent.m_Type = TickEvent::Type::EndFrame;
-  s_TickEvent.Broadcast(s_LastTickEvent);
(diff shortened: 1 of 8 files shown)


Code/Tools/Libs/GuiFoundation/UIServices/Implementation/UIServices.cpp
```diff
@@ -454,63 +465,71 @@ namespace ezQtUtils
 {
   bool IsEquivalentQtKey(const QKeyEvent* e, Qt::Key reference)
   {
+    // X11 (xcb) keycodes are hardware scan codes + 8 offset, Wayland uses raw evdev codes
+#if EZ_ENABLED(EZ_PLATFORM_LINUX)
+    const quint32 SC_OFFSET = (QGuiApplication::platformName() == "xcb") ? 8 : 0;
```
jankrassnigg, line 470: Shouldn't that then also be the case in game? Can you test whether the exported games work on this platform?

Also, do you have a link to some reference that mentions this offset?

Sanakan8472: I found this link : https://unix.stackexchange.com/questions/537982/why-do-evdev-keycodes-and-x11-keycodes-differ-by-8 but nothing in terms of official documentation. The engine works fine, I assume GLFW removes this problem internally.

PR #1815, opened 2026-02-14
Cross-platform tracing system (ETW, LTTNG, Perfetto)

Sanakan8472:
## Tracing System
* Tracing was already supported on Windows / Linux via `ezETW`, but it wasn't a unified API and didn't work on Android.
* Platform backends: ETW TraceLogging on Windows, LTTNG-UST on Linux via the tracing wrapper library as before, Perfetto SDK on Android, no-op stubs everywhere else.
    * Added Perfetto SDK as a third-party dependency (`Code/ThirdParty/perfetto`).
* Added a `EZ_USE_TRACING` platform define to be able to turn all tracing macros into no-ops.
* The log writer from `ezETW` has been moved into `ezTraceWriter` which now works the same on all platforms.
* Tracing must be enabled on a per-dynamic-library / application level via two files to setup the provider.
* Caveats:
    * To use tracing, only include your libraries `TraceProvider.h` in a CPP. Never include `#include <Foundation/Tracing/Tracing.h>` directly outside of the provider setup.
    * All events with different schema (number / type of values attached) must have a different name. Two events with the same name but different schema will corrupt Linux traces. The exception is begin / end macros which must have the same name but will be renamed in the backend if necessary.
    * Never trace from a header or include tracing headers from another header. Each trace macro must exist exactly once or linker errors will occur.

```C++
  // MyLibrary/Tracing/TraceProvider.h
  #pragma once
  #include <Foundation/Tracing/Tracing.h>
  EZ_DECLARE_TRACE_PROVIDER(g_ezTrace_MyLibrary);
  #define EZ_TRACE_PROVIDER g_ezTrace_MyLibrary

  // MyLibrary/Tracing/TraceProvider.cpp
  #include <MyLibrary/MyLibraryPCH.h>
  #include <MyLibrary/Tracing/TraceProvider.h>
  EZ_IMPLEMENT_TRACE_PROVIDER(g_ezTrace_MyLibrary, "ez_MyLibrary");
```

* To emit events, use the macros `EZ_TRACE_EVENT`, `EZ_TRACE_SCOPE`, `EZ_TRACE_SCOPE_BEGIN/END`, `EZ_TRACE_ASYNC_BEGIN/END`, `EZ_TRACE_FLUSH` like this:

```C++
  // Instant event with typed fields
  EZ_TRACE_EVENT("TextureLoaded", ezTraceLevel::Info,
    EZ_TRACE_VALUE("Path", szTexturePath),
    EZ_TRACE_VALUE("Width", uiWidth));

  // Scoped event (RAII, measures duration)
  {
    EZ_TRACE_SCOPE("PhysicsStep", ezTraceLevel::Verbose,
      EZ_TRACE_VALUE("NumBodies", uiBodyCount));
    // ... work ...
  } // end event emitted automatically

  // Manual scope (when RAII doesn't fit)
  EZ_TRACE_SCOPE_BEGIN("LoadPhase", ezTraceLevel::Info);
  // ... work ...
  EZ_TRACE_SCOPE_END("LoadPhase");

  // Async activity correlated across threads by ID which must be unique globally.
  EZ_TRACE_ASYNC_BEGIN("AsyncLoad", uiRequestId, ezTraceLevel::Info,
    EZ_TRACE_VALUE("Resource", szPath));
  // ... on another thread or later ...
  EZ_TRACE_ASYNC_END("AsyncLoad", uiRequestId);
```

# Capturing Traces

* Added `Capture-Trace.ps1`, a cross-platform trace capture script for Windows, Linux, Android (add `-Android` switch). Without parameters it will start a trace, wait for you to press any key and then stop the trace.
* Added `ezTraceProvider.wprp` (EZ WPR profile with process/thread context) and `ez-perfetto.pbtx` (Perfetto config).
* Added LTTNG dev packages (`liblttng-ust-dev`, `lttng-tools`) to `RunCMake.sh` Linux setup.

# Misc
* Rewrote `ezCommandLineUtils::SplitCommandLineString` to properly handle quoted arguments and added unit tests.
* Android test framework now reads command-line arguments from intent extras via `SplitCommandLineString` when starting the main activity.
* Added `AndroidTest.ps1` documentation with parameter help and usage examples as well as the new `-Parameters` parameter that allows passing e,g. `-filter Tracing` to the Android activity.


Code/UnitTests/FoundationTest/Utility/CommandLineUtilsTest.cpp
```diff
@@ -185,4 +185,66 @@ EZ_CREATE_SIMPLE_TEST(Utility, CommandLineUtils)
     EZ_TEST_DOUBLE(CmdLn.GetFloatOption("-Test4", 5.6), -11, 0.0);
     EZ_TEST_DOUBLE(CmdLn.GetFloatOption("-Test5"), 0, 0.0);
   }
+
+  EZ_TEST_BLOCK(ezTestBlock::Enabled, "SplitCommandLineString")
```
jankrassnigg, line 189: shouldnt all of these have already worked before?

jankrassnigg (commented):
This definitely needs high level documentation on our website.

Both, how to set this up in code, how to start / stop traces, links to the necessary tools (WPF, Perfetto) and, of course, some explanation what the difference is to regular performance profiling.

Code/UnitTests/FoundationTest/Utility/CommandLineUtilsTest.cpp
Sanakan8472: No the old code always dropped the last argument as it would exit out of the loop via `while (*currentChar != '\0')` but then not write the last argument to `out_args`. It also made arguments out of redundant spaces between arguments.

Sanakan8472:
> This definitely needs high level documentation on our website.
> 
> Both, how to set this up in code, how to start / stop traces, links to the necessary tools (WPF, Perfetto) and, of course, some explanation what the difference is to regular performance profiling.

Here you go: https://github.com/ezEngine/docs-src/pull/19

PR #1816, opened 2026-02-14
Add SendEventMessage support for AngelScript messages

AlexandrKedrov:
Adds SendEventMessage and PostEventMessage implementation for ezAngelScriptMessage. 

Example of usage. To work properly it requires to enable PassThroughUnhandledEvents in the sender ScriptComponent .
```as
GetOwner().SendEventMessage(msg, GetOwnerComponent());
```
Otherwise, I couldn't make AngelScript components communicate with their parent objects

**Note** I made `ezMsgDeliverAngelScriptMsg` be derived from `ezEventMessage` instead of the original `ezMessage`. I am not sure whether we should make two different message classes for AngelScript or not. If it is required, I think it can be done cleanly via a template class.


Code/EnginePlugins/AngelScriptPlugin/TypeRegistration/AsRegisterTypes_World.cpp
```diff
@@ -177,6 +177,62 @@ void ezGameObject_PostAsMessageRecursive(asIScriptGeneric* pGen)
   pObj->PostMessageRecursive(msg, delay, static_cast<ezObjectMsgQueueType::Enum>(delivery));
 }
 
+void ezGameObject_SendAsEventMessage(asIScriptGeneric* pGen)
```
jankrassnigg, line 180: This looks pretty good, great work !

Could you take a look at adding a unit test?
Tests for AngelScript are in the GameEngineTests project, in AngelScriptTest.cpp.
However, most tests are actually not done in C++ code, but are part of an editor project. You can find the project in "Data\UnitTests\GameEngineTest\AngelScript". They are triggered from C++ in "ezGameEngineTestApplication_AngelScript::SubTestBasisExec" and basically just get a message over and over, until they return that they are done. The rest of the test is handled in AS code.

So extend the project with AS test code, probably just extend the "Messaging" sub test and test there that the event messages arrive as expected.

Otherwise, I think the code can stay as it is. The ezEventMessage just adds some properties that are useful for event messages. Since no user ever sees this message directly, it's not a problem to always have these extra properties.

AlexandrKedrov: I've made a separate test suite "EventMessagingTest" to use it with the different set of objects and account to different message traversal path. 

Also, `git status` yields these changes, but those file seem to be generated at the runtime, so I am not sure should I commit them or not

```git
        modified:   Data/UnitTests/GameEngineTest/AngelScript/Editor/AngelScript/AllDeclarations.txt
        modified:   Data/UnitTests/GameEngineTest/AngelScript/Editor/AngelScript/Enums.txt
        modified:   Data/UnitTests/GameEngineTest/AngelScript/Editor/AngelScript/GlobalFunctions.txt
        modified:   Data/UnitTests/GameEngineTest/AngelScript/Editor/AngelScript/Methods.txt
        modified:   Data/UnitTests/GameEngineTest/AngelScript/Editor/AngelScript/Namespaces.txt
        modified:   Data/UnitTests/GameEngineTest/AngelScript/Editor/AngelScript/NotRegisteredDecls.txt
        modified:   Data/UnitTests/GameEngineTest/AngelScript/Editor/AngelScript/Properties.txt
        modified:   Data/UnitTests/GameEngineTest/AngelScript/Editor/AngelScript/Types.txt
        modified:   Data/UnitTests/GameEngineTest/AngelScript/as.predefined
```

jankrassnigg: You can commit these files, though they are not important.

AlexandrKedrov: Done

jankrassnigg:
Looks good, thanks for the contribution!

PR #1823, opened 2026-02-19
Replace test override folder with file suffix system

jankrassnigg:
The tests now use suffixes for image comparisons to allow to have multiple different images to cover different situations.

For platforms it uses: "" (windows), "linux", "osx", "android"
For GPUs: "nvidia", "amd", "intel", "d3dref", "llvmpipe", "swiftshader"
For APIs: "" (d3d11), "vulkan"

Now when an image comparison is done, it first checks the standard image, e.g. "image.png". If that check fails, it also tries files with different suffixes, e.g. "image-linux.png" or "image-vulkan.png". This will also be combined, e.g. "image-linux-amd.png" or "image-android-swiftshader.png"

If any such image exists and the test succeeds, the image comparison test succeeds. Only if none of the existing files succeed, the whole test fails, with the minimum MSE that was encountered.

@C-Core you could for instance use this now for the RML UI tests, to add comparison images using "image-linux.png" to check in images that use different fonts.


C-Core:
Just curious, why is this better than the override folder?

jankrassnigg:
Because you can set multiple tags, e.g. "linux", "amd", "vulkan" and then have images with names "image-linux.png", "image-linux-amd.png" etc. It automatically tries all variants. So you can combine platform / hardware / api tags.

I'd also add a tag "alt2", "alt3", so that you can have different alternatives (for weird driver issues, that are not specific to one platform or vendor).

PR #1824, opened 2026-02-22
Command history now resets modified state when undoing to a saved state

jankrassnigg:
When you make a modification, the document gets marked 'modified'. Undoing the change would not reset this 'modified' flag. Now it does.


C-Core:
That was always my trick to force save a document :(

jankrassnigg:
> That was always my trick to force save a document :(

Just do "change -> save -> undo -> save" instead of "change -> undo -> save".

PR #1841, opened 2026-03-02
Remove body from add queue if they are destroyed in the same frame

C-Core:


jankrassnigg (approved):
Good that someone finally fixed this 👍

PR #1844, opened 2026-03-04
Add option to disable OS keys

jankrassnigg:
ezPlayer now disables the Windows key, since it uses that for its internal menu. Applications can now toggle themselves whether the Windows key is handled or not.


veksha:
For "Export and run" it works OK now. but for "Play the game" we see OS menu and Scene selector at the same time.
Should be fixed too?

jankrassnigg:
"Export and run" starts the ezPlayer, which disables the OS key. "Play the game" runs in the editor process. I think the editor shouldn't disable the OS key. The ezPlayer acts more like a real game, where it is more typical to do so.

I definitely wanted to add the option to disable the OS keys, but I'm still unsure which behavior the ezPlayer should have. Another option could of course be, to use another key for the map selection. Or to disable the menu entirely when running in the editor.

Take a look at user/jk/os-key-2, I've made a quick implementation of the latter there.

PR #1853, opened 2026-03-08
Added ezArgEnum format string helper for enum/bitflags values

Sanakan8472:
## ezArgEnum
* You can now print enumerations easily like this:
```
// Before:
ezStringBuilder sUsage;
ezReflectionUtils::EnumerationToString(ezGetStaticRTTI<ezTexConvUsage>(), m_Usage.GetValue(), sUsage, ezReflectionUtils::EnumConversionMode::ValueNameOnly);
ezLog::Info("-usage is '{}'", sUsage);

// After:
ezLog::Info("-usage is '{}'", ezArgEnum(m_Usage));
// Fully qualified name (e.g. "MyEnum::Value1" instead of "Value1"):
ezLog::Info("-usage is '{}'", ezArgEnum(m_Usage, true));
```

## Misc
* Moved the `Text` trace value for `EZ_TRACE_EVENT("LogMessage"` to the front as the WPA lttng import support limits the number of arguments so that right now the text is always omitted if you import a ctf trace into WPA.


jankrassnigg (approved):
Very nice.

PR #1860, opened 2026-03-10
Temp allocator

C-Core:
* Added a new stack allocation policy that follows a real stack pattern (out of order deallocation is supported as well). In contrast to the linear allocation policy it can re-use the memory right after it has been deallocated which is ideal for short lived, temporary allocations.
* Added ezTempAllocator which makes use of that new stack allocation policy.
* Added an ezTempArray and an ezTempHybridArray class that use the ezTempAllocator for convience:
    * ezTempArray<T> = ezDynamicArray<T>(ezTempAllocator::Get())
    * ezTempHybridArray<T, Size> = ezHybridArray<T, Size>(ezTempAllocator::Get())
    * Note that the temp array classes are not typed by the allocator but constructed with the temp allocator. This makes them compatible with the default array types. E.g. if a function takes an ezDynamicArray<int> as input you can pass an ezTempArray<int> or an ezTempHybridArray.
* Made use of the temp allocator in other places as well.

### DIFF
--- a/Code/Engine/Foundation/Memory/Policies/AllocPolicyLinear.h
+++ b/Code/Engine/Foundation/Memory/Policies/AllocPolicyLinear.h
@@ -32,6 +32,7 @@ public:
     }
   }
 
+  /// \brief Sets the size of the next bucket to allocate. This can be used to prevent an excessive number of buckets if the required total allocation size is known in advance.
   EZ_FORCE_INLINE void SetNextBucketSize(ezUInt32 uiSize)
   {
     m_uiNextBucketSize = uiSize;
--- a/Code/Engine/Foundation/Memory/Policies/AllocPolicyLinear.h
+++ b/Code/Engine/Foundation/Memory/Policies/AllocPolicyLinear.h
@@ -1,6 +1,6 @@
 #pragma once
 
-#include <Foundation/Containers/HybridArray.h>
+#include <Foundation/Containers/SmallArray.h>
 
 /// \brief This policy implements a linear allocator that can only grow and at some point all allocations gets reset at once.
 ///
@@ -30,6 +30,11 @@ public:
     {
       m_pParent->Deallocate(bucket.GetPtr());
     }
+  }
+
+  EZ_FORCE_INLINE void SetNextBucketSize(ezUInt32 uiSize)
+  {
+    m_uiNextBucketSize = uiSize;
   }
 
   EZ_FORCE_INLINE void* Allocate(size_t uiSize, size_t uiAlign)
@@ -121,5 +126,5 @@ private:
 
   ezUInt8* m_pNextAllocation = nullptr;
 
-  ezHybridArray<ezArrayPtr<ezUInt8>, 4> m_Buckets;
+  ezSmallArray<ezArrayPtr<ezUInt8>, 4> m_Buckets;
 };
--- a/Code/Engine/Foundation/Memory/Policies/AllocPolicyStack.h
+++ b/Code/Engine/Foundation/Memory/Policies/AllocPolicyStack.h
@@ -131,7 +131,7 @@ public:
   EZ_ALWAYS_INLINE ezAllocator* GetParent() const { return m_pParent; }
 
 private:
-  ezUInt32 GetMaxAllocationSize() const { return EZ_BIT(m_uiMaxAllocSizeLog2); }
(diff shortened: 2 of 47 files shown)


Code/Engine/Foundation/Memory/Policies/AllocPolicyLinear.h
```diff
@@ -32,6 +32,11 @@ class ezAllocPolicyLinear
     }
   }
 
+  EZ_FORCE_INLINE void SetNextBucketSize(ezUInt32 uiSize)
```
jankrassnigg, line 36: Add a comment ?

Code/Engine/Foundation/Memory/Policies/AllocPolicyStack.h
```diff
@@ -0,0 +1,173 @@
...
+  EZ_FORCE_INLINE ~ezAllocPolicyStack()
+  {
+    EZ_ASSERT_DEV(m_uiCurrentBucketIndex == 0 && m_uiCurrentBucketOffset == 0 && m_Allocations.IsEmpty(), "There is still something allocated!");
+    for (auto& bucket : m_Buckets)
+    {
+      m_pParent->Deallocate(bucket.GetPtr());
+    }
+  }
+
+  EZ_FORCE_INLINE void SetMaxAllocationSize(ezUInt32 uiSize)
```
jankrassnigg, line 39: add comment, to explain when this is needed

Code/Engine/GameEngine/Animation/Skeletal/Implementation/AnimatedMeshComponent.cpp
```diff
@@ -105,18 +105,19 @@ void ezAnimatedMeshComponent::InitializeAnimationPose()
     const ozz::animation::Skeleton* pOzzSkeleton = &pSkeleton->GetDescriptor().m_Skeleton.GetOzzSkeleton();
     const ezUInt32 uiNumSkeletonJoints = pOzzSkeleton->num_joints();
 
-    ezArrayPtr<ozz::math::Float4x4> pPoseMatrices = EZ_NEW_ARRAY(ezFrameAllocator::GetCurrentAllocator(), ozz::math::Float4x4, uiNumSkeletonJoints);
-    EZ_ASSERT_DEBUG(ezMemoryUtils::IsAligned(pPoseMatrices.GetPtr(), alignof(ozz::math::Float4x4)), "Unaligned cast");
+    ezTempArray<ozz::math::Float4x4> poseMatrices;
```
jankrassnigg, line 108: I'm not sure this is correct.
The previous code used the frame allocator, knowing that the allocation will be alive for the rest of the frame.

The replacement actually deletes the data, when it goes out of scope.
Can you double check, that the data isn't used later anymore? (ie that we store the arrayptr somewhere and update gpu buffers from it after this scope).

Code/Engine/GameEngine/Animation/Skeletal/Implementation/LodAnimatedMeshComponent.cpp
```diff
@@ -474,19 +474,19 @@ void ezLodAnimatedMeshComponent::InitializeAnimationPose()
     const ozz::animation::Skeleton* pOzzSkeleton = &pSkeleton->GetDescriptor().m_Skeleton.GetOzzSkeleton();
     const ezUInt32 uiNumSkeletonJoints = pOzzSkeleton->num_joints();
 
-    ezArrayPtr<ozz::math::Float4x4> pPoseMatrices = EZ_NEW_ARRAY(ezFrameAllocator::GetCurrentAllocator(), ozz::math::Float4x4, uiNumSkeletonJoints);
-
-    EZ_ASSERT_DEBUG(ezMemoryUtils::IsAligned(pPoseMatrices.GetPtr(), alignof(ozz::math::Float4x4)), "Unaligned cast");
+    ezTempArray<ozz::math::Float4x4> poseMatrices;
```
jankrassnigg, line 477: not sure about this one either

jankrassnigg (approved):
Please double check that the animations don't break.

Code/Engine/GameEngine/Animation/Skeletal/Implementation/AnimatedMeshComponent.cpp
C-Core: The stack allocator overwrites the memory with a pattern in debug builds. I couldn't see any issues, also from looking at the code it seems fine.

PR #1861, opened 2026-03-10
Spline editing improvements

jankrassnigg:
* Added `ezStringAlgorithms` to ToolsFoundation. Currently only contains `ezStringAlgorithms::ComputeNameBetween()`. This implements the *Dewey decimal scheme* for building a unique numbered name even when inserting in the middle. TL;DR: Makes spline node names better.

* Spline node insert points are now actually in the middle of a spline segment.

* Clicking a spline insertion point twice without moving the mouse, doesn't insert another node anymore.

* When a spline point gets inserted, it is now immediately selected.

* Pressing `Ctrl+B` for *undo selection* now skips undo steps that failed to do something (due to deleted objects).

* Added some documentation comments to manipulator code.

### DIFF
--- a/Code/Editor/EditorFramework/DocumentWindow/EngineViewWidget.moc.h
+++ b/Code/Editor/EditorFramework/DocumentWindow/EngineViewWidget.moc.h
@@ -90,6 +90,9 @@ public:
   /// \brief Starts a picking operation for the given pixel position in this view. Returns the most recent picking information in the meantime.
   const ezObjectPickingResult& PickObject(ezUInt16 uiScreenPosX, ezUInt16 uiScreenPosY) const;
 
+  /// \brief Clears the last stored picking position. Only needed when it is vital that no stale picking data can be used next time.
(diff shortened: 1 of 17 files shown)


C-Core:
The auto selection of inserted spline nodes could be annoying in the following situations:
- The spline has no nodes so far. We need at least two nodes to produce something useful
- When the user wants to append a couple of spline nodes at the end or beginning at once

jankrassnigg:
> The auto selection of inserted spline nodes could be annoying in the following situations:
> 
>     * The spline has no nodes so far. We need at least two nodes to produce something useful
> 
>     * When the user wants to append a couple of spline nodes at the end or beginning at once

True, but those cases aren't great either way. Appending multiple nodes currently doesn't work well, because the distance is getting shorter each time, so you need to reposition the nodes right away. 

We could not auto select, if this was the first node.

There is much more work needed to make spline editing comfortable, though.

PR #1863, opened 2026-03-11
Manipulator improvements

jankrassnigg:
* Spline manipulator is now available even when a node is selected. Makes spline editing much more comfortable.
* You can now cycle through available manipulators on the selected objects with Q, rather than just toggle it on and off.
* Properties with manipulators now also display a toggle button that makes it easier to notice.
* Game objects now always have an icon in the tree view
* Allow to set negative fog distances

### DIFF
--- a/Code/Tools/Libs/GuiFoundation/PropertyGrid/Implementation/ManipulatorManager.cpp
+++ b/Code/Tools/Libs/GuiFoundation/PropertyGrid/Implementation/ManipulatorManager.cpp
@@ -42,8 +42,7 @@ ezManipulatorManager::~ezManipulatorManager()
   ezDocumentManager::s_Events.RemoveEventHandler(ezMakeDelegate(&ezManipulatorManager::DocumentManagerEventHandler, this));
 }
 
-const ezManipulatorAttribute* ezManipulatorManager::GetActiveManipulator(
-  const ezDocument* pDoc, const ezHybridArray<ezPropertySelection, 8>*& out_pSelection) const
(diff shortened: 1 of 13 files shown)


Code/Tools/Libs/GuiFoundation/PropertyGrid/Implementation/ManipulatorManager.cpp
```diff
@@ -143,6 +143,149 @@ void ezManipulatorManager::ToggleHideActiveManipulator(const ezDocument* pDoc)
...
+  if (pCurrentObject == nullptr)
+    return;
+
+  EZ_ASSERT_DEV(pDoc->GetManipulatorSearchStrategy() != ezManipulatorSearchStrategy::None,
+    "The document type '{}' has to override the function 'GetManipulatorSearchStrategy()'", pDoc->GetDynamicRTTI()->GetTypeName());
+
+  // Collect available manipulator attributes from the last selected object (or its children).
+  // Deduplicates by type + primary property so that e.g. multiple ezSplineNodeComponent children
+  // that share the same ezSplineManipulatorAttribute only contribute one entry.
+  ezHybridArray<const ezManipulatorAttribute*, 8> available;
```
C-Core: ezTempHybridArray, also below

jankrassnigg: Yeah, yeah, it's what the cool kids use these days.

PR #1864, opened 2026-03-11
Array usage cleanup

C-Core:
* Replaced ezHybridArray out parameter with ezDynamicArray and const ezHybridArray in parameter with ezArrayPtr. 
* Made use of ezTempHybridArray instead of ezHybridArray where applicable.


jankrassnigg:
Woah, I'm not going to review all that, summarize it in one word!

jankrassnigg (approved):
So, any measurable performance difference ?

C-Core:
It saves us a few allocations in visual script for example. But I couldn't find a use case where there was a clear performance improvement.

jankrassnigg:
> It saves us a few allocations in visual script for example. But I couldn't find a use case where there was a clear performance improvement.

I'm not surprised about that. Since the amount of allocations in ez is already so low, that even in a debug build it's not measurable, I would have been surprised if this made a big difference.

But I guess it can reduce performance spikes in rare cases. Definitely useful nontheless.

PR #1874, opened 2026-03-22
Screenspace effects (SSR, SSS)

wdstudiosma:
Adds Screen Space Reflections (Custom, Temporal) & Screen Space Shadows (Spartan Engine: https://github.com/PanosK92/SpartanEngine)

This PR is listed as a Draft to fix Performance Issues.

### DIFF
--- a/Code/Engine/RendererCore/Lights/Implementation/SSRDataProvider.cpp
+++ b/Code/Engine/RendererCore/Lights/Implementation/SSRDataProvider.cpp
@@ -1,0 +1,35 @@
+#include <RendererCore/RendererCorePCH.h>
+
+#include <RendererCore/Lights/SSRDataProvider.h>
+#include <RendererCore/RenderContext/RenderContext.h>
+#include <RendererFoundation/Device/Device.h>
+
+void ezSSRData::BindResources(ezRenderContext* pRenderContext)
+{
+  if (!m_hSSRTexture.IsInvalidated() && ezGALDevice::GetDefaultDevice()->GetTexture(m_hSSRTexture) != nullptr)
+  {
+    ezBindGroupBuilder& bindGroup = pRenderContext->GetBindGroup(EZ_GAL_BIND_GROUP_RENDER_PASS);
+    bindGroup.BindTexture("SSRTexture", m_hSSRTexture);
+  }
+}
+
+// clang-format off
+EZ_BEGIN_DYNAMIC_REFLECTED_TYPE(ezSSRDataProvider, 1, ezRTTIDefaultAllocator<ezSSRDataProvider>)
+EZ_END_DYNAMIC_REFLECTED_TYPE;
+// clang-format on
+
+ezSSRDataProvider::ezSSRDataProvider() = default;
+
+ezSSRDataProvider::~ezSSRDataProvider() = default;
+
+void ezSSRDataProvider::SetSSRTexture(ezGALTextureHandle hTexture)
+{
+  m_Data.m_hSSRTexture = hTexture;
+}
+
+void* ezSSRDataProvider::UpdateData(const ezRenderViewContext& renderViewContext, const ezExtractedRenderData& extractedData)
+{
+  return &m_Data;
+}
+
+EZ_STATICLINK_FILE(RendererCore, RendererCore_Lights_Implementation_SSRDataProvider);
--- a/Code/Engine/RendererCore/Lights/Implementation/ScreenSpaceShadowDataProvider.cpp
+++ b/Code/Engine/RendererCore/Lights/Implementation/ScreenSpaceShadowDataProvider.cpp
@@ -1,0 +1,35 @@
+#include <RendererCore/RendererCorePCH.h>
+
+#include <RendererCore/Lights/ScreenSpaceShadowDataProvider.h>
+#include <RendererCore/RenderContext/RenderContext.h>
+#include <RendererFoundation/Device/Device.h>
+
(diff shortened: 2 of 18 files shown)


C-Core:
Thanks for the contribution!
I only quickly looked through the code:
* SSS: The implementation looks very bruteforce but might work as a first step. I will measure the perf impact. This feature still requires some work to integrate it properly but should be doable.
* SSR: The approach looks interesting but I don't know how we would integrate that since EZ uses a forward renderer and we don't want to switch to deferred. I know there is this semi deferred approach that DOOM 2016 uses but I don't like the overhead only for SSR. Will have to think about it a bit more and do some more research.

wdstudiosma:
> Thanks for the contribution!
> 
> I only quickly looked through the code:
> 
> * SSS: The implementation looks very bruteforce but might work as a first step. I will measure the perf impact. This feature still requires some work to integrate it properly but should be doable.
> 
> * SSR: The approach looks interesting but I don't know how we would integrate that since EZ uses a forward renderer and we don't want to switch to deferred. I know there is this semi deferred approach that DOOM 2016 uses but I don't like the overhead only for SSR. Will have to think about it a bit more and do some more research.

SSS is very brute force, and will needs some work tailoring it to EZ, which was the main reason for leaving the PR as a draft. SSR was a back port from our fork, we switched over to clustered deferred, so that's why you may see GBuffer references in the passes. They still work natively for ezEngines Forward+ pipeline.

Currently performance for them can be improved, so I'll keep pushing changes to help in that regard.

Sanakan8472:
Very exciting!
I know it's just a draft but I tried running it nevertheless. A few things I noticed:
1. Include not found: `Error: File 'Shaders/Pipeline/SSRClassify.cs', Line 12 (0): Invalid #include 'Shaders/Common/GBufferEncoding.h'`. I think this is the only blocker to be able to play around with it. Not sure what's in it.
2. `m_vBrightestDirectionalLightDirection` undefined.
3. `pConstants->ScreenSpaceShadowEnabled` undefined.
4. `renderViewContext.m_pRenderContext->GetCommandEncoder()->UpdateBuffer(m_hIndirectArgsBuffer, 0, ezMakeArrayPtr(reinterpret_cast<const ezUInt8*>(clearData), sizeof(clearData)), ezGALUpdateMode::AheadOfTime);`
This is missing the `ezGALUpdateMode::AheadOfTime` at the end. Maybe we don't assert on this in DX11 as it's all the same there?

Regarding the gbuffer thing: As diffuse and specular lighting are purely additive, we could make the opaque renderer skip computing specular and instead output a normal+roughness texture which I assume is basically what the gbuffer is in SSR? Then SSR could run on that and add the result on top of the opaque output image with fallback to computing the specular term skipped in the opaque pass?

wdstudiosma:
> Very exciting! I know it's just a draft but I tried running it nevertheless. A few things I noticed:
> 
> 1. Include not found: `Error: File 'Shaders/Pipeline/SSRClassify.cs', Line 12 (0): Invalid #include 'Shaders/Common/GBufferEncoding.h'`. I think this is the only blocker to be able to play around with it. Not sure what's in it.
> 2. `m_vBrightestDirectionalLightDirection` undefined.
> 3. `pConstants->ScreenSpaceShadowEnabled` undefined.
> 4. `renderViewContext.m_pRenderContext->GetCommandEncoder()->UpdateBuffer(m_hIndirectArgsBuffer, 0, ezMakeArrayPtr(reinterpret_cast<const ezUInt8*>(clearData), sizeof(clearData)), ezGALUpdateMode::AheadOfTime);`
>    This is missing the `ezGALUpdateMode::AheadOfTime` at the end. Maybe we don't assert on this in DX11 as it's all the same there?
> 
> Regarding the gbuffer thing: As diffuse and specular lighting are purely additive, we could make the opaque renderer skip computing specular and instead output a normal+roughness texture which I assume is basically what the gbuffer is in SSR? Then SSR could run on that and add the result on top of the opaque output image with fallback to computing the specular term skipped in the opaque pass?



> Very exciting! I know it's just a draft but I tried running it nevertheless. A few things I noticed:
> 
> 1. Include not found: `Error: File 'Shaders/Pipeline/SSRClassify.cs', Line 12 (0): Invalid #include 'Shaders/Common/GBufferEncoding.h'`. I think this is the only blocker to be able to play around with it. Not sure what's in it.
> 2. `m_vBrightestDirectionalLightDirection` undefined.
> 3. `pConstants->ScreenSpaceShadowEnabled` undefined.
> 4. `renderViewContext.m_pRenderContext->GetCommandEncoder()->UpdateBuffer(m_hIndirectArgsBuffer, 0, ezMakeArrayPtr(reinterpret_cast<const ezUInt8*>(clearData), sizeof(clearData)), ezGALUpdateMode::AheadOfTime);`
>    This is missing the `ezGALUpdateMode::AheadOfTime` at the end. Maybe we don't assert on this in DX11 as it's all the same there?
> 
> Regarding the gbuffer thing: As diffuse and specular lighting are purely additive, we could make the opaque renderer skip computing specular and instead output a normal+roughness texture which I assume is basically what the gbuffer is in SSR? Then SSR could run on that and add the result on top of the opaque output image with fallback to computing the specular term skipped in the opaque pass?

From my local clone of the dev branch of ezEngine, SSR will work normally without any GBuffer0 Supplied within the pass, but that is a good idea. ill play around with it, and push it to the PR if it works out. 

Regarding those errors, let me fix them right quick.

wdstudiosma:
Closing PR due to Bend Studio SSS has been added & fixed by @C-Core ,SSR will be readded in another PR with the performance improvements.

PR #1881, opened 2026-03-26
Added Expression Particle Behavior

jankrassnigg:
Added a new particle behavior to evaluate expressions.

Reads from up to 4 particle attributes and forwards to an expression. The expression result can then be used to either write another attribute directly, lookup a value from a curve, or discard the particle, if it fails a threshold test.

This allows users to do more complex behavior, where color, size or speed may change in complex ways.

Using the curve lookup and custom curves you can do complex logic:
* A step in a curve (suddenly dropping from 1 to 0 for instance) is like an "if" (threshold). E.g. "if speed below X, set alpha to 0.5"
* A staircase curve acts like a switch statement (map different ranges to different values)
* The expression VM supports lerp, min, max, saturate, so can do various forms of remapping values.
* Using some creative math (min, max, multiply, abs) of input values, one can create logic combinations to decide when to discard a particle.

Original code and idea by WD Studios.


Code/EnginePlugins/ParticlePlugin/Behavior/ParticleAttributes.h
```diff
@@ -0,0 +1,103 @@
...
+
+/// Particle attribute that can be read by behaviors.
+struct EZ_PARTICLEPLUGIN_DLL ezParticleAttributeRead
+{
+  using StorageType = ezUInt8;
+
+  enum Enum
+  {
+    Age,
+    PositionX,
```
C-Core, line 17: The expression VM can take a vec3 etc as input so it would make more sense to have just Position, Velocity etc. as attributes.

Code/EnginePlugins/ParticlePlugin/Behavior/ParticleAttributes.h
```diff
@@ -0,0 +1,103 @@
...
+/// Particle attribute that can be written by behaviors.
+struct EZ_PARTICLEPLUGIN_DLL ezParticleAttributeWrite
+{
+  using StorageType = ezUInt8;
+
+  enum Enum
+  {
+    // note that modifying position or velocity really only makes sense when simulating in local space
+    // as otherwise these values are in global coordinates
+    PositionX,
```
C-Core, line 46: Same here as above

Code/EnginePlugins/ParticlePlugin/Behavior/ParticleBehavior_Expression.cpp
```diff
@@ -0,0 +1,288 @@
...
+#include <ParticlePlugin/System/ParticleSystemInstance.h>
+
+// clang-format off
+EZ_BEGIN_DYNAMIC_REFLECTED_TYPE(ezParticleBehaviorFactory_Expression, 1, ezRTTIDefaultAllocator<ezParticleBehaviorFactory_Expression>)
+{
+  EZ_BEGIN_PROPERTIES
+  {
+    EZ_ENUM_MEMBER_PROPERTY("Mode", ezParticleExpressionMode, m_Mode),
+    EZ_MEMBER_PROPERTY("Expression", m_sExpression),
+    EZ_ENUM_MEMBER_PROPERTY("InputA", ezParticleAttributeRead, m_InputA),
```
C-Core, line 17: Why only 4 inputs and 1 output? Expressions can have arbitrary number of inputs and outputs.

Code/EnginePlugins/ParticlePlugin/Behavior/ParticleBehavior_Expression.cpp
```diff
@@ -0,0 +1,288 @@
...
+#include <ParticlePlugin/Finalizer/ParticleFinalizer_ApplyVelocity.h>
+#include <ParticlePlugin/System/ParticleSystemInstance.h>
+
+// clang-format off
+EZ_BEGIN_DYNAMIC_REFLECTED_TYPE(ezParticleBehaviorFactory_Expression, 1, ezRTTIDefaultAllocator<ezParticleBehaviorFactory_Expression>)
+{
+  EZ_BEGIN_PROPERTIES
+  {
+    EZ_ENUM_MEMBER_PROPERTY("Mode", ezParticleExpressionMode, m_Mode),
+    EZ_MEMBER_PROPERTY("Expression", m_sExpression),
```
C-Core, line 16: We have an expression edit widget with syntax highlighting. See visual script expression node.

Code/EnginePlugins/ParticlePlugin/Behavior/ParticleBehavior_Expression.cpp
```diff
@@ -0,0 +1,288 @@
...
+  m_bBytecodeValid = false;
+  m_sCompiledExpression = m_sExpression;
+
+  if (m_sExpression.IsEmpty())
+    return;
+
+  ezExpressionParser parser;
+
+  ezExpression::StreamDesc inputs[4];
+  inputs[0] = ezExpression::StreamDesc({s_InputA, ezProcessingStream::DataType::Float});
```
C-Core, line 150: It is not very convinient to only have a, b, c, d as input. They should rather be named like the attributes. Since can have attributes as input and output maybe something like inPosition, outPosition etc.

Code/EnginePlugins/ParticlePlugin/Behavior/ParticleBehavior_Expression.cpp
```diff
@@ -0,0 +1,288 @@
...
+static constexpr ezUInt32 s_uiBatchSize = 128;
+
+void ezParticleBehavior_Expression::UpdateElements(ezUInt64 uiStartIndex, ezUInt64 uiNumElements)
+{
+  if (m_pByteCode == nullptr || uiNumElements == 0)
+    return;
+
+  EZ_PROFILE_SCOPE("PFX: Expression");
+
+  float inputA[s_uiBatchSize], inputB[s_uiBatchSize], inputC[s_uiBatchSize], inputD[s_uiBatchSize];
```
C-Core, line 222: It makes no sense to copy all the attributes from and to the streams. The expression VM takes ezProcessingStream as input and output which is exactly what the particle systems use.

Code/EnginePlugins/ParticlePlugin/Behavior/ParticleBehavior_Expression.cpp
```diff
@@ -0,0 +1,288 @@
...
+    }
+
+    if (m_VM.Execute(*m_pByteCode, vmInputs, vmOutputs, uiBatchCount).Failed())
+    {
+      ezLog::Warning("Particle Expression: Failed to execute expression");
+      m_pByteCode = nullptr;
+      return;
+    }
+
+    if (m_Mode == ezParticleExpressionMode::Discard)
```
C-Core, line 257: I think this can be integrated much nicer. We can add a special discard function to the expression and another function that samples a curve. That would be much more powerful.

jankrassnigg: That all sounds really good. I wasn't aware that the VM is this powerful by now, but it does make sense. Will make another round of improvements.

PR #1882, opened 2026-03-27
Added Detour Crowd agent component

x0r04rg:
Just the bare minimum to make it work. I may add more features in future if anyone needs them.


jankrassnigg:
I can't get this component to do anything. Can you add a sample scene, where some agents randomly walk between points, or something like that? I tried, but it's not working for me.

x0r04rg:
Yep, had a bug in there. I fixed it and added a demo scene.

jankrassnigg:
Fixed some details, please review. Overall it looks good. I don't really have the time to test this further, right now, but I'm willing to just merge it and thus give people two options to choose from.

x0r04rg:
Looks good to me. The m_uiAllowPartialPathBit just sets the m_uiSteeringFailedBit to 1 in case the pathfinding fails, right? The thing is - Detour agents are allowed to have partial paths by default, so if we want to track and prevent that, we have to check the value of dtAgent->partial. I'm not 100% sure though, it needs further testing. But anyway, it should work as is for now.

PR #1884, opened 2026-03-29
Added Expression Particle Behavior

jankrassnigg:
Added a new particle behavior that evaluates an expressions and then writes directly to particle properties.

This allows to read from:
* inPos (vec3 position)
* inVel (vec4 velocity, with xyz == direction, w == speed)
* inColor (vec4 color and transparency)
* inLifeTime (vec2, remaining lifetime and inverse of max lifetime)
* inSize (float)
* inRotSpeed (float)

The expression can then do math operations and write to the same streams by using the names "outXYZ" (e.g. "outPos"). Except LifeTime, that one is read-only. Additionally, there is "outDiscard" a boolean stream. If set to true, the particle is discarded.

The expression may also read global variables. Currently "timeDiff" (float) is provided, as well as all the particle effect input variables (user provided floats and colors).

On top of that, the expression behavior can define input curves, which the expression may sample using "sampleCurve(index, at)".

Original code and idea by WD Studios.

---------

Co-authored-by: wdstudiosma <mikaelamanfo@wdstudios.tech>


Code/EnginePlugins/ParticlePlugin/Behavior/ParticleBehavior_Expression.cpp
```diff
@@ -0,0 +1,633 @@
...
+EZ_END_DYNAMIC_REFLECTED_TYPE;
+
+EZ_BEGIN_DYNAMIC_REFLECTED_TYPE(ezParticleBehavior_Expression, 1, ezRTTIDefaultAllocator<ezParticleBehavior_Expression>)
+EZ_END_DYNAMIC_REFLECTED_TYPE;
+// clang-format on
+
+//////////////////////////////////////////////////////////////////////////
+// sampleCurve(index, at) expression function
+
+static ezHashedString s_sParticleCurves = ezMakeHashedString("ParticleCurves");
```
C-Core, line 53: We have basically the same code in the proc gen plugin. Should we move that to a common place?

C-Core (approved):

Code/EnginePlugins/ParticlePlugin/Behavior/ParticleBehavior_Expression.cpp
jankrassnigg: If you have a good place where to put it, sure. I don't think it's important atm, though.

PR #1893, opened 2026-04-04
LTC Area Lights

PlasmaDev5:
## Overview
Add LTC area lights based on [https://eheitzresearch.wordpress.com/415-2/](https://eheitzresearch.wordpress.com/415-2/) 
Currently this change includes Disk and Rectangle Lights. For now they are setup to add simple spotlight style shadowing. This is because correct area light shadowing is much more complicated and often depends on RT.

## Missing changes
- Icons for editor and scene
- Manipulators and debug selections

## Notes
Because the light texture comes from the light cookie atlas i had to get a little creative as from my understanding normally it makes use of sampler settings to get the correct behavior at the edge of lights.
Also added light falloff and two sided to the light data. The falloff is useful for controlling how soft the area light appears. 
Two sided on the other hand allows for the area light to display on both sides of the plane (only works when shadows disabled)

### Rectangle Lights:
Screenshot 2026-04-04 010120
Screenshot 2026-04-04 010101

### Disk Lights:
Screenshot 2026-04-04 010412


C-Core:
This PR would still require quite a bit of work and cleanup to properly integrate. We might want to start with simpler area lights first and adding textures later. However there are currently other more important things on the agenda like e.g. tonemapping refactor, GI etc. so it might take a while until we come back to the topic of area lights.

jankrassnigg:
I'm closing this PR for now. We'll come back to this later, but it's not clear yet when we'll have the time for it.

PR #1900, opened 2026-04-09
Integrated Screen Space Shadows by Bend Studio

C-Core:
* Integrated Screen Space Shadows by Bend Studio as described here: https://www.bendstudio.com/blog/inside-bend-screen-space-shadows.

SSS

The shadow distance is limited in screenspace to keep a stable performance. In the future we might want to run SSS in halfres to further improve performance and increase shadow distance.

* Also implemented proper light sorting by screen space size and other criteria.


jankrassnigg:
Really nice visual improvement.

SSS off:

SSS-off

SSS on:

SSS-on

PR #1909, opened 2026-04-20
Debian Trixie (13) support

tccountus:
You will also have to merge https://github.com/ezEngine/thirdparty-assimp/pull/5
This PR is tested on debian trixie, i'm not sure if we have to edit the regex for debian testing or debian sid


jankrassnigg:
Since assimp is in a submodule, please add a commit that updates it's reference. I merged the change in it already.

tccountus:
Should be done

jankrassnigg:
Thanks for the contribution!

PR #1910, opened 2026-04-21
Added "Normal Weight" parameter to "Simplify Mesh" feature; ...

veksha:
- Added "Normal Weight" parameter to "Simplify Mesh" feature;
- updated meshoptimizer from 1.0 to 1.1

https://github.com/user-attachments/assets/466f9d8e-28b9-4f10-9ff0-fa366aa1393b


Code/Tools/Libs/ModelImporter2/ImporterAssimp/AssimpMeshImport.cpp
```diff
@@ -51,7 +51,13 @@ namespace ezModelImporter2
     }
     else
     {
-      numNewIndices = meshopt_simplify(simplifiedIndices.GetData(), indices.GetData(), numIndices, &pMesh->mVertices[0].x, pMesh->mNumVertices, sizeof(aiVector3D), numTargetIndices, fTargetError, 0, &err);
+      const float nrm_weight = m_Options.m_fNormalWeight;
+      const float attr_weights[3] = {nrm_weight, nrm_weight, nrm_weight};
+
```
jankrassnigg: I am not sure whether we could have meshes here where mNormals is nullptr (meshes that were exported without normals).

Maybe add a null check and fall back to the old behavior in that case ?

jankrassnigg (approved):
Awesome! Thanks for the very thorough code change (you even updated the cgmanifest :) )

See one small comment below.

Code/Tools/Libs/ModelImporter2/ImporterAssimp/AssimpMeshImport.cpp
veksha: added check, please see. tested with simple cube exported from blender without normals.

veksha:
> Awesome! Thanks for the very thorough code change (you even updated the cgmanifest :) )

no problem! BTW, what is cgmanifest? is it important to update it?

jankrassnigg:
> > Awesome! Thanks for the very thorough code change (you even updated the cgmanifest :) )
> 
> no problem! BTW, what is cgmanifest? is it important to update it?

It's not really important. It is a file where you track dependencies, such that automated tools can know which ones you have and at what version. In some companies this is used to scan for vulnerabilities or legal requirements.
We don't use it in any automated way, but are just used to keeping them up-to-date as well.

C-Core:
> > > Awesome! Thanks for the very thorough code change (you even updated the cgmanifest :) )
> > 
> > 
> > no problem! BTW, what is cgmanifest? is it important to update it?
> 
> It's not really important. It is a file where you track dependencies, such that automated tools can know which ones you have and at what version. In some companies this is used to scan for vulnerabilities or legal requirements. We don't use it in any automated way, but are just used to keeping them up-to-date as well.

There are useful for humans as well. That way there is a unified place to quickly check which version of a library you have.

PR #1913, opened 2026-04-23
PostMessage fixes and cleanups + Lighting fixes

C-Core:
* To prevent infinite post message loops we used a workaround that we added a small delay to all message of the same queue while we were processing this loop. To clean this up we now use the common "swap queue" before processing technique. This has the additional benefit that we now need less locks while processing.
* Timed messages are now also processed in the Editor even when the simulation is not running. The behavior before led to constant confusion why certain messages were not delivered in the Editor.

* Allow custom material shaders to skip all screen space techniques like e.g. SSAO and SSShadows.
* Fixed point lights which would wrongly sample light cookies since their cookieParams0 could be != 0 because we store the right dir for tube lights in there now. The shader now only checks the lower 16 bit which contain the cookie index.
* Particle lighting now also applies light cookies.

### DIFF
--- a/Data/Base/Shaders/Materials/MaterialPixelShader.h
+++ b/Data/Base/Shaders/Materials/MaterialPixelShader.h
@@ -7,6 +7,10 @@
 #define USE_WORLDPOS
 #define USE_DATAOFFSETS
 
+#if !defined(NO_SCREEN_SPACE_TECHNIQUES)
+#  define USE_SCREEN_SPACE_TECHNIQUES
+#endif
+
 #if (BLEND_MODE == BLEND_MODE_MASKED || BLEND_MODE == BLEND_MODE_DITHERED) && RENDER_PASS != RENDER_PASS_WIREFRAME
 
 // No need to do alpha test again if we have a depth prepass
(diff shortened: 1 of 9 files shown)


Data/Base/Shaders/Materials/MaterialPixelShader.h
```diff
@@ -111,13 +115,13 @@ PS_OUT main(PS_IN Input)
...
-  bool applySSAO = true;
+#    if defined(USE_SCREEN_SPACE_TECHNIQUES) && (BLEND_MODE == BLEND_MODE_OPAQUE || BLEND_MODE == BLEND_MODE_MASKED || BLEND_MODE == BLEND_MODE_DITHERED)
+  bool useScreenSpaceTechniques = true;
 #    else
-  bool applySSAO = false;
+  bool useScreenSpaceTechniques = false;
 #    endif
 
-  AccumulatedLight light = CalculateLighting(matData, clusterData, Input.Position.xyw, applySSAO);
+  AccumulatedLight light = CalculateLighting(matData, clusterData, Input.Position.xyw, useScreenSpaceTechniques);
```
jankrassnigg, line 124: I'm confused, ssao is a post processor, how can the material shader, which is run in the forward rendering pass, deactivate it? And what's the typical use case?

Data/Base/Shaders/Particles/ParticleCommonVS.h
```diff
@@ -211,6 +211,11 @@ float3 CalculateParticleLighting(float4 screenPosition, float3 worldPosition, fl
         attenuation *= lightData.intensity;
         float3 lightColor = RGB8ToFloat3(lightData.colorAndType);
 
+        if ((lightData.cookieParams0 & 0xFFFF) != 0)
```
jankrassnigg, line 214: And how can this be used? Iirc the particle light type doesn't expose cookies.

C-Core: This about lighting on particles NOT the particle light type. E.g. if you had a white spot light but with a red cookie texture the lighting on particles was still white because it was not sampling the cookie texture.

Data/Base/Shaders/Materials/MaterialPixelShader.h
C-Core: The forward rendering passes sample the result of SSAO or SS Shadows. With that flag it is possible to skip the sampling/application of SSAO etc.
My usecase: I want my grass to not use SSAO or SS Shadows because it looks weird. Now I forced my grass to be rendered in the transparent pass with the "LitTransparent" category but I still use the masked blending type. Usually we apply SSAO etc depending on the blending type, now there is an option for material shaders to disable the sampling of SSAO etc.

jankrassnigg: So the rendering issue on cloth, for instance, could be fixed by using a material that disabled ssao with a dedicated cloth material? 
At least for now.

C-Core: Yes, could be used for that as well

veksha:
Hi @C-Core . after this PR i get crash in my project and MonsterAttack project too.

MonsterAttackPlugin\GameState\MonsterAttackGameState.cpp(51):'MonsterAttackGameState::BeforeWorldUpdate'

line is: `EZ_LOCK(m_pMainWorld->GetWriteMarker());`


(Dev build)

C-Core:
@veksha you need to compile your custom code plugin again since the world data layout and interface has changed. That worked for me.

veksha:
yes, works! can't believe i didn't try that right away. 😄

PR #1916, opened 2026-04-24
Debug and gizmo rendering improvements

jankrassnigg:
ezDebugRenderer geometry is now rendered dimmed when occluded. Allows to see debug geometry through walls while keeping some perception of occlusion. Not sure yet, whether this should be enabled on everything by default, but difficult to implement differently.

Editor gizmos can now render themselves with arbitrary lines. Uses the debug renderer, so lines will show through walls (now). Allows for more dynamic gizmos.

Added an ezRoundedRectVisualizerAttribute, which visualizes a rounded rectangle around the object.

Added ezShapeIconAlwaysVisibleAttribute. If used, shape icons for that object shine through walls. Used on spline nodes, which makes it easier to use splines when they are inside objects (e.g. the spline meshes).

grafik

image

### DIFF
--- a/Code/EditorPlugins/Scene/EnginePluginScene/RenderPipeline/EditorShapeIconsExtractor.cpp
+++ b/Code/EditorPlugins/Scene/EnginePluginScene/RenderPipeline/EditorShapeIconsExtractor.cpp
@@ -3,7 +3,9 @@
 #include <EnginePluginScene/RenderPipeline/EditorShapeIconsExtractor.h>
 #include <Foundation/IO/FileSystem/FileSystem.h>
 #include <Foundation/IO/TypeVersionContext.h>
+#include <Foundation/Reflection/Implementation/PropertyAttributes.h>
 #include <RendererCore/Components/SpriteComponent.h>
+#include <RendererCore/Debug/DebugRenderer.h>
 #include <RendererCore/Pipeline/RenderDataManager.h>
 #include <RendererCore/Pipeline/View.h>
 
@@ -131,7 +133,7 @@ void ezEditorShapeIconsExtractor::ExtractShapeIcon(const ezGameObject* pObject, 
     {
       pRenderData->m_hTexture = pShapeIconInfo->m_hTexture;
       pRenderData->m_fSize = m_fSize;
-      pRenderData->m_fMaxScreenSize = m_fMaxScreenSize;
+      pRenderData->m_fMaxScreenSize = m_fMaxScreenSize * ezDebugRenderer::GetTextScale();
       pRenderData->m_fAspectRatio = 1.0f;
       pRenderData->m_BlendMode = ezSpriteBlendMode::ShapeIcon;
       pRenderData->m_texCoordScale = ezVec2(1.0f);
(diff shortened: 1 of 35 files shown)


Data/Base/Shaders/Debug/DebugGeometry.ezShader
```diff
@@ -71,6 +77,13 @@ VS_OUT main(VS_IN_POS_ONLY Input)
 
 float4 main(PS_IN Input) : SV_Target
 {
+#if OCCLUDED_PASS
```
C-Core: Should be done in the vertex shader, same for the other debug shaders

Data/Base/Shaders/Debug/DebugPrimitive.ezShader
```diff
@@ -56,8 +62,27 @@ VS_OUT main(VS_IN Input)
...
+}
+
 float4 main(PS_IN Input) : SV_Target
 {
+#if OCCLUDED_PASS
+  {
+    float viewDist = Input.Position.w;
+    const float fadeStart = 30.0;
+    const float fadeDist = 70.0;
+    float fade = saturate(1.0 - (viewDist - fadeStart) / fadeDist);
```
C-Core, line 74: DebugPrimitives use alpha blending so we don't need to use dithering here

Data/Base/Shaders/Materials/SpriteMaterial.ezShader
```diff
@@ -107,6 +117,11 @@ VS_OUT main(uint VertexID : SV_VertexID, uint InstanceID : SV_InstanceID)
 
 #include <Shaders/Materials/MaterialHelper.h>
 
+float interleavedGradientNoise(float2 fragCoord) {
```
C-Core: We already have a InterleavedGradientNoise function in Common.h

Code/EditorPlugins/Scene/EnginePluginScene/RenderPipeline/EditorShapeIconsExtractor.cpp
```diff
@@ -131,7 +133,7 @@ void ezEditorShapeIconsExtractor::ExtractShapeIcon(const ezGameObject* pObject,
     {
       pRenderData->m_hTexture = pShapeIconInfo->m_hTexture;
       pRenderData->m_fSize = m_fSize;
-      pRenderData->m_fMaxScreenSize = m_fMaxScreenSize;
+      pRenderData->m_fMaxScreenSize = m_fMaxScreenSize * ezDebugRenderer::GetTextScale();
```
C-Core: This makes the icons very big on my 4k monitor. I'm so used to the small icons by now ;) It makes sense in principal to scale with text scale but maybe we should add an editor preference for the icon scale. The ShapeIconsExtractor already has a "MaxScreenSize" property but it is not exposed anywhere.

C-Core (approved):

Code/EditorPlugins/Scene/EnginePluginScene/RenderPipeline/EditorShapeIconsExtractor.cpp
jankrassnigg: Added a preference that let's you scale them down. Also fixed that the gizmo size preference was not sent to the engine on startup.

PR #1917, opened 2026-04-25
when opening prefab - expand scenegraph and select first object.

veksha:


Code/EditorPlugins/Scene/EditorPluginScene/Scene/SceneDocumentWindow.cpp
```diff
@@ -96,6 +96,15 @@ ezQtSceneDocumentWindow::ezQtSceneDocumentWindow(ezSceneDocument* pDocument)
     m_pDockManager->addDockWidgetTab(ads::LeftDockWidgetArea, pPanelTree);
   }
 
+  // If prefab: expand scenegraph and select the root object, as that is the most likely one to be edited.
+  if (pDocument->IsPrefab()) {
+    pDocument->TriggerExpandScenegraph();
+
+    ezDeque<const ezDocumentObject*> sel;
+    sel.PushBack(pDocument->GetObjectManager()->GetRootObject()->GetChildren()[0]);
```
jankrassnigg: There is an overload that takes just a single object.
Internally it does the same as you did here, though.

veksha: nice. changed.

PR #1922, opened 2026-04-28
Fixed #1914: Implemented copy&paste for component, maps and arrays

### LINKED ISSUE #1914: [FEATURE] copy/paste component or its values
**Describe the solution you'd like:**
- [ ] copy/paste component from one object to another.
- [ ] copy/paste all property values to component on a different object.

Pasting the values could be also allowed to **any** component _type_ (as long as property names are matching)
For example, from **default** character controller to **custom** character controller (which has almost the same properties)

**Additional context:**

How it looks in Unity:

Image

jankrassnigg:
Right click on a component header to copy or paste the entire object. Also works for unrelated types, for properties whose name and type match. Also fixed copy/paste of array and map properties.

### DIFF
--- a/Code/Tools/Libs/GuiFoundation/PropertyGrid/Implementation/PropertyBaseWidget.cpp
+++ b/Code/Tools/Libs/GuiFoundation/PropertyGrid/Implementation/PropertyBaseWidget.cpp
@@ -1,6 +1,8 @@
 #include <GuiFoundation/GuiFoundationPCH.h>
 
 #include <Foundation/IO/MemoryStream.h>
+#include <Foundation/Serialization/AbstractObjectGraph.h>
+#include <Foundation/Serialization/DdlSerializer.h>
 #include <Foundation/Serialization/ReflectionSerializer.h>
 #include <Foundation/Serialization/RttiConverter.h>
 #include <Foundation/Strings/TranslationLookup.h>
@@ -16,6 +18,7 @@
 #include <GuiFoundation/Widgets/InlinedGroupBox.moc.h>
 #include <ToolsFoundation/Command/TreeCommands.h>
 #include <ToolsFoundation/Object/ObjectAccessorBase.h>
+#include <ToolsFoundation/Serialization/DocumentObjectConverter.h>
 
 #include <GuiFoundation/GuiFoundationDLL.h>
 #include <QClipboard>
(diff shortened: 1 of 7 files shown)


Code/Tools/Libs/GuiFoundation/PropertyGrid/Implementation/PropertyBaseWidget.cpp
```diff
@@ -287,6 +383,109 @@ void ezQtPropertyWidget::ExtendContextMenu(QMenu& m)
...
+          ezRawMemoryStreamReader graphReader(content.m_ObjectGraph.GetData(), content.m_ObjectGraph.GetElementCount());
+          ezAbstractObjectGraph graph;
+          if (ezAbstractGraphDdlSerializer::Read(graphReader, &graph).Failed())
+            return;
+
+          const ezAbstractObjectNode* pNode = graph.GetNodeByName("root");
+          if (pNode == nullptr)
+            return;
+
+          m_pObjectAccessor->StartTransaction("Paste Component Values");
```
Sanakan8472, line 417: If pasting between non-matching types is not necessary, you can reduce this code quite a bit by using this:
```C++
          m_pObjectAccessor->StartTransaction("Paste Component Values");
          const ezRTTI* pClipboardType = ezRTTI::FindTypeByName(pNode->GetType());
          if (pClipboardType == nullptr)
            return;
          for (const ezDocumentObject* pTarget : targets)
          {
            if (pTarget->GetType() != pClipboardType)
              continue;
            ezAbstractObjectGraph baseGraph;
            {
              ezDocumentObjectConverterWriter writer(&baseGraph, pTarget->GetDocumentObjectManager());
              writer.AddObjectToGraph(pTarget, "root");
            }
            const ezAbstractObjectNode* pBaseRoot = baseGraph.GetNodeByName("root");
            if (pBaseRoot == nullptr)
              continue;

            // Remap clipboard GUIDs to match the target object's GUIDs so the diff can be applied
            // Note we are remapping the graph over and over, this is fine, only the target guid is important.
            graph.ReMapNodeGuidsToMatchGraph(pNode, baseGraph, pBaseRoot);
            ezDeque<ezAbstractGraphDiffOperation> diff;
            graph.CreateDiffWithBaseGraph(baseGraph, diff);
            if (diff.IsEmpty())
              continue;
            
            ezDocumentObjectConverterReader::ApplyDiffToObject(m_pObjectAccessor, pTarget, diff);
          }
          m_pObjectAccessor->FinishTransaction(); });
```
This would also allow pasting of sub-objects, not just by-value properties.

jankrassnigg: Allowing to paste into unrelated types was indeed intentional.

PR #1923, opened 2026-04-28
smooth player movement

veksha:
simple smoothing of player movement that works.
don't know if there are better methods.


jankrassnigg:
Works nicely, thanks!

PR #1927, opened 2026-04-30
Fixed #1926: EventMessageSenderHelper was accessing null object

### LINKED ISSUE #1926: [BUG] FPS template: crash when pushing door button.
**Bug description:**
FPS template: crash when pushing door button.

**How to Reproduce:**
1. create new project, select FPS template
2. start the game and try to push door button.
3. crash

maybe bug with SendEventMessage?

**Screenshots:**

Image

jankrassnigg:
This is a regression that was introduced in https://github.com/ezEngine/ezEngine/pull/1862

### DIFF
--- a/Code/Engine/Core/Messages/Implementation/EventMessageSender.cpp
+++ b/Code/Engine/Core/Messages/Implementation/EventMessageSender.cpp
@@ -26,7 +26,7 @@ namespace ezInternal
 
   bool EventMessageSenderHelper::SendEventMessage(ezMessage& ref_msg, ezComponent* pSenderComponent, ezGameObject* pSearchObject, ezSmallArray<ezComponentHandle, 1>& inout_cachedReceivers)
   {
-    ezWorld* pWorld = pSearchObject->GetWorld();
+    ezWorld* pWorld = pSenderComponent->GetWorld();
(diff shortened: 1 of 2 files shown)


Code/Engine/Core/Messages/Implementation/EventMessageSender.cpp
```diff
@@ -26,7 +26,7 @@ namespace ezInternal
 
   bool EventMessageSenderHelper::SendEventMessage(ezMessage& ref_msg, ezComponent* pSenderComponent, ezGameObject* pSearchObject, ezSmallArray<ezComponentHandle, 1>& inout_cachedReceivers)
   {
-    ezWorld* pWorld = pSearchObject->GetWorld();
```
clkern, line 29: This was changed intentionally, because the sender component is now also used as a filter and can be null.

C-Core: We could maybe assert that either the sender component or the search objects is not null and then do something like: ezWorld* pWorld = pSearchObject != nullptr ? pSearchObject->GetWorld() : pSenderComponent->GetWorld();

jankrassnigg: Changed.

C-Core (approved):

PR #1932, opened 2026-05-05
Add radius parameter to the projectile component

AlexandrKedrov:
Main changes are the following:
1. Add radius parameter to the projectile component
2. If radius > 0 then use `physicsInterface.SweepTestSphere` instead of `physicsInterface.Raycast`
3. Also if radius > 0 then adjust object position based on the hit position information

### DIFF
--- a/Code/EnginePlugins/GameComponentsPlugin/Gameplay/Implementation/ProjectileComponent.cpp
+++ b/Code/EnginePlugins/GameComponentsPlugin/Gameplay/Implementation/ProjectileComponent.cpp
@@ -29,6 +29,7 @@ EZ_BEGIN_STATIC_REFLECTED_TYPE(ezProjectileSurfaceInteraction, ezNoBase, 3, ezRT
     EZ_MEMBER_PROPERTY("ImpulseType", m_uiImpulseType)->AddAttributes(new ezDynamicEnumAttribute("PhysicsImpulseType")),
     EZ_MEMBER_PROPERTY("Impulse", m_fImpulse),
     EZ_MEMBER_PROPERTY("Damage", m_fDamage),
+    EZ_MEMBER_PROPERTY("InertiaRatio", m_fInertiaRatio)->AddAttributes(new ezDefaultValueAttribute(5.0f)),
   }
   EZ_END_PROPERTIES;
 }
@@ -45,6 +46,7 @@ EZ_BEGIN_COMPONENT_TYPE(ezProjectileComponent, 8, ezComponentMode::Dynamic)
     EZ_RESOURCE_MEMBER_PROPERTY("OnDeathPrefab", m_hDeathPrefab)->AddAttributes(new ezAssetBrowserAttribute("CompatibleAsset_Prefab", ezDependencyFlags::Package)),
     EZ_MEMBER_PROPERTY("CollisionLayer", m_uiCollisionLayer)->AddAttributes(new ezDynamicEnumAttribute("PhysicsCollisionLayer")),
     EZ_MEMBER_PROPERTY("Radius", m_fRadius)->AddAttributes(new ezClampValueAttribute(0.0f, ezVariant())),
+    EZ_MEMBER_PROPERTY("StaticVelocityRatio", m_fStaticVelocityRatio)->AddAttributes(new ezDefaultValueAttribute(0.05f), new ezClampValueAttribute(0.0f, ezVariant())),
     EZ_BITFLAGS_MEMBER_PROPERTY("ShapeTypesToHit", ezPhysicsShapeType, m_ShapeTypesToHit)->AddAttributes(new ezDefaultValueAttribute(ezVariant(ezPhysicsShapeType::Default & ~(ezPhysicsShapeType::Trigger)))),
     EZ_ACCESSOR_PROPERTY("FallbackSurface", GetFallbackSurfaceFile, SetFallbackSurfaceFile)->AddAttributes(new ezAssetBrowserAttribute("CompatibleAsset_Surface", ezDependencyFlags::Package)),
     EZ_ARRAY_MEMBER_PROPERTY("Interactions", m_SurfaceInteractions),
@@ -87,6 +89,7 @@ ezProjectileComponent::ezProjectileComponent()
   m_fMetersPerSecond = 10.0f;
   m_uiCollisionLayer = 0;
   m_fRadius = 0.0f;
+  m_fStaticVelocityRatio = 0.05f;
   m_fGravityMultiplier = 0.0f;
   m_vVelocity.SetZero();
   m_bSpawnPrefabOnStatic = false;
@@ -96,6 +99,8 @@ ezProjectileComponent::~ezProjectileComponent() = default;
 
 void ezProjectileComponent::Update()
 {
+  const float fPenetrationDepth = m_fRadius * 0.5f;
+
   if (m_fGravityMultiplier == 0.0f && m_fMetersPerSecond == 0.0f)
     return;
 
@@ -132,7 +137,7 @@ void ezProjectileComponent::Update()
     if (QueryCollision(*pPhysicsInterface, castResult, vCurPosition, vCurDirection, fDistance, queryParams))
     {
       const ezVec3 vNewCenterPosition = (m_fRadius > 0.0f)
-        ? CalculateSphereCenterPosition(vCurPosition, vCurDirection, castResult, m_fRadius * 0.5f)
+        ? CalculateSphereCenterPosition(vCurPosition, vCurDirection, castResult, fPenetrationDepth)
         : castResult.m_vPosition;
 
       const ezSurfaceResourceHandle hSurface = castResult.m_hSurface.IsValid() ? castResult.m_hSurface : m_hFallbackSurface;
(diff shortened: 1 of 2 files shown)


Code/EnginePlugins/GameComponentsPlugin/Gameplay/Implementation/ProjectileComponent.cpp
```diff
@@ -44,6 +44,7 @@ EZ_BEGIN_COMPONENT_TYPE(ezProjectileComponent, 7, ezComponentMode::Dynamic)
     EZ_MEMBER_PROPERTY("SpawnPrefabOnStatic", m_bSpawnPrefabOnStatic),
     EZ_RESOURCE_MEMBER_PROPERTY("OnDeathPrefab", m_hDeathPrefab)->AddAttributes(new ezAssetBrowserAttribute("CompatibleAsset_Prefab", ezDependencyFlags::Package)),
     EZ_MEMBER_PROPERTY("CollisionLayer", m_uiCollisionLayer)->AddAttributes(new ezDynamicEnumAttribute("PhysicsCollisionLayer")),
+    EZ_MEMBER_PROPERTY("Radius", m_fRadius)->AddAttributes(new ezDefaultValueAttribute(0.0f), new ezClampValueAttribute(0.0f, ezVariant())),
```
jankrassnigg: ezDefaultValueAttribute(0) is not needed. The default for all values is 0 without this attribute.

Code/EnginePlugins/GameComponentsPlugin/Gameplay/Implementation/ProjectileComponent.cpp
```diff
@@ -64,10 +65,44 @@ EZ_BEGIN_COMPONENT_TYPE(ezProjectileComponent, 7, ezComponentMode::Dynamic)
...
+      // It says that the hit position is "behind" (if we move along vDirection) vOrigin
+      // Likely this can only happen when there is already a collision before the particle has started moving
+      // Thus, return the original line position
+      return vOrigin;
+    }
+
+    // Distance from hit pos to the line passing through the vOrigin along vDirection
+    // This distance is equal to r * sin(theta),
+    // where theta is the angle between vDirection and the vector connecting hit pos and supposed new sphere center (vNewCenter)
+    const float fDistSquared = vHitRelative.GetLengthSquared() - fHitToDirProj * fHitToDirProj;
```
jankrassnigg: You can't calculate with squared distances. You can only really use them for comparisons.

Say one distance is 4, and another is 2. So their difference is 2.
If you work with their squares you get: `(4*4) = 16` and `(2*2) = 4`, so then `16 - 4 = 12`.
`sqrt(12)` does not equal 2.

Once you want to do mathematical operations with lengths, you do need to calculate the real lengths right away. But since you do a sqrt down below anyway, it doesn't make a difference in performance, just do it at the right time.

Code/EnginePlugins/GameComponentsPlugin/Gameplay/Implementation/ProjectileComponent.cpp
```diff
@@ -64,10 +65,44 @@ EZ_BEGIN_COMPONENT_TYPE(ezProjectileComponent, 7, ezComponentMode::Dynamic)
 EZ_END_DYNAMIC_REFLECTED_TYPE;
 // clang-format on
 
+namespace
+{
+  /// \brief Helper function to recalculate the sphere position from hit position
+  /// \param vOrigin is expected to be the initial position of the particle before SweepTest
+  ezVec3 CalculateSphereCenterPosition(const ezVec3& vOrigin, const ezVec3& vDirection, float fRadius, const ezVec3& vHitPosition)
```
jankrassnigg: The sphere sweep castResult already contains an m_fDistance value, that tells you how far the sweep got. It will also be clamped to 0, so no need for a "behind" check.

Code/EnginePlugins/GameComponentsPlugin/Gameplay/Implementation/ProjectileComponent.cpp
```diff
@@ -64,10 +65,44 @@ EZ_BEGIN_COMPONENT_TYPE(ezProjectileComponent, 7, ezComponentMode::Dynamic)
...
+    const float fDistSquared = vHitRelative.GetLengthSquared() - fHitToDirProj * fHitToDirProj;
+
+    // Here fOffset is r * cos(theta),
+    // it equals to the distance between vHitPosition and vNewCenter projected to vDirection
+    const float fOffset = ezMath::Sqrt(ezMath::Max(0.0f, fRadius * fRadius - fDistSquared));
+
+    const ezVec3 vNewCenter = vOrigin + (fHitToDirProj - fOffset) * vDirection;
+
+    // Lerp result to make projectile go a bit "inside" into the object it collided
+    return ezMath::Lerp(vNewCenter, vHitPosition, 0.1f);
```
jankrassnigg: I think this entire function boils down to:

```cpp
vOrigin + vDirection * (castResult.m_fDistance + fPenetrationDepth)
```

`castResult.m_fDistance` should give you the exact distance that the projectile can move forwards to touch the collider. This considers the sphere radius already.

`fPenetrationDepth` is how deep you want the projectile to get stuck inside the geometry. I think `fRadius*0.5f` would be a sensible hardcoded default, I wouldn't expose this as another property.

jankrassnigg (commented):
Overall this looks very good!

Code/EnginePlugins/GameComponentsPlugin/Gameplay/Implementation/ProjectileComponent.cpp
AlexandrKedrov: It is not an  actual difference. It is Pythagorean theorem to calculate the 3rd side of a triangle, which also appears to be a distance from hit point to the sphere travel trajectory 

It is an exact mathematical result, it can also be obtained using x^2 + y^2 + z^2 = r^2 formula

Anyway, I will try checking for m_fDistance from SweepTest

Code/EnginePlugins/GameComponentsPlugin/Gameplay/Implementation/ProjectileComponent.cpp
```diff
@@ -188,16 +212,16 @@ void ezProjectileComponent::Update()
 
 
           GetWorld()->DeleteObjectDelayed(GetOwner()->GetHandle());
-          vNewPosition = castResult.m_vPosition;
+          vNewPosition = vNewCenterPosition;
         }
         else if (interaction.m_Reaction == ezProjectileReaction::Reflect || interaction.m_Reaction == ezProjectileReaction::Bounce)
         {
           /// \todo Should reflect around the actual hit position
           /// \todo Should preserve travel distance while reflecting
```
AlexandrKedrov: I think I can also take a look at these todos

Code/EnginePlugins/GameComponentsPlugin/Gameplay/Implementation/ProjectileComponent.cpp
AlexandrKedrov: The condition `m_vVelocity.GetLength() < 1.0f` looks very strict. I think it can be improved. What was the original idea to make it this way?

AlexandrKedrov (commented):
I found a few more things I would try to improve

Code/EnginePlugins/GameComponentsPlugin/Gameplay/Implementation/ProjectileComponent.cpp
jankrassnigg: That is a good question. I have no idea 😆 You could look at the git history, whether it came in a commit that gives you a clue. Most code in here is really old, but it also got tweaked over time.

AlexandrKedrov: Okay. I will investigate it. Unfortunately, I have a busy working schedule. I think, I will be able to take a look at it only on the weekends

jankrassnigg: That's fine, no hurry. Just let me know when it's good to merge.

AlexandrKedrov: This code is from the original commit which added bouncing interaction type. 
I made an additional check that the projectile is on ground (otherwise, the projectile can get stuck in the air) and made the threshold to be adjustable

Code/EnginePlugins/GameComponentsPlugin/Gameplay/Implementation/ProjectileComponent.cpp
AlexandrKedrov: That is the difficult part. 
I made the following changes:
1) Added a within-frame position correction after reflection/bounce. The projectile first advances to the estimated impact moment, then continues for the remaining frame time

The thing I am worried about here - because the position is adjusted, there is a higher risk that projectile will be stuck in geometry

2) I changed the rotation behavior. In simple words, now the rotation is proportional to the velocity. I think it is appropriate to call it "spinning" 

Conceptually, I would say there are two spinning mechanism, and they rotate the object in the opposite directions (it helps for me to visualize it as a brick hitting the wall):
- “sliding”: due to low-friction contact
- “sticking and pushing away”: due to high-friction contact. 

I made an empirical parameter `fInertiaRatio`, which can take positive (stick behavior) and negative (sliding) sign and describes how the projectile is affected by spinning

I will provide a demonstrating video in a few days. Also I can elaborate on some things and open for discussions.

AlexandrKedrov: Also I would add an optional full reflection of the owner object orientation about the surface plane. It is not realistic, but may find some use

Code/EnginePlugins/GameComponentsPlugin/Gameplay/Implementation/ProjectileComponent.cpp
```diff
@@ -188,52 +217,61 @@ void ezProjectileComponent::Update()
...
+            if (interaction.m_Reaction == ezProjectileReaction::Bounce)
             {
-              m_vVelocity = ezVec3::MakeZero();
-              m_fGravityMultiplier = 0.0f;
+              ezResourceLock<ezSurfaceResource> pSurface(hSurface, ezResourceAcquireMode::BlockTillLoaded);
+
+              if (pSurface)
+              {
+                vNewVelocity *= pSurface->GetDescriptor().m_fPhysicsRestitution;
+              }
```
AlexandrKedrov, line 249: This is from the original version. 
But physically it is a bit incorrrect. 

The surface mostly absorbs and reflects not the velocity, but the normal projection of the velocity.  Thus if there is any absorption of the energy, the velocity won't be perfectly reflected - its reflection angle will deviate from incident angle

The tangent component may also be absorbed - I think it happens due to friction of the surface, but it is a bit different mechanism 

So shortly, the normal and tangent components should realistically be scaled separately, and the second I think is smaller

AlexandrKedrov:
Original. The main part I didn't like, that the projectile would make large rotation even for small velocity changes (when the object is almost stil)

https://github.com/user-attachments/assets/69103a9b-55a1-40ae-8578-52dd54493c4b

https://github.com/user-attachments/assets/5c2dbed0-1327-4b8f-a3bc-ac669d45b1f8

Proposed changes:

Inertia ratio: 5 (sticking)

https://github.com/user-attachments/assets/8a64f2ae-7c09-40c0-80a9-9ab7c60b43f1

Inertia ratio: -15 (sliding) 

https://github.com/user-attachments/assets/eefaa1d4-04be-42de-8ba6-2c5e641b34d9

Third person, inertia ratio 5 (sticking)

https://github.com/user-attachments/assets/b2191f1d-c67d-4445-9e92-713e43f2be0c

AlexandrKedrov:
@jankrassnigg, what do you think? Mainly, I am concerned about the rotation behaviour. Do we keep the old version, or try to adjust a new one

Code/EnginePlugins/GameComponentsPlugin/Gameplay/Implementation/ProjectileComponent.cpp
```diff
@@ -356,6 +416,72 @@ void ezProjectileComponent::DeserializeComponent(ezWorldReader& inout_stream)
...
+
+bool ezProjectileComponent::ShouldStopProjectile(const ezPhysicsWorldModuleInterface& physicsInterface, const ezPhysicsCastResult& castResult, const ezVec3& vVelocity)
+{
+  const ezVec3 vGravity = physicsInterface.GetGravity();
+  if (!vGravity.IsZero())
+  {
+    const ezVec3 vGravityDir = vGravity.GetNormalized();
+    // Check that projectile has hit the ground
+    // if not - return false to make sure it won't hang in the air after being stopped
+    if (vGravityDir.Dot(castResult.m_vNormal) > -0.6f)
```
jankrassnigg: So that's 53 degree? Quite steep still. Btw you should be able to write:

```cpp
if (vGravityDir.Dot(castResult.m_vNormal) > ezMath::Cos(ezAngle::MakeFromDegree(40)))
```

Due to constexpr evaluation, this should still boil down to a constant, but is a bit clearer to understand.

jankrassnigg:
We don't use the bounce behavior in our samples so far and generally this component is meant mostly as an example how things could be done. So the component doesn't have to solve all problems, I would expect many games to implement their own projectile components, at least if they want to do something slightly out of the ordinary. Your changes don't break anything in our samples and I'm fine with integrating the proposed changes, so that there is more useful behavior to play around with.

I actually always wanted to add a "grenade launcher" weapon to the samples, though I never took the time. Maybe that would be easier now.

AlexandrKedrov:
I've finished with the coding part. How do I make changes to the documentation? Is there a separate repository?

jankrassnigg:
> I've finished with the coding part. How do I make changes to the documentation? Is there a separate repository?

Yes there is a separate repository here:
https://github.com/ezEngine/docs-src

The file to update is here:
https://github.com/ezEngine/docs-src/blob/main/pages/docs/gameplay/projectile-component.md

For simple updates it's sufficient to just edit a file and make a commit. For more complex updates, run the top level `GenerateDocs.bat` file to update the TOC, check the links, etc. If you don't do it locally, the build machine will make those changes automatically while updating the website.

Code/EnginePlugins/GameComponentsPlugin/Gameplay/Implementation/ProjectileComponent.cpp
AlexandrKedrov: Physically it leads to the following effect: if the object falls on the floor, it will continue moving for some time on its surface 

I could add an additional scaling parameter for tangent component. However,  I am a bit worried it can lead to some ill conditioned behavior when the normal component is extremely small, and the tangent component is large. Thus, I didn't implement this

I still think it is a nice effect, though

jankrassnigg: I think we have enough parameters now, the component is already quite complex.

PR #1935, opened 2026-05-08
Enhanced error message for missing package dependencies.

veksha:
I think this error message can be improved:
image

to something like this:
image

With this info user can find the place of an error and replace missing script, for example.

image

Not sure if it's the best way of doing this.


Code/Editor/EditorFramework/Assets/Implementation/AssetCurator.cpp
```diff
@@ -2655,3 +2657,69 @@ ezAssetCurator::ReplaceAssetResult ezAssetCurator::ReplaceAssetReferenceInUses(e
 
   return result;
 }
+
+void ezAssetCurator::FindAssetUsagesInGameObjects(ezUuid assetToFind, const ezDocumentObject* pObject, ezDynamicArray<ezString>& out_usages)
```
jankrassnigg: This function shouldn't be in ezAssetCurator, since it is specific to the GameObjectDocument, which the asset curator should not know anything about.

It would make more sense to add a virtual "FindAssetUsages" function to the base class ezAssetDocument. Then ezGameObjectDocument can override it. The base class implementation would just be empty.

Then AssetCuratorPanel only has to cast a document to ezAssetDocument, which is fine, and call that function on it.

Code/Editor/EditorFramework/Assets/Implementation/AssetCurator.cpp
```diff
@@ -2655,3 +2657,69 @@ ezAssetCurator::ReplaceAssetResult ezAssetCurator::ReplaceAssetReferenceInUses(e
...
+            if (const ezGameObjectDocument* pGameDoc = ezDynamicCast<const ezGameObjectDocument*>(pObject->GetDocumentObjectManager()->GetDocument()))
+            {
+              pGameDoc->GenerateFullDisplayName(pObject, sFullPath);
+              out_usages.PushBack(sFullPath);
+            }
+          }
+        }
+      }
+    }
+    break;
```
jankrassnigg: Just some general optimization advice:

Always do the check first, that has the most impact. This loop is quite expensive, since you go through everything. And afterwards you also recurse into children. But at the very inner place you only check, whether this is an ezGameObjectDocument, at all.

The order should be:

1. Check if this is an ezGameObjectDocument, if not RETURN. No loop, no recursion.
2. Convert the asset GUID to string once.
3. Enter the loop
4. Check for the ezAssetBrowserAttribute, this is the most distinguishing attribute. Its existance pretty much guarantees the rest.
5. Then do the other checks (string type, temporary attribute, doesn't make a big difference anymore in terms of how many objects get filtered out).
6. DON'T convert the property string to GUID. This is not so cheap and will be done many times. Instead, compare the string with the asset GUID string that we created in step 2.

Finally, there should probably be a limit to how many items get returned. Ie, once you have 3 names, there is no need to continue. Should be an argument to pass in, and when you return, you have to return the recursion below also has to early out.


If you've copied this function from somewhere, please let me know, then I'll double check whether the source also needs improvement.

EDIT: Since this is called recursively, we shouldn't even give it an asset GUID and convert that to string, but instead pass in the stringified asset GUID right away, and do the whole GUID -> string conversion only once outside.

Code/Editor/EditorFramework/Panels/AssetCuratorPanel/Implementation/AssetCuratorPanel.cpp
```diff
@@ -189,8 +189,37 @@ void ezQtAssetCuratorPanel::UpdateIssueInfo()
...
+      ezString sDocumentPath = pSubAsset->m_pAssetInfo->m_Path.GetAbsolutePath();
+
+      // Open the document (without requesting a window)
+      ezDocument* pDocument = ezQtEditorApp::GetSingleton()->OpenDocument(sDocumentPath, ezDocumentFlags::None);
+      ezDynamicArray<ezString> uses;
+      if (pDocument != nullptr)
+      {
+        // Find all direct uses of this asset
+        ezAssetCurator::GetSingleton()->FindAssetUsagesInGameObjects(guid, pDocument->GetObjectManager()->GetRootObject(), uses);
+      }
```
jankrassnigg: I would pass in a limit of a few max elements to report (3 to 5), and then also use an ezTempHybridArray with that limit for the results.

Code/Editor/EditorFramework/Panels/AssetCuratorPanel/Implementation/AssetCuratorPanel.cpp
```diff
@@ -189,8 +189,37 @@ void ezQtAssetCuratorPanel::UpdateIssueInfo()
...
+        for (const ezString& use : uses)
+        {
+          if (!usesString.IsEmpty())
+          {
+            usesString.Append(", ");
+          }
+          usesString.Append(use);
+        }
+
+        sTmp.SetFormat("Dependency GUID: {}. Used by objects: {}", sDep, usesString);
```
jankrassnigg: I don't know whether the asset curator output supports line breaks. You could give that a try. Otherwise try shortening the final string a bit.

jankrassnigg (commented):
Overall a good idea. I've had the same problem as well.

This is a good start, at some point instead of just names, one could even report back the game object GUID or so, such that when opening the document through the asset curator, it could automatically get selected. But such things can be done later.

I would restructure where the code lives and improve the efficiency of the search algorithm (see comments below).

Also I'd like @Sanakan8472 's opinion on this.

Code/Editor/EditorFramework/Panels/AssetCuratorPanel/Implementation/AssetCuratorPanel.cpp
veksha: in LogModel.cpp i see `sTemp.ReplaceAll("\n", " ");`
so line breaks are replaced.

Code/Editor/EditorFramework/Assets/Implementation/AssetCurator.cpp
veksha: Done. Please check latest changes.

Code/Editor/EditorFramework/Assets/Implementation/AssetCurator.cpp
veksha: yes, I copied code from `ezAssetCurator::ReplaceAssetReferenceInObject`, you might want to check it too.

Code/Editor/EditorFramework/Panels/AssetCuratorPanel/Implementation/AssetCuratorPanel.cpp
veksha: I set it to 3

Code/Editor/EditorFramework/Document/Implementation/GameObjectDocument.cpp
```diff
@@ -1048,3 +1048,78 @@ void ezGameObjectDocument::HandleEngineMessage(const ezEditorEngineDocumentMsg*
...
   }
 }
+
+// the following method is similar to "ezAssetCurator::ReplaceAssetReferenceInObject"
+
+void ezGameObjectDocument::FindAssetUsages(ezString assetToFind, const ezDocumentObject* pObject, ezTempHybridArray<ezString, 3>& out_usages, uint maxResults) const
+{
+  const ezGameObjectDocument* pGameDoc = ezDynamicCast<const ezGameObjectDocument*>(pObject->GetDocumentObjectManager()->GetDocument());
+  if (pGameDoc == nullptr)
+    return;
```
jankrassnigg: "this" is already an ezGameObjectDocument. There is no need for the pGameDoc pointer, at all.

Code/Editor/EditorFramework/Assets/AssetDocument.h
```diff
@@ -129,6 +129,9 @@ class EZ_EDITORFRAMEWORK_DLL ezAssetDocument : public ezDocument
   /// \brief Handles all messages received from the corresponding ezEngineProcessDocumentContext on the engine process.
   virtual void HandleEngineMessage(const ezEditorEngineDocumentMsg* pMsg);
 
+  /// \brief Finds all usages of the given asset in this document and appends them to out_usages. The default implementation does nothing, override this if your document can reference other assets.
+  virtual void FindAssetUsages(ezString assetToFind, const ezDocumentObject* pObject, ezTempHybridArray<ezString, 3>& out_usages, uint maxResults) const {}
```
jankrassnigg: pass assetToFind as an ezStringView, and name it "sAssetToFind"

out_usages can be an ezDynamicArray, no need for the same type here, you can pass an ezTempHybridArray reference to it anyway.

"uint maxResults" -> ezUInt32

It's also quite weird that the caller is supposed to pass in an object here (pObject). This should be handled internally. That means here in the signature, remove pObject. In GameObjectDocument you may then need two functions, one as the entry point, another for the recursive call where you do pass in the current object.

Code/Editor/EditorFramework/Document/GameObjectDocument.h
```diff
@@ -140,6 +140,9 @@ class EZ_EDITORFRAMEWORK_DLL ezGameObjectDocument : public ezAssetDocument
 
   virtual void HandleEngineMessage(const ezEditorEngineDocumentMsg* pMsg) override;
 
+  /// \brief Finds all usages of the given asset in this document and appends them to out_usages. The default implementation does nothing, override this if your document can reference other assets.
+  virtual void FindAssetUsages(ezString assetToFind, const ezDocumentObject* pObject, ezTempHybridArray<ezString, 3>& out_usages, uint maxResults = 3) const override;
```
jankrassnigg: same changes as in ezAssetDocument

Code/Editor/EditorFramework/Panels/AssetCuratorPanel/Implementation/AssetCuratorPanel.cpp
```diff
@@ -189,8 +191,40 @@ void ezQtAssetCuratorPanel::UpdateIssueInfo()
...
+      // Open the document (without requesting a window)
+      ezDocument* pDocument = ezQtEditorApp::GetSingleton()->OpenDocument(sDocumentPath, ezDocumentFlags::None);
+      ezTempHybridArray<ezString, 3> uses;
+      if (pDocument != nullptr)
+      {
+        // cast a document to ezAssetDocument to access the FindAssetUsages function.
+        ezAssetDocument* pAssetDoc = ezDynamicCast<ezAssetDocument*>(pDocument);
+
+        // Find all direct uses of this asset
+        pAssetDoc->FindAssetUsages(sDep, pDocument->GetObjectManager()->GetRootObject(), uses, 3);
```
jankrassnigg:
```suggestion
        if (ezAssetDocument* pAssetDoc = ezDynamicCast<ezAssetDocument*>(pDocument))
        {
          pAssetDoc->FindAssetUsages(sDep, pDocument->GetObjectManager()->GetRootObject(), uses, 3);
        }
```

Code/Editor/EditorFramework/Document/Implementation/GameObjectDocument.cpp
```diff
@@ -1048,3 +1048,78 @@ void ezGameObjectDocument::HandleEngineMessage(const ezEditorEngineDocumentMsg*
...
+
+  auto pDocument = pObject->GetDocumentObjectManager()->GetDocument();
```
jankrassnigg: no need to get pDocument, your "this" pointer is already the document that you work on here

Code/Editor/EditorFramework/Document/Implementation/GameObjectDocument.cpp
```diff
@@ -1048,3 +1048,78 @@ void ezGameObjectDocument::HandleEngineMessage(const ezEditorEngineDocumentMsg*
...
+          ezVariant value;
+          if (pAccessor->GetValue(pObject, pProp, value).Succeeded())
+          {
+            auto sValue = value.Get<ezString>();
+            if (sValue == assetToFind)
+            {
+              ezStringBuilder sFullPath;
+              pGameDoc->GenerateFullDisplayName(pObject, sFullPath);
+              out_usages.PushBack(sFullPath);
+            }
```
jankrassnigg, line 1159: Move the maxResults check after the PushBack here

Code/Editor/EditorFramework/Panels/AssetCuratorPanel/Implementation/AssetCuratorPanel.cpp
```diff
@@ -1,3 +1,5 @@
+#include "EditorFramework/Assets/AssetDocument.h"
```
jankrassnigg: Use angle brackets (`< >`) for the includes

Code/Editor/EditorFramework/Document/Implementation/GameObjectDocument.cpp
```diff
@@ -1048,3 +1048,78 @@ void ezGameObjectDocument::HandleEngineMessage(const ezEditorEngineDocumentMsg*
...
+      continue;
+
+    // Skip temporary properties
+    if (pProp->GetAttributeByType<ezTemporaryAttribute>() != nullptr)
+      continue;
+
+    switch (pProp->GetCategory())
+    {
+      // also add Array, Set, Map categories as in "ezAssetCurator::ReplaceAssetReferenceInObject"?
+      case ezPropertyCategory::Member:
```
jankrassnigg, line 1086: Looking at `ezAssetCurator::ReplaceAssetReferenceInObject`, since you only took the `Member` codepath, this wouldn't find references that are in sets, maps and arrays. Shouldn't be too hard to add those cases as well.

Issue #1942, opened 2026-05-19
[BUG]character controller: inconsistent velocity near a tilted wall, tilted ground.


veksha:
**Bug description:**

Issue was not present before this merge: https://github.com/ezEngine/ezEngine/pull/1637
by @AlexandrKedrov 

When player is moving up on a tilted surface near a tilted wall some strange force will add up to it's velocity even if player released move key already.
I assume it is Jolt stair stepping code tries to move player up a step.

so when you have a lot of objects lying around (stones) on an uneven ground and you walk near them, sometimes player speed will increase and i think it is because of this issue with stepping.

**How to Reproduce:**

[test.zip](https://github.com/user-attachments/files/28010292/test.zip)

1. import test.ezScene to Testing Chambers
2. run the scene and go along the wall (pointing camera between moving direction and a wall).

**Videos:**

https://github.com/user-attachments/assets/bb4e9800-2dba-46ef-83fd-11fc2e810b62

https://github.com/user-attachments/assets/41677f68-8f47-46b8-99a7-db7ffb4a68f1

AlexandrKedrov:
Hi! I will look into that. Have you tried decreasing the foot radius? Perhaps, it is large for your current character collision shape

veksha:
Hi, Alexandr!
I tried now, but it makes no difference.

veksha:
I found out that movement of character after key was released is because I added this PR: https://github.com/ezEngine/ezEngine/pull/1923
It is player movement smoothing. Basically, it will slowly interpolate key input to zero after key release.

but even without that smoothing - problem is still there, velocity of movement is increased near a tilted wall.
Also on regular stairs smoothing works ok, it only conflicts in situation that was described in this issue it seems.

AlexandrKedrov:
I think I will be able to investigate this only at the weekends. Besides, I assume the cause of the problem might be in the collision handling mechanism. If I am not mistaken, it just adds some velocity to the character in the direction of the shortest way out of the object, in which the character has "got stuck". Because the shape is irregular and small, this chosen direction might be in some sense unintuitive, and move the character somewhere upwards

veksha:
Turns out Jolt Samples have this problem too. I will try to do some workarounds. For me it is crucial to make speed of the character stable and predictable on uneven ground around uneven colliders.

https://github.com/user-attachments/assets/d3459df9-c657-4681-aca4-51323fc2e34c

AlexandrKedrov:
[...]

PR #1943, opened 2026-05-24
#1942 Correct horizontal stepping calculation

AlexandrKedrov:
This PR is mainly for testing purposes, it (at least partially) resolves issue #1942. 

In the original version, the step forward calculation was very large, when the character appeared to be on the tilted ground. It has led to the artificial acceleration when stair stepping mode got enabled.  To make calculations correct, the `up`  vector is replaced (which in most cases is  the zAxis) with the ground normal vector


AlexandrKedrov:
The issue is already resolved in JoltPhysics repository

PR #1948, opened 2026-05-29
Render Graph

Sanakan8472:
## Render Graph

The render graph lives in `Code/Engine/RendererCore/RenderGraph/`. It is a declarative, deferred recording of GPU work where each pass announces the transient textures and buffers it reads/writes; the graph then culls dead passes, allocates and aliases transient resources, computes the required barriers, and finally executes the recorded passes in declaration order. Graphs are created via the `ezRenderGraphManager`:

```cpp
// ### Application start:
ezSharedPtr<ezRenderGraph> pGraph = ezRenderGraphManager::CreateRenderGraph("MyGraph", ezRenderGraphPhase::Render);

// ### Each frame during ezRenderWorldRenderEvent::Type::BeginRender or ezGALDeviceEvent::AfterBeginFrame:
pRenderGraph->Reset();
// *Record passes* (see below)
// ...
ezRenderGraphManager::EnqueueRenderGraph(pRenderGraph);
```

The phases (`PreRender`, `Render`, `PostRender`) set during creation define a coarse ordering; within a phase graphs run in FIFO order of their `EnqueueRenderGraph` calls.

### Recording

To record a graph, it must first be reset via `Reset()`. Once enqueued via `EnqueueRenderGraph`, it cannot be modified until `Reset()` is called again. There are three types of passes that can be added to a graph: `AddGraphicsPass`, `AddComputePass` and `AddTransferPass`. Transfer passes only allow you can execute copy operations on the command encoder.

To be able to interact with mutable resources in a graph, these need to be registered with the graph via:
```cpp
// Create transient texture:
ezGALTextureCreationDescription desc;
// Setup texture description...
ezRenderGraphTextureHandle hTexture = graph.CreateTexture(desc);

// Import existing GAL texture:
ezRenderGraphTextureHandle hGraphTexture = graph.ImportTexture(hDepthTexture, ezGALResourceState::DepthStencilRead, ezGALShaderStageFlags::Auto);
```

The same can be done for buffers in the same way. During pass execution, `ResolveTexture` / `ResolveBuffer` needs to be called to resolve a graph handle to a GAL handle. You can also use `ReplaceImportedTexture` to replace a graph handle with an actual GAL handle (if compatible) after recording. This is mainly used to repalce swapchain textures. A simple transfer pass might look like this:
```cpp
auto pass = graph.AddTransferPass("CopyTexture");
pass.ReadTexture(hInput, {}, ezGALResourceState::CopySource);
pass.WriteTexture(hOutput, {}, ezGALResourceState::CopyDestination);
pass.SetExecuteCallback([=](const ezRenderGraphContext& ctx)
{
  ctx.GetCommandEncoder()->CopyTexture(ctx.ResolveTexture(hOutput), ctx.ResolveTexture(hInput));
});
```
Every pass consists of the following:
1. Declaring what (non-immutable) resources are read and in which state. A new enum `ezGALResourceState` is added to define the states.
1. Declaring what (non-immutable) resources are written and in which state.
1. An execution callback that is called when the graph is executed. Note, that you can skip this if you just want to set barriers.
1. `HasSideEffects` can be added if you don't want you pass be culled due to nobody consuming its output.

The graphics pass extends this with the following:
1. Setting color and depth targets.
1. Setting clear behaviour.
1. Setting stereoscopic mode.

This is roughly the same interface as `ezGALRenderingSetup` has. Note that adding color / depth targets will automatically call `ReadTexture` / `WriteTexture` for you.
```cpp
auto pass = m_pRenderGraph->AddGraphicsPass("Render");
pass.AddColorTarget(hTempColor, {}, ezGALRenderTargetLoadOp::Clear);
pass.SetClearColor(0);
pass.AddDepthStencilTarget(hTempStencil, {},
  ezGALRenderTargetLoadOp::Clear, ezGALRenderTargetStoreOp::Discard,
  ezGALRenderTargetLoadOp::Clear, ezGALRenderTargetStoreOp::Discard);
pass.SetClearDepth();
pass.SetClearStencil();
pass.SetStereoscopic(camera.IsStereoscopic());
pass.SetExecuteCallback([=](const ezRenderGraphContext& ctx)
{
  //...
});
```
The pass type is `ezRenderGraphPassBuilder`. Note that in each scope, only one instance can be alive so you must scope multiple `Add*Pass` calls.

### Compilation

Before graphs can be executed, they need to be compiled which consists of:
1. `ValidateGraph` checks for dead GAL handles.
1. `BuildDependencyGraph` creates a DAG to be able to cull passes that produce no used output.
1. `CullDeadPasses` removes passes with no used output.
1. `BuildSortedPassList` creates a linear list of remaining passed to be executed.
1. `ComputeResourceLifetimes` computes for each graph handle the first and last pass it is used in.
1. `AllocateTransientResources` creates actual GAL handles for the graph handles via the `ezRenderGraphResourceAllocator`. The code tries to share GAL handles for multiple graph handles.
1. `BuildRenderingSetups` create the `ezGALRenderingSetup` used for render passes.

Afterwards the `ezRenderGraphManager` uses the new class `ezGALResourceStateTracker` to compute barriers for the entire frame by calling `ComputeBarriers` on each graph in order of execution. Finally all graphs are executed and then all modified resources are barriered back into the default state via the `ezGALResourceStateTracker`.

### Execution

The `ezRenderGraphManager` will execute all graphs during `ExecuteRenderGraphs`:

- `ezRenderGraphManager::EnqueueRenderGraph(graph)` queues a graph for the current frame.
- `ezRenderGraphManager::ExecuteRenderGraphs(device)` drives compilation, barrier computation and execution between `BeforeBeginFrame` and `AfterEndFrame`.
- The shared `ezRenderGraphResourcePool` (`RenderGraphResourcePool.h`) and per-graph `ezRenderGraphResourceAllocator` (`RenderGraphResourceAllocator.h`) recycle transient GPU resources across graphs and frames; resources unreferenced for a configurable number of frames are garbage-collected.

A more in depth description can be found in the documentation: https://github.com/ezEngine/docs-src/pull/21

## RendererFoundation

- **ezGALCommandEncoder**:
  - Removed `GenerateMipMaps`. Instead, the new `ezRenderGraphUtils::GenerateMipMaps` function can be used which generates the necessary passes in a graph.
  - Added `TextureBarrier` and `BufferBarrier` functions. These are driven by the render graph and should not be called manually.
  - In debug builds (via define `EZ_BARRIER_VALIDATION`), all barriers, bind groups and render targets are tracked and validated. Failures are propagated via callbacks to the high level renderer via `ezTextureValidationError` and `ezBufferValidationError` so that the render graph can output the barriers observed that lead to the failure of this resource. Note, that errors are only output once per unqiue error to not spam the log.
- **Descriptors**:
  - Buffer and texture `m_ResourceAccess.m_bImmutable` is set to `false` by default and only set to `true` for resources and other truly immutable things. `ezGALCommandEncoder` now fails if immutable resources are tried to be modified.
  - Added `ezGALTextureCreationDescription::Validate()`. Now used by both the device and render graph to check for validity.
  - Added `ezGALBufferCreationDescription::GetDefaultState()` and `ezGALTextureCreationDescription::GetDefaultState()` these return the `ezGALResourceState` used as the initial state. At the end of a frame, all resources return to their default state.
- **Enumerations**:
  - Added `ezGALResourceState` enumeration. See render graph docs.
  - Removed `ezGALTextureUsageFlags::DynamicMipGeneration`. Should be replaced with `RenderTarget` when mip map generation is needed.
  - Added `ezGALTextureUsageFlags::Presentable`. This is only used for swapchains and results in them defaulting to the `ezGALResourceState::Present` state.
  - Added`ezGALShaderStageFlags::Auto` (bit 15). Reserved for the render graph to infer a shader stage from a resource state. When in doubt, it's safe to always leave the stage at `Auto`.
  - Added reflection to all enums.
- **Misc**:
  - Added `ezGALDevice::s_SwapChainUpdatedEvent` to inform higher level code that the swap-chain was changed, e.g. during acquire image as the swap-chain was out of date.
  - Added `ezGALResourceStateTracker`. A tracker for both buffer and texture states on a subresource level. It is used both inside the command encoder for validation and inside the render graph to compute barrier sequences.
  - New POD `ezGALRenderTargetRange`. Defines the sub-resources a render target view is rendering to. Used by the render graph to define a render target.

## RendererCore

- Added `ezRenderViewContext::UpdateViewport`. Updates global constants and encoder with the viewport information of the view data. Needed as this was previously done behind the scenes.
- Added `ezViewData::m_uiSkyIrradianceIndex` which is needed to create proper barriers before rendering starts.
- Added `ezRenderContext::ResetBindGroup`. This is called on all bind groups between render pass execution so that no state bleeds between unrelated passes.
- Extended `ezBindGroupBuilder` to allow bindings to overwrite `ezGALTextureType`. 
 
## GameEngine

- `ezWindowOutputTargetBase::CaptureImage` was replaced with an asynchronous `StartCaptureImage` / `WaitCaptureImage` pair (returning a new enum `ezCaptureImageResult`). `ezWindowOutputTargetGAL` now drives capture through a dedicated render graph.
- `ezRenderTargetActivatorComponent`, `ezEditorSelectedObjectsExtractor` and other code that calls `ezRenderWorld::AddViewToRender` will now also call `ezRenderWorld::AddViewDependency` to define how the view is consumed to place proper barriers. This is forwarded to `ezRenderPipeline::AddViewDependency` eventually during rendering.

## Vulkan Changes

- Removed `ezPipelineBarrierVulkan` and the related `ezImageCopyVulkan`. Their image-layout tracker, deferred barrier lists and full-screen blit helpers are no longer needed: layout transitions are now expressed through `ezGALTextureBarrier` / `ezGALBufferBarrier` and emitted explicitly by the render graph.
- Added `ezBarrierUtilsVulkan` which translates GAL barriers (and a raw `vk::Image`/`vk::Buffer` variant for swap-chains and init-time uploads) into `vkCmdPipelineBarrier(2)` calls. It uses `VK_KHR_synchronization2` (`vkCmdPipelineBarrier2KHR`) when the device supports it and falls back to the Vulkan 1.1 path otherwise.
- **Synchronization2 + swap-chain maintenance**: `ezGALDeviceVulkan` now detects and enables `VK_KHR_synchronization2`, `VK_EXT_surface_maintenance1` and `VK_EXT_swapchain_maintenance1`. The function pointers (`vkCmdPipelineBarrier2KHR`) are loaded through the dispatch context.
- `ezGALCommandEncoderImplVulkan`'s lazy render-pass begin/clear bookkeeping (`m_bRenderPassActive`, `m_bClearSubmitted`, `m_pBoundRenderTargets`, `GenerateMipMapsPlatform` …) is removed. Rendering begin/end and barriers are now driven by the higher level.
- `ezGALTextureVulkan` lost its preferred-layout/access/stage caches `m_preferredLayout`, `m_stages`, `m_access`. `ComputeCreateInfo` no longer fills layout/stage/access outputs; layout selection now goes through `ezConversionUtilsVulkan::GetTextureLayout(ezGALResourceState)` and `ConvertResourceState(...)` (with both Synchronization1 and Synchronization2 overloads).
- **InitContext / SwapChainVulkan / BufferVulkan / SharedTextureVulkan** were updated to issue explicit GAL-state transitions via `ezBarrierUtilsVulkan` instead of going through the deleted pipeline-barrier helper.

## Misc

- `ezQtEngineViewWidget` now uses a timer instead of resizing the engine viewport when the parent window resizes because Nvidia driver deadlocks on Linux if you do so in rapid succession.
- Fixed 100% CPU core usage by moving dock-layout capture to the `showEvent` in `ezQtDocumentWindow` instead of polling.
- Compile fix for Qt 6.19.0 in `ezQtTreeSearchFilterModel`.
- Only run "Press the any key to continue..." at the end of unit tests on Windows as Linux generally does not open terminals during execution which need to be kept open.
- Undefined behaviour fix on Vulkan for screen space shadows: Added `EZ_IMAGE_FORMAT` macro to change the UAV resource to the correct format: `EZ_IMAGE_FORMAT("r8") RWTexture2DArray<unorm float> OutputTexture;`. Without this, writing 32 bit to an 8 bit UAV would result in undefined behaviour and validation warnings.
- Fixed `ezGALDeviceDX11::FillCapabilitiesPlatform` to correctly set MSAA capabilities on depth formats.

## BREAKING CHANGES

The old `ezRenderPipelinePass` interface (`GetRenderTargetDescriptions`, `InitRenderPipelinePass`, `Execute`) has been replaced. All passes must now implement `AddRenderPasses` instead:

1. **Remove** `GetRenderTargetDescriptions`, `InitRenderPipelinePass`, and `Execute` overrides.
1. **Implement** `AddRenderPasses(viewData, camera, graph, inputs, outputs)`.
1. **Read inputs** via `inputs[m_PinX.m_uiInputIndex].m_TextureHandle`. Validate with `IsInvalidated()`.
1. **Create outputs** via `graph.CreateTexture(desc)` and assign to `outputs[m_PinX.m_uiOutputIndex].m_TextureHandle`. For pass-through pins, forward the input handle directly.
1. **Add render graph passes** via `graph.AddGraphicsPass` / `AddComputePass` / `AddTransferPass`. Declare resource access (`ReadTexture`, `WriteTexture`, `AddColorTarget`, `AddDepthStencilTarget`).
1. **Move rendering code** into `pass.SetExecuteCallback(lambda)`. Use `ctx.ResolveTexture()` to get real GPU handles inside the callback.
1. **Remove** manual `BeginRendering` / `EndRendering` calls — handled automatically by the render graph.
1. If your pass produces outputs and can be toggled off, **override** `AddRenderPassesInactive` to create a cleared output texture.
1. Add `const ezRenderViewContext& renderViewContext = *ctx.GetUserData<ezRenderViewContext>();` at the start of each execution callback to gain access to the view context.
1. Call `renderViewContext.UpdateViewport();` to apply the viewport settings and set global constants in case you are rendering a pass that renders materials / objects that were extracted in this view.
1. Replace `SetupResources` with the new `ezRenderPipelinePass::BindDataProviderResources` call inside the execution callback. You also need to call on the pass `ezRenderPipelinePass::SetupResourceDependencies` which sets up the resource states for `BindDataProviderResources` to work.


Code/Engine/RendererCore/RenderWorld/RenderWorld.h
```diff
@@ -27,14 +27,10 @@ struct ezRenderWorldRenderEvent
   enum class Type
   {
     BeginRender,             ///< Fired before rendering begins.
-    BeforePipelineExecution, ///< Fired before executing a render pipeline.
```
C-Core, line 30: Why have these callbacks been removed? I use them in my game to bind custom global resources to the "frame group"

C-Core (approved):

Code/Engine/RendererCore/RenderWorld/RenderWorld.h
Sanakan8472: The problem with these is that they no longer do what they previously did. `BeforePipelineExecution` was firing before rendering but the render world does not render anymore. Rendering is deferred until the render graph executes at the end of the frame. Thus, I thought it be best to remove them (as I didn't see any use of it) and instead add `ezRenderGraphRenderEvent::BeforeGraphExecution`.
What is the use case for which you made use if these?

Sanakan8472: Oh binding global resources outside of the pass that renders them will no longer work. All bind groups are purged after every pass. The problem is that resources would bleed between passes that actually made use of them unintentionally, making barriers impossible to place. E.g. a shadow map render pass would bind and use deferred decal textures that were bound in a previous pipeline. Or slots were named the same and random resources were bound from other passes in the past.
We could maybe be a bit more selective with what we purge in the bind groups, e.g. don't purge the frame bind group with a graph. They must be purged between graphs though as that is were most of the bleed through happens. The old callback wouldn't help you there as that is too far detached from rendering at this point.

Sanakan8472: @C-Core I've pushed a change that should address the issue: The user data is now an `ezReflectedClass` so you can dyncast it. Also `ResetBindGroups` is now only called between graphs before executing the `ezRenderGraphRenderEvent::Type::BeforeGraphExecution` event.
So you can hook into `BeforeGraphExecution` instead of the old `BeforePipelineExecution` which should be at the same point in time during rendering.
To figure out which pipeline is currently being rendered run this in the callback:
```cpp
    if (auto pViewContext = ev.m_pContext->GetUserData<ezRenderViewContext>())
    {
      //pViewContext->m_pPipeline 
    }
```

PR #1955, opened 2026-06-06
Track resource dependencies through extraction (and more polishing of resource states)

Sanakan8472:
- Add texture and buffer dependency records to `ezMsgExtractRenderData` via the new methods `AddDependency(ezGALTextureHandle hTexture, ezRenderData::Category category, ...)` and `AddDependency(ezGALBufferHandle hBuffer, ezRenderData::Category category, ...)`. You do not need to validate the handle, the functions will ignore invalid handles. When ever you are extracting `ezRenderData`, call these functions if you are using resources that are mutable on the GPU.
These will be cached for static objects like the render data.
  - You will need to declare the dependencies in the render graph passes before rendering each category. Wherever you called `RenderDataWithCategory` you now must call with the same category `DeclareRendererDependenciesForCategory` when declaring the pass.
- Added `ezRenderWorld::AddViewDependency` overload for buffers.
- Extend barrier validation to copy, readback, resolve, update, draw, dispatch, vertex/index buffer and fixed proxy texture validation.
- Rebinding Vulkan descriptor sets after pipeline layout changes, and clamping pushed constant bytes to the shader range to fix validation errors.
- Add allocator-wrapper template support to `ezDelegate` and use an aligned allocator wrapper for render graph execute callbacks to be able to capture structs declared via the shaders which are 16byte aligned and will fail to be captured otherwise.
- Show the mouse cursor for the GameEngine animation test window runtime config.

### DIFF
--- a/Code/Engine/RendererCore/Pipeline/Implementation/Extractor.cpp
+++ b/Code/Engine/RendererCore/Pipeline/Implementation/Extractor.cpp
@@ -165,11 +165,51 @@ void ezExtractor::ExtractRenderData(const ezView& view, const ezGameObject* pObj
 #endif
   };
 
+  // Forwards the barrier dependencies recorded on the message into the extracted render data. The stored category is the (possibly redirected) category passed to msg.AddDependency; it is resolved here to the static/dynamic variant the same way render data categories are resolved, so dependencies land in the same category as the render data they accompany.
+  auto AddDependenciesFromMessage = [&](bool bDynamic) {
+    for (ezTextureDependency dep : msg.m_TextureDependencies)
+    {
+      dep.m_uiCategory = (msg.m_OverrideCategory != ezInvalidRenderDataCategory)
+                           ? msg.m_OverrideCategory.m_uiValue
+                           : ezRenderData::ResolveCategory(ezRenderData::Category(dep.m_uiCategory), bDynamic).m_uiValue;
+      extractedRenderData.AddDependency(dep);
+    }
+
+    for (ezBufferDependency dep : msg.m_BufferDependencies)
+    {
+      dep.m_uiCategory = (msg.m_OverrideCategory != ezInvalidRenderDataCategory)
+                           ? msg.m_OverrideCategory.m_uiValue
+                           : ezRenderData::ResolveCategory(ezRenderData::Category(dep.m_uiCategory), bDynamic).m_uiValue;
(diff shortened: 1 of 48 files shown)


Code/Engine/RendererCore/Pipeline/Implementation/Extractor.cpp
```diff
@@ -231,10 +273,22 @@ void ezExtractor::ExtractRenderData(const ezView& view, const ezGameObject* pObj
...
+          for (ezBufferDependency& dep : msg.m_BufferDependencies)
+          {
+            dep.m_uiCategory = ezRenderData::ResolveCategory(ezRenderData::Category(dep.m_uiCategory), false).m_uiValue;
+          }
+
+          ezRenderWorld::CacheRenderData(view, pObject->GetHandle(), pComponent->GetHandle(), uiComponentVersion, newCacheEntries, msg.m_TextureDependencies, msg.m_BufferDependencies);
         }
 
         AddRenderDataFromMessage(msg);
+        AddDependenciesFromMessage(false);
```
C-Core: It doesn't change anything, but it would be more consistent to pass the msg as parameter for the new lambda as well. And it would be easier to read.

Code/Engine/RendererCore/RenderWorld/Implementation/RenderWorld.cpp
```diff
@@ -101,7 +101,15 @@ namespace ezInternal
       ezUInt16 m_uiVersion = 0;
     };
 
+    struct PerObjectDependenciesCache
+    {
+      ezHybridArray<ezTextureDependency, 4> m_TextureDependencies;
```
C-Core: Could use ezSmallArray here

Code/Engine/RendererCore/RenderWorld/Implementation/RenderWorld.cpp
```diff
@@ -113,6 +121,8 @@ namespace ezInternal
       ezGameObjectHandle m_hOwnerObject;
       ezComponentHandle m_hOwnerComponent;
       PerObjectCache m_Cache;
+      ezHybridArray<ezTextureDependency, 2> m_TextureDependencies;
```
C-Core: Also here, ezSmallArray to make the struct a bit more compact

C-Core (approved):

PR #1957, opened 2026-06-07
Fixed #1954: Added CVar to change the default texture filtering mode

### LINKED ISSUE #1954: Default texture samplers should be modifyable
Currently when a texture is set up, it gets a sampler state and then that's it, the filtering options can't be changed anymore.

But for the "default" samplers (default quality, low quality, high quality), it should be possible to just redirect to them, so that changing them after the texture is loaded is still possible.

jankrassnigg:
Texture samplers now know which default mode they represent, and if their settings change, they get recreated right away.

This allows to change the texture filtering mode at any time and see the visual / performance difference.

### DIFF
--- a/Code/Engine/RendererFoundation/Device/Device.h
+++ b/Code/Engine/RendererFoundation/Device/Device.h
@@ -406,7 +406,7 @@ protected:
   virtual ezGALSamplerState* CreateSamplerStatePlatform(const ezGALSamplerStateCreationDescription& Description) = 0;
   virtual void DestroySamplerStatePlatform(ezGALSamplerState* pSamplerState) = 0;
   /// \brief Destroys and reinitializes a sampler state in-place, applying AdjustSamplerStateDescription to pick up current quality settings.
-  virtual void RecreateSamplerStatePlatform(ezGALSamplerState* pSamplerState) {};
+  virtual void RecreateSamplerStatePlatform(ezGALSamplerState* pSamplerState) = 0;
 
   virtual ezGALBindGroupLayout* CreateBindGroupLayoutPlatform(const ezGALBindGroupLayoutCreationDescription& Description) = 0;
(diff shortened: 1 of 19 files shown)


Code/Engine/RendererFoundation/Device/Implementation/Device.cpp
```diff
@@ -339,6 +339,80 @@ void ezGALDevice::DestroyHashedResource(Handle& inout_hResource, Table& table, e
   inout_hResource.Invalidate();
 }
 
+void ezGALDevice::SetTextureQualityMode(ezUInt32 uiMode, ezGALTextureQuality::Enum quality)
+{
+  m_QualityModes[uiMode] = quality;
```
C-Core: Assert for out of bounds access

Code/Engine/RendererFoundation/Device/Device.h
```diff
@@ -482,6 +501,8 @@ class EZ_RENDERERFOUNDATION_DLL ezGALDevice
   ezUInt32 m_uiGraphicsPipelines = 0;
   ezUInt32 m_uiComputePipelines = 0;
   ezGALCommandEncoderStats m_EncoderStats;
+
+  ezEnum<ezGALTextureQuality> m_QualityModes[8];
```
C-Core: why 8 when we only ever use 5? It would be much clearer what is meant here if we would use an actual enum ("LowestQuality", "lowQuality", "default" etc.) than just a number as mode.

jankrassnigg: Fixed, uses an enum now.

Code/Engine/RendererFoundation/Device/Implementation/Device.cpp
jankrassnigg: fixed

C-Core (approved):

PR #1962, opened 2026-06-09
Non-destructive Brush Based Terrain System

jankrassnigg:
### Added a Terrain plugin which implements non-destructive terrain editing functionality.

image

The "terrain patch" component represents a heightfield, which can be modified by placing "terrain brushes". These are objects with brush components attached, which define how to modify (raise, lower, paint) the terrain. When moving a brush, the affected terrain patches are updated in real-time.

Patches can also be initialized from image data, to import height information from external sources.

Collision meshes are pre-baked on scene export, making the system compatible with the procedural placement system.

Terrain patches require dedicated shaders for rendering. The necessary infrastructure can easily be included in a custom shader. The system supports many different layers, since the shader simply uses a 2D array texture. The exact defails are up to the shader. On the heightfields themselves, up to 5 layers may be blended over a single vertex without artifacts (can be lowered to 4 or 3 to make the shader faster).

image

Brushes use "rounded rect" signed-distance fields, so they can be rectangular, circular and anything in between. They also allow to use noise. Brushes can be placed in prefabs to easily adjust terrain below an object (e.g. a building). When an object has both a brush and a spline component attached, the brush will follow the spline curve, to easily generate trails and mountains.

### Functionality that's NOT available at the moment:

* LOD - all terrain patches are always rendered at full detail. May be added later.
* Erosion - not really suitable for this kind of system.
* Caves / overhangs - not possible with heightfields.
* Great terrain material shader - there is a lot that could be improved to make the terrain look better (noise, height base layer blending).


Code/EnginePlugins/TerrainPlugin/Components/TerrainPatchComponent.cpp
```diff
@@ -0,0 +1,496 @@
...
+  pRenderData->m_uiSortingKey = 0;
+  pRenderData->m_uiNumInstances = 1;
+  pRenderData->m_DataOffsets.m_uiInstance = m_InstanceDataOffset.m_uiOffset;
+  pRenderData->m_hInstanceDataBuffer = hInstanceDataBuffer;
+
+  msg.AddRenderData(pRenderData, ezDefaultRenderDataCategories::LitOpaque, ezRenderData::Caching::IfStatic);
+
+  // The vertex shader samples these persistent GPU buffers, so declare them as render-graph
+  // dependencies for the categories that render this patch.
+  if (!hHeightBuffer.IsInvalidated())
```
Sanakan8472, line 340: Nit: No need to check for validity, the `AddDependency` function already does that for you.

Code/EnginePlugins/TerrainPlugin/TerrainSystem.cpp
```diff
@@ -0,0 +1,855 @@
...
+  // Always allocate at least one element so the binding is always valid.
+  // BrushCount=0 prevents the shader from indexing into it.
+  if (brushes.IsEmpty())
+    brushes.ExpandAndGetRef();
+
+  ezGALBufferCreationDescription brushDesc;
+  brushDesc.m_uiStructSize = sizeof(TerrainBrushData);
+  brushDesc.m_uiTotalSize = brushes.GetCount() * sizeof(TerrainBrushData);
+  brushDesc.m_BufferFlags = ezGALBufferUsageFlags::StructuredBuffer | ezGALBufferUsageFlags::ShaderResource;
+  brushDesc.m_ResourceAccess.m_bImmutable = false;
```
Sanakan8472: Nit: I've switched the default for `m_bImmutable` in the render graph PR. It is false by default now. It really only needs to be set in e.g. resources that load immutable data which is only like 2 or 3 places in the code base.

Code/EnginePlugins/TerrainPlugin/TerrainSystem.cpp
```diff
@@ -0,0 +1,855 @@
...
+{
+  if (uiIdx >= m_Heightfields.GetCount())
+    return;
+
+  auto& data = m_Heightfields[uiIdx];
+  uiIdx = ezInvalidIndex;
+
+  ezGALDevice* pDevice = ezGALDevice::GetDefaultDevice();
+
+  if (!data.m_hBakedHeights.IsInvalidated())
```
Sanakan8472: No need to check for validity. Just call `Destroy*`, it will do the check for you and also invalidate the handle reference you passed in.

Code/EnginePlugins/TerrainPlugin/TerrainSystem.cpp
```diff
@@ -0,0 +1,855 @@
...
+  m_hHeightfieldSharedMask = pDevice->CreateBuffer(maskDesc);
+  m_uiHeightfieldSharedMaskStoredSize = uiStoredSize;
+}
+
+void ezTerrainSystem::DestroyHeightfields()
+{
+  ezGALDevice* pDevice = ezGALDevice::GetDefaultDevice();
+  for (auto& patch : m_Heightfields)
+  {
+    if (!patch.m_hBakedHeights.IsInvalidated())
```
Sanakan8472: Same here and below, just call `DestroyBuffer` without validity checks.

Code/EnginePlugins/TerrainPlugin/TerrainSystem.cpp
```diff
@@ -0,0 +1,855 @@
...
+  ezGALBufferHandle hSourceBuffer = pDevice->CreateBuffer(srcDesc, ezArrayPtr<const ezUInt8>(reinterpret_cast<const ezUInt8*>(srcHeights.GetData()), srcHeights.GetCount() * sizeof(float)));
+
+  const ezUInt32 uiGroups = (uiStoredSize + 15) / 16;
+
+  // The intermediate mask is shared bake scratch — grow it to fit this patch's stored grid.
+  EnsureSharedHeightfieldScratch(uiStoredSize);
+
+  // Import all buffers into the graph. Transient buffers start in ShaderResource (default for SRV-only).
+  // Persistent bake buffers default to UnorderedAccess (they have both SRV and UAV flags).
+  auto hGraphSrc = graph.ImportBuffer(hSourceBuffer, ezGALResourceState::ShaderResource);
```
Sanakan8472: Note that if you set a state during `ImportBuffer` / `ImportTexture` this creates an immediate barrier into this state at the start of the graph. As your very first pass below calls `ReadBuffer` on these, setting the desired state during import isn't really necessary. Doesn't really affect the outcome here though.

Code/EnginePlugins/TerrainPlugin/TerrainSystem.cpp
```diff
@@ -0,0 +1,855 @@
...
+  srcDesc.m_ResourceAccess.m_bImmutable = false;
+
+  ezGALBufferHandle hSourceBuffer = pDevice->CreateBuffer(srcDesc, ezArrayPtr<const ezUInt8>(reinterpret_cast<const ezUInt8*>(srcHeights.GetData()), srcHeights.GetCount() * sizeof(float)));
+
+  const ezUInt32 uiGroups = (uiStoredSize + 15) / 16;
+
+  // The intermediate mask is shared bake scratch — grow it to fit this patch's stored grid.
+  EnsureSharedHeightfieldScratch(uiStoredSize);
+
+  // Import all buffers into the graph. Transient buffers start in ShaderResource (default for SRV-only).
```
Sanakan8472: The comment here is technically correct, but there is no guarantee that these resources still hold the default state at this time during the frame if somebody else touched them so you always need to assume they are in the wrong state.

Sanakan8472 (approved):

Issue #1706, opened 2025-11-01
Animated sprite for sprite component


Lt-droid:
On sprite component,

- add spritesheet for sprite anim.(x and y parameter)
- add **loop** option (default to enabled)
- add **number of loops** parameter(default to -1 -> infinite)
- **at animation end** (Restart/Delete/Nothing)(default to restart)
- add **framerate** linked to "sprite time"
- add **number of sprite** parameter default to sprite
- make **max screensize** optional( Enable/Disable).

C-Core:
Note that this can be implemented entirely on the component side since the rendering code already supports a tex coord offset/scale that is currently set to default. It can be used to select a "subframe" in a spritesheet.

Mr-Fishy:
Hello! I found this issue and would like to work on it! I have experience working on game engines before and would love to help out! Are there any further pointers or changes to the issue since it was last updated?

jankrassnigg:
Yes, feel free to work on it, there haven't been any changes since then.

I don't think it's complicated, just take a look and ask questions when you run into problems.

PR #1967, opened 2026-06-22
Feature: Animated Sprite Component Addition

Mr-Fishy:
Adding sprite animation support per issue #1706. 

This PR enhances the existing `ezSpriteComponent` to support native spritesheet animations and introduces a toggle for the maximum screen size constraint. 

Key Additions:

- Added the requested `UseMaxScreenSize` toggle. 
- Defined grid dimensions (`Columns`, `Rows`) and `Framerate` to automatically slice and animate a texture.
- Added playback controls via an `IsAnimated` toggle and an `EndAction` enum (Loop, Stop, Destroy) to control behavior when the animation concludes.
- Added a `Loops` property to specify exactly how many times an animation should play before triggering a Stop or Destroy action.
- Single-frame static sprites retain their IfStatic render data caching, while animated sprites dynamically disable caching to allow UV updates. 
- Versioning has been bumped to 4 to ensure backward compatibility in serialization.
- Altered data member packing and reduced class size by ~12 bytes.

I decided to go with the "Is Animated" toggle rather than the "Loop" toggle as it is more self-descriptive and reduces confusion between "Loop" and "Restart" terminology. Instead the "Is Animated" controls whether the sprite is considered for caching as well as it allows quick-exit from update calls. Then I added the requested "Restart", "Delete", "Nothing" actions to "Loop", "Stop", "Destroy" as they are more intuitive for end-users. 

If these adjustments to the implementation are too far against the requested additions then I am happy to realign my implementation!


Code/Engine/RendererCore/Components/Implementation/SpriteComponent.cpp
```diff
@@ -55,16 +61,25 @@ bool ezSpriteRenderData::CanBatch(const ezRenderData& other0) const
...
+EZ_BEGIN_COMPONENT_TYPE(ezSpriteComponent, 4, ezComponentMode::Static)
 {
   EZ_BEGIN_PROPERTIES
   {
     EZ_RESOURCE_ACCESSOR_PROPERTY("Texture", GetTexture, SetTexture)->AddAttributes(new ezAssetBrowserAttribute("CompatibleAsset_Texture_2D")),
     EZ_ENUM_MEMBER_PROPERTY("BlendMode", ezSpriteBlendMode, m_BlendMode),
     EZ_ACCESSOR_PROPERTY("Color", GetColor, SetColor)->AddAttributes(new ezExposeColorAlphaAttribute()),
     EZ_ACCESSOR_PROPERTY("Size", GetSize, SetSize)->AddAttributes(new ezClampValueAttribute(0.0f, ezVariant()), new ezDefaultValueAttribute(1.0f), new ezSuffixAttribute(" m")),
+
+    EZ_MEMBER_PROPERTY("UseMaxScreenSize", m_bUseMaxScreenSize)->AddAttributes(new ezDefaultValueAttribute(false)),
```
C-Core: This should be default "true" otherwise it will change the behavior of existing components

Code/Engine/RendererCore/Components/Implementation/SpriteComponent.cpp
```diff
@@ -55,16 +61,25 @@ bool ezSpriteRenderData::CanBatch(const ezRenderData& other0) const
...
     EZ_ACCESSOR_PROPERTY("MaxScreenSize", GetMaxScreenSize, SetMaxScreenSize)->AddAttributes(new ezClampValueAttribute(0.0f, ezVariant()), new ezDefaultValueAttribute(64.0f), new ezSuffixAttribute(" px")),
     EZ_MEMBER_PROPERTY("AspectRatio", m_fAspectRatio)->AddAttributes(new ezClampValueAttribute(0.0f, ezVariant()), new ezDefaultValueAttribute(1.0f)),
+
+    EZ_MEMBER_PROPERTY("IsAnimated", m_bIsAnimated)->AddAttributes(new ezDefaultValueAttribute(false)),
```
C-Core: defaultvalue is not needed here, bools are always false by default

Code/Engine/RendererCore/Components/Implementation/SpriteComponent.cpp
```diff
@@ -55,16 +61,25 @@ bool ezSpriteRenderData::CanBatch(const ezRenderData& other0) const
...
+    EZ_MEMBER_PROPERTY("Columns", m_uiColumns)->AddAttributes(new ezClampValueAttribute(1, ezVariant()), new ezDefaultValueAttribute(1)),
+    EZ_MEMBER_PROPERTY("Rows", m_uiRows)->AddAttributes(new ezClampValueAttribute(1, ezVariant()), new ezDefaultValueAttribute(1)),
+    EZ_MEMBER_PROPERTY("Framerate", m_fFramerate)->AddAttributes(new ezClampValueAttribute(0.0f, ezVariant()), new ezDefaultValueAttribute(24.0f)),
+    EZ_ENUM_MEMBER_PROPERTY("EndAction", ezSpriteAnimationEndAction, m_EndAction)->AddAttributes(new ezDefaultValueAttribute(ezSpriteAnimationEndAction::Default)),
```
C-Core: default value is not needed, it will automatically be the enum's default

Code/Engine/RendererCore/Components/SpriteComponent.h
```diff
@@ -105,10 +123,24 @@ class EZ_RENDERERCORE_DLL ezSpriteComponent : public ezRenderComponent
...
+  ezEnum<ezSpriteBlendMode> m_BlendMode;
+  ezEnum<ezSpriteAnimationEndAction> m_EndAction = ezSpriteAnimationEndAction::Default;
+  bool m_bUseMaxScreenSize = true;
+  bool m_bIsAnimated = false;
+
   float m_fSize = 1.0f;
   float m_fMaxScreenSize = 64.0f;
   float m_fAspectRatio = 1.0f;
+
+  ezUInt32 m_uiColumns = 1;
```
C-Core: ezUInt8 should be enough for column and row

Code/Engine/RendererCore/Components/SpriteComponent.h
```diff
@@ -105,10 +123,24 @@ class EZ_RENDERERCORE_DLL ezSpriteComponent : public ezRenderComponent
...
+  ezUInt32 m_uiRows = 1;
+  float m_fFramerate = 24.0f;
+
+  ezInt32 m_uiLoops = 1;
```
C-Core: should be m_iLoops if it is a signed integer.

Code/Engine/RendererCore/Components/Implementation/SpriteComponent.cpp
```diff
@@ -55,16 +61,25 @@ bool ezSpriteRenderData::CanBatch(const ezRenderData& other0) const
...
+    EZ_MEMBER_PROPERTY("Loops", m_uiLoops)->AddAttributes(new ezClampValueAttribute(1, ezVariant()), new ezDefaultValueAttribute(1)),
```
C-Core: what if I want to have infinite loops? I would suggest allowing this to go to -1 = infinite. You can use ezMinValueTextAttribute("Infinite") to make the -1 show as infinite in the UI. 0 and 1 would then mean the same, just one loop.

Code/Engine/RendererCore/Components/Implementation/SpriteComponent.cpp
```diff
@@ -85,6 +100,43 @@ EZ_END_COMPONENT_TYPE;
...
+  const ezUInt32 uiTotalFrames = m_uiColumns * m_uiRows;
+  if (uiTotalFrames <= 1)
+    return; // Nothing to animate.
+
+  ezTime tDiff = GetWorld()->GetClock().GetTimeDiff();
+  m_TimeSinceStart += tDiff;
+
+  const float fTotalAnimTime = uiTotalFrames / ezMath::Max(0.001f, m_fFramerate);
+
+  ezLog::Info("Time Since Start: {}, Total Animation Time: {}", m_TimeSinceStart.GetSeconds(), fTotalAnimTime);
```
C-Core: This will spam the log and should be removed or commented out

Code/Engine/RendererCore/Components/Implementation/SpriteComponent.cpp
```diff
@@ -85,6 +100,43 @@ EZ_END_COMPONENT_TYPE;
...
+  if (m_TimeSinceStart.GetSeconds() >= fTotalAnimTime)
+  {
+    m_uiCurrentLoop++;
+    ezLog::Info("Current Loop Changed to {}", m_uiCurrentLoop);
```
C-Core: Same log spam here

Code/Engine/RendererCore/Components/Implementation/SpriteComponent.cpp
```diff
@@ -85,6 +100,43 @@ EZ_END_COMPONENT_TYPE;
...
+
+    m_TimeSinceStart -= ezTime::MakeFromSeconds(fTotalAnimTime);
+
+    if (m_EndAction != ezSpriteAnimationEndAction::Loop)
+    {
+      if (m_uiCurrentLoop >= m_uiLoops)
+      {
+        m_bIsFinished = true;
+
+        if (m_EndAction == ezSpriteAnimationEndAction::Destroy)
```
C-Core: Might be worth looking into ezOnComponentFinishedAction. I think you don't need the ezSpriteAnimationEndAction enum then. The loop variable itself lets you decide whether to loop or not and ezOnComponentFinishedAction decides what happens on finish. You probable also don't need m_bIsFinished since you can just check for m_uiCurrentLoop >= m_uiLoops.

C-Core (commented):
Thanks a lot for the contribution! Overall this looks pretty good, I've added a few ideas for further improvements.

Code/Engine/RendererCore/Components/Implementation/SpriteComponent.cpp
jankrassnigg: 0 could also just be "infinite"

Code/Engine/RendererCore/Components/SpriteComponent.h
Mr-Fishy: Kept `m_uiLoops` and changed the type to `ezUInt32` and will use `0` as infinite loops.

Code/Engine/RendererCore/Components/Implementation/SpriteComponent.cpp
Mr-Fishy: Using `ezOnComponentFinishedAction` would definitely work! 

I will remove the `ezSpriteAnimationEndAction` and use the predefined finished action. Then the looping control will depend on if the `m_uiLoops` variable like you suggest (0 == infinite looping, > 0 == n-loops until end action). Then if the end action is `None`, the animation stops, otherwise it will do the specified action.

If I were to remove `m_bIsFinished` then the `m_bIsAnimated`  would need to be set to `false` after the end condition. Otherwise using `m_uiCurrentLoop >= m_uiLoops` at the quick-exit condition in the update function would never allow the end-action to be reached. 

I will go ahead and reduce the `m_bIsFinished` to use `m_bIsAnimated` instead. My initial thought was that having them separate would allow better control from the user side for scripting. But similar handling can be achieved by grouping and smart component management.

Code/Engine/RendererCore/Components/Implementation/SpriteComponent.cpp
```diff
@@ -85,6 +94,42 @@ EZ_END_COMPONENT_TYPE;
...
+  {
+    m_uiCurrentLoop++;
+
+    m_TimeSinceStart -= ezTime::MakeFromSeconds(fTotalAnimTime);
+
+    if (m_uiLoops != 0 && m_uiCurrentLoop >= m_uiLoops)
+    {
+      m_bIsAnimated = false;
+
+      if (m_EndAction == ezOnComponentFinishedAction::DeleteComponent)
```
C-Core: Use ezOnComponentFinishedAction::HandleFinishedAction here. There is actually more logic to it than what you have here, that's why we have the helper function. You also need to add a message handler for ezMsgDeleteGameObject and call ezOnComponentFinishedAction::HandleDeleteObjectMsg in there. You can look at the ezFmodEventComponent for an example.

Code/Engine/RendererCore/Components/Implementation/SpriteComponent.cpp
```diff
@@ -55,16 +55,25 @@ bool ezSpriteRenderData::CanBatch(const ezRenderData& other0) const
...
     EZ_ACCESSOR_PROPERTY("MaxScreenSize", GetMaxScreenSize, SetMaxScreenSize)->AddAttributes(new ezClampValueAttribute(0.0f, ezVariant()), new ezDefaultValueAttribute(64.0f), new ezSuffixAttribute(" px")),
     EZ_MEMBER_PROPERTY("AspectRatio", m_fAspectRatio)->AddAttributes(new ezClampValueAttribute(0.0f, ezVariant()), new ezDefaultValueAttribute(1.0f)),
+
+    EZ_MEMBER_PROPERTY("IsAnimated", m_bIsAnimated),
+    EZ_MEMBER_PROPERTY("Columns", m_uiColumns)->AddAttributes(new ezClampValueAttribute(1, ezVariant()), new ezDefaultValueAttribute(1)),
+    EZ_MEMBER_PROPERTY("Rows", m_uiRows)->AddAttributes(new ezClampValueAttribute(1, ezVariant()), new ezDefaultValueAttribute(1)),
+    EZ_MEMBER_PROPERTY("Framerate", m_fFramerate)->AddAttributes(new ezClampValueAttribute(0.0f, ezVariant()), new ezDefaultValueAttribute(24.0f)),
+    EZ_MEMBER_PROPERTY("Loops", m_uiLoops)->AddAttributes(new ezClampValueAttribute(0, ezVariant()), new ezDefaultValueAttribute(0), new ezMinValueTextAttribute("Infinite")),
+    EZ_ENUM_MEMBER_PROPERTY("LoopEndAction", ezOnComponentFinishedAction, m_EndAction)
```
C-Core: Would be good to call this "OnFinishedAction" for consistency with existing components

C-Core (approved):
I pushed a small fix and cleanup for the message handlers. It looks good now. I would suggest that we merge these changes and then you can work on your remaining todos if you want to. Again thanks a lot for this contribution!

Mr-Fishy:
Thank you! I should be able to get to finishing out the features this weekend for the sprite animations.

PR #1974, opened 2026-07-02
Render Pipeline Blackboards

C-Core:
### Blackboards
Previously, one would always need special code to set render pipeline properties. Now ezView can pull the properties out of blackboard(s) which makes it much easier to work with in gameplay code and script.
* Added a blackboard to ezWorld so world specific data can be stored in there. This adds a new layer in between local and global blackboards. The ezBlackboardComponent::FindBlackboard function will return the world's blackboard if no local blackboard is found and no name is specified.
* Added an optional blackboard to ezView for view specific data. The view will pull data from the world's blackboard and its own blackboard. The view blackboard has priority and will overwrite world data with the same key. The keys of render pipeline properties need to be in the form of "RenderPassName.PropertyName" or "ExtractorName.PropertyName".
* The volume sampler component can now also write the sampled values to a blackboard. With that the volume sampler component can act like the post processing component while being more generic. Therefore the post processing component has been removed and while be automatically migrated to a volume sampler component. **NOTE:** When a post processing component was attached to a render-to-texture camera you will need to manually add blackboard component, otherwise the new volume sampler component will write to the world's blackboard and thus alter the properties of the main view. If the volume sampling functionality was not needed for the render-to-texture camera you can also set the properties on the blackboard component directly and remove the volume sampler component.

### Fog
The volume sampler component and script was used to change fog properties depending on the camera position. This was tedious to setup and had did not allow to easily change the fog properties in the component anymore.
* Now the fog component can read values in the form of e.g. "Fog.Color", "Fog.Density" etc. from the world's blackboard and combine them with its own properties.
* The FogSampler script has been removed from the testing chambers project and replaced with the new mechanism above.

### Misc
* Fixed stale texture handle issue in automatic texture assigment for render passes


jankrassnigg:
I guess these need to be updated:

https://ezengine.net/pages/docs/effects/post-processing/post-processing-component.html

https://ezengine.net/pages/docs/effects/post-processing/volume-components.html

Sanakan8472 (approved):

PR #1986, opened 2026-07-16
Various fixes

C-Core:
* Introduced world handles to check whether a world has been deleted. Skip render pipeline which are associated with a deleted world. This prevents binding deleted buffers etc.
* Fixed linking of custom spline tangents
* Moved procgen plugin data to be consistent with other plugins
* Exposed Create Game Object, Create Component, String::IsEmpty, Component::SetUniqueID to visual script
* Fixed reflected string view properties in visual script
* Fixed some crashes and memory corruptions in VS editor and compiler
* Allow moving static objects through exposed parameters


Code/Engine/Core/World/Implementation/World.cpp
```diff
@@ -674,6 +658,27 @@ const ezWorldModule* ezWorld::GetModule(const ezRTTI* pRtti) const
...
+  desc.m_hParent = hParent;
+  desc.m_LocalPosition = vLocalPosition;
+  desc.m_LocalRotation = qLocalRotation;
+  desc.m_LocalScaling = vLocalScale;
+
+  // Prevent zero scale, which can easily happen when creating objects from script, as this can easily break all sort of things down the line.
+  if (desc.m_LocalScaling.IsZero(ezMath::DefaultEpsilon<float>()))
+  {
+    desc.m_LocalScaling.Set(0.0001f);
+  }
```
jankrassnigg, line 677: What about uniform scale, some components only support that. Do we extract it from the local scale, or should it be possible, to set it directly?

jankrassnigg (approved):

Code/Engine/Core/World/Implementation/World.cpp
C-Core: Strictly speaking, uniform scale is only there for editing convinience. It does not exist anymore in global transforms, so components that only support uniform scale will just take the max component of the global scale. 
Nevertheless, I have added uniform scale to the script function for consistency.

Sanakan8472 (approved):

PR #1991, opened 2026-07-21
Use portable SIMD on MSVC ARM64

BillyONeal:
MSVC does not provide __builtin_shufflevector, which the NEON SIMD implementation requires. Select the FPU implementation for the native MSVC frontend while retaining NEON for Clang.

This change was authored by GPT 5.6-Sol


jankrassnigg:
Made a better fix in #1993

BillyONeal:
Thank you!

PR #1992, opened 2026-07-21
Apply build filters before dependency checks

BillyONeal:
Return from excluded optional projects before calling dependency helpers supplied by their optional source trees. This allows focused build filters to configure without those dependencies present.

(I am touching this because the vcpkg port says `-DEZ_BUILD_FILTER=FoundationOnly` and we intentionally do not provide those dependencies. https://github.com/microsoft/vcpkg/blob/4bca8fd8654e5ba76f92661db7bfe954768ad8ef/ports/ezfoundation/portfile.cmake#L22 )

This change authored by GPT 5.6-Sol


jankrassnigg:
Slightly better fix is in #1993.

BillyONeal:
Thanks!

PR #1999, opened 2026-07-22
Decals now support atlas textures for more variation

jankrassnigg:
Added row/column count to decal asset. If available, a random one is used. Also added sample textures for bullet holes on metal, wood and glass that use this functionality.

image

### DIFF
--- a/Code/Engine/RendererCore/Decals/DecalComponent.h
+++ b/Code/Engine/RendererCore/Decals/DecalComponent.h
@@ -115,6 +115,15 @@ public:
   void SetDecal(ezUInt32 uiIndex, const ezDecalResourceHandle& hResource); // [ property ]
   const ezDecalResourceHandle& GetDecal(ezUInt32 uiIndex) const;           // [ property ]
 
+  /// \brief Selects which variation of the decal texture to display.
+  ///
+  /// Only has an effect if the decal asset declares that its texture contains a grid of variations.
+  /// Zero means that a variation is chosen randomly, based on the owner object's stable random seed.
+  /// Otherwise the value is a one-based index into the grid, values that exceed the number of
+  /// available variations wrap around.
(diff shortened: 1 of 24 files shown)


Code/Engine/RendererCore/Decals/DecalComponent.h
```diff
@@ -115,6 +115,15 @@ class EZ_RENDERERCORE_DLL ezDecalComponent final : public ezRenderComponent
   void SetDecal(ezUInt32 uiIndex, const ezDecalResourceHandle& hResource); // [ property ]
   const ezDecalResourceHandle& GetDecal(ezUInt32 uiIndex) const;           // [ property ]
 
+  /// \brief Selects which variation of the decal texture to display.
+  ///
+  /// Only has an effect if the decal asset declares that its texture contains a grid of variations.
+  /// Zero means that a variation is chosen randomly, based on the owner object's stable random seed.
+  /// Otherwise the value is a one-based index into the grid, values that exceed the number of
+  /// available variations wrap around.
+  void SetVariation(ezUInt8 uiVariation); // [ property ]
```
C-Core: It's a bit odd that we use a one based index here, we never do that elsewhere. I think a signed int 8 with -1 as default would be a more fitting choice.

jankrassnigg: Yeah, I didn't want to use a signed 16 bit integer. And theoretically 128 values are insufficient, because you can have a 16x16 grid. But maybe we can ignore that.

C-Core: With a one based index you can't have a 16x16 grid either if you reserve 0 for random variation

jankrassnigg: Haha, true, off-by-one error.

PR #2000, opened 2026-07-22
Add build-specific plugin package dependencies

AlexandrKedrov:
Added separate dependency lists for Dev/Debug and Shipping builds. Also added `RmlDebugger.dll` to the RmlUi bundle as a Dev/Debug dependency.

### DIFF
--- a/Code/Editor/EditorFramework/Project/Implementation/ProjectExport.cpp
+++ b/Code/Editor/EditorFramework/Project/Implementation/ProjectExport.cpp
@@ -508,6 +508,17 @@ ezResult ezProjectExport::ExportProject(const char* szTargetDirectory, const ezP
         binariesFilter.AddFilter(dep, true);
       }
 
+#if EZ_ENABLED(EZ_COMPILE_FOR_DEVELOPMENT)
+      for (const auto& dep : it.Value().m_PackageDevDebugDependencies)
+      {
+        binariesFilter.AddFilter(dep, true);
+      }
+#else
+      for (const auto& dep : it.Value().m_PackageShippingDependencies)
+      {
+        binariesFilter.AddFilter(dep, true);
+      }
+#endif
       for (const auto& dep : it.Value().m_RuntimePlugins)
       {
         ezStringBuilder tmp = dep;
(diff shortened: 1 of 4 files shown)


Code/Editor/EditorFramework/Project/Implementation/ProjectExport.cpp
```diff
@@ -508,6 +508,17 @@ ezResult ezProjectExport::ExportProject(const char* szTargetDirectory, const ezP
...
+      for (const auto& dep : it.Value().m_PackageDevDebugDependencies)
+      {
+        binariesFilter.AddFilter(dep, true);
+      }
+#else
+      for (const auto& dep : it.Value().m_PackageShippingDependencies)
+      {
+        binariesFilter.AddFilter(dep, true);
+      }
+#endif
```
jankrassnigg: Very good that you layer the DevDebug and Shipping on top of the generic list of dependencies. The only thing I'm wondering, is whether we should split DevDebug into two lists, so that we have all three separately. I don't know whether there are cases where this would be necessary.

@C-Core any opinion?

We could also leave it like this, and if necessary, add separate Dev + Debug lists later. But right now I think just having three separate lists is cleaner, and future proof for all use cases.

C-Core: I would also go for three separate lists

AlexandrKedrov: Done

Code/EnginePlugins/RmlUiPlugin/RmlUi.ezPluginBundle
```diff
@@ -9,6 +9,8 @@ PluginInfo
 	string %EditorEnginePlugins{"ezEnginePluginRmlUi"}
 	string %RuntimePlugins{"ezRmlUiPlugin"}
 	string %PackageDependencies{"freetype.dll", "RmlCore.dll"}
+	string %PackageDevDependencies{"RmlDebugger.dll"}
```
jankrassnigg: One minor adjustment, please rename them to "PackageDependenciesDev", "PackageDependenciesDebug" and "PackageDependenciesShipping".

jankrassnigg (approved):
Very nice, thank you 👍

PR #2003, opened 2026-07-23
Fix precision-loss warning caused by implicit conversion

AlexandrKedrov:
Visual Studio 2022. When building with Tracy support disabled (`EZ_3RDPARTY_TRACY_SUPPORT=OFF`), I get a precision-loss warning that is treated as an error.

This change fixes the warning, and my build completes successfully. However, I would note that it may lower the precision. The original code uses double precision for the division, and the equivalent code would require `(Type)(1.0 / s)`.

I’m not sure whether this could be a problem, but quaternion arithmetic might be delicate.

### DIFF
--- a/Code/Engine/Foundation/Math/Implementation/Quat_inl.h
+++ b/Code/Engine/Foundation/Math/Implementation/Quat_inl.h
@@ -88,7 +88,7 @@ void ezQuatTemplate<Type>::GetRotationAxisAndAngle(ezVec3Template<Type>& out_vAx
   }
   else
   {
-    const Type ds = 1.0 / s;
+    const Type ds = (Type)1.0 / s;
     out_vAxis.x = x * ds;
     out_vAxis.y = y * ds;
     out_vAxis.z = z * ds;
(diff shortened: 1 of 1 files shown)


Code/Engine/Foundation/Math/Implementation/Quat_inl.h
```diff
@@ -88,7 +88,7 @@ void ezQuatTemplate<Type>::GetRotationAxisAndAngle(ezVec3Template<Type>& out_vAx
   }
   else
   {
-    const Type ds = 1.0 / s;
+    const Type ds = (Type)1.0 / s;
```
jankrassnigg: What if you do

```cpp
(Type)(1.0 / s)
```
So cast after the calculation. Does that still work for you?

AlexandrKedrov: Yes, it does

AlexandrKedrov: I chose `(Type)1.0 / s` to reduce the number of potential float -> double -> float types of conversions

jankrassnigg: That shouldn't be a problem. The FPU works with 80 bit registers anyway. But this way you maybe get slightly better precision.

AlexandrKedrov: It is actually an interesting topic. Frankly, I would say keeping all the data the same size, is easier to vectorize using SIMD instructions.

But I agree, the performance difference is likely to be negligible 

UPD I tried compiling both versions: gcc and clang yield the same assembly using single precision AVX instructions

PR #2004, opened 2026-07-27
improved process of downloading project from git

veksha:
i had an issue when downloading Store sample. progress bar was not updating and after 60 secs ezEngine tried to open downloaded project but failed, because cloning was still in progress.

in code we see that ezEngine waits only 60 secs for git cloning to finish then it tries to open the cloned project.
i changed it so it now waits indefinitely. this change fixed the problem.

I also made progress bar to actually show the cloning progress.

Tested only on Windows.

### DIFF
--- a/Code/Editor/EditorFramework/EditorApp/EditorApp.cpp
+++ b/Code/Editor/EditorFramework/EditorApp/EditorApp.cpp
@@ -590,13 +590,10 @@ ezStatus ezQtEditorApp::MakeRemoteProjectLocal(ezStringBuilder& inout_sFilePath)
   // if it is a git repository, clone it
   if (sType == "git" && !sUrl.IsEmpty())
   {
-    ezProgressRange progress("Downloading Project", 2, true);
-    progress.SetStepWeighting(0, 0.8f);
-    progress.SetStepWeighting(1, 0.2f);
-
     {
       QStringList args;
       args << "clone";
+      args << "--progress"; // it appears that this flag forces non-buffered I/O so it flushes output immediately after each write
       args << ezMakeQString(sUrl);
       args << ezMakeQString(sName);
 
@@ -606,6 +603,7 @@ ezStatus ezQtEditorApp::MakeRemoteProjectLocal(ezStringBuilder& inout_sFilePath)
 
       ezProgressRange cloneProgress("Downloading Project", true);
       bool bRecursion = false;
+      bool bFirstProgress = true;
 
       QObject::connect(&proc, &QProcess::readyReadStandardOutput, [&]()
         {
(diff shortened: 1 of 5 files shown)


Code/Editor/EditorFramework/EditorApp/EditorApp.cpp
```diff
@@ -651,7 +657,10 @@ ezStatus ezQtEditorApp::MakeRemoteProjectLocal(ezStringBuilder& inout_sFilePath)
         return ezStatus(ezFmt("Running 'git' to download the remote project failed."));
       }
 
-      proc.waitForFinished(60 * 1000);
+      while (!proc.waitForFinished())
```
jankrassnigg: 60 seconds was a pretty naive implementation. However, this is now going to be a problem when the clone fails, which can also happen. 

Best case would be to handle different error scenarios. Eg have a timeout that keeps getting reset when we see new output, ie when progress is made. 

I haven't dug deeper into this, but maybe you can take another look how to make this a bit more robust. Git should already handle most things, though, so maybe waiting indefinitely is also good enough. Would be useful to check the git documentation how it handles errors.

AlexandrKedrov: A little workaround suggestion - is to use progress bar with a "Break"/"Cancel" option. Than a user can manually control whether something has gone wrong or not.

UPD. Sorry, I missed that the progress bar already has "cancel" option.  Then doesn't it make the proposed solution robust enough?

veksha: It seems that the Cancel button doesn't terminate all git child processes, and cloning can't be stopped by Cancel button. but maybe we can use ezProcessGroup class to make it work.

jankrassnigg: The cancel button of course needs to be wired into something. The process group could work, it has a terminate option. Give it a try. 

If it doesn't work, your current implementation is also okay.

veksha:
new changes:
1. used ezProcessGroup. now Cancel button works good. also needed to support "\r" when reading the stderr pipe (Foundation/Platform/Win/Process_Platform.inl). tested on Windows only!
2. progressbar now supports value reset (Foundation/Utilities/Implementation/Progress.cpp). this is needed to simplify git output parsing, no need for hacks like skipping first progress and hacks like parsing English text ("Receiving objects: 2%"), text could be non-english, so it is better to parse only "%" symbol. now every new progress reported by git resets progressbar, nice.
3. on git error show last stderr line in a message box.
4. changed progress dialog modality from WindowModal to ApplicationModal, so dashboard can't be interacted with.
5. another cancel button fix - cancel was working only on progress change, now instantly.
6. progress bar precision was bad because of this code `e.m_pProgressbar->GetCompletion() * 1000.0`. float was converted to double and 0.02 became 0.01999, and progress bar showed 1% instead of 2% and so on.. fixed.

Code/Editor/EditorFramework/EditorApp/EditorApp.cpp
veksha: please see message below

jankrassnigg (approved):
Thanks a lot!

veksha:
> Thanks a lot!

You are welcome.

Jan, Do you know if it working on Linux too?

jankrassnigg:
> Jan, Do you know if it working on Linux too?

No idea. It generally looks good, but I can't really test. @Sanakan8472 is regularly working on Linux, so maybe he can give it a try some time.

PR #2014, opened 2026-08-05
added simple weapon bobbing and sway procedural animations (visual script)

veksha:
bonus: weapon animation when jumping and a little breathing animation while player is idle.

Feel free to suggest your ideas and optimizations.

https://github.com/user-attachments/assets/affdcf2b-7599-4373-9463-1a89f2276073


veksha:
oh, i forgot to make it fps-independent, will do it later, please wait.

C-Core:
I took a quick look at the visual script, looks quite nice overall. There are couple of things I don't understand:
* Why do you need an extra blackboard to store e.g. sway_position? Can't you just store that in a script variable?
* What is the jump counter coroutine supposed to do? I don't see why you would need a coroutine for that?

veksha:
>Why do you need an extra blackboard to store e.g. sway_position? Can't you just store that in a script variable?

I will try, I thought they are local scope, so I decided to use blackboard as they are global (i need to store previous value for Lerp)

>What is the jump counter coroutine supposed to do? I don't see why you would need a coroutine for that?

i needed a counter variable to make jumping animation always the same, (starting from 0 everytime). maybe it's better to use time difference var for that, but i used coroutine just for fun, why not.

C-Core:
Coroutines are useful for distributing work over multiple frames or running work in "parallel". I don't see why you would need any of those features for the counter which is updated every frame anyways. Since you also read the counter in the rest of the function you create some sort of a race condition. You don't know exactly if the counter is set before or after the read. 
Furthermore coroutines come with overhead since the script runtime needs to allocate extra memory for the coroutine state and all its temporary variables.
Unless you have good reason I would get rid of the coroutine in that script.

veksha:
@C-Core thanks for info! good points

veksha:
@C-Core 
1. replaced counter based on coroutine by time difference counter
2. made frame-rate independent by using delta time
tested on 120 fps and 20 fps

some difference of jumping animation is observed on 20 fps, but i guess it's because TouchingGround variable is lagging behind for some reason (maybe bug in default character controller)

jankrassnigg:
This is quite amazing! Makes the game feel so much more natural!

What @C-Core mentioned in the beginning, the use of the global blackboard is unnecessary. The script does exist across frames. In your init step you query a global blackboard by name. That name is empty, which should even be disallowed, I'm not sure what the script does in this case. But anyway, you get a pointer to the blackboard apparently, and then you store that in your "global_blackboard" variable. Later you then access the blackboard through that pointer to read and write for instance "sway_position":

image

This is unnecessarily complicated. First of all, you don't need the "Get" node, on the GetEntryValue node, you can pass the "Name" directly as a property. The pin connection is only needed, when the value can be dynamic. But anyway, you don't need the blackboard here, at all. Just use a script variable directly. E.g. here you add a variable called "sway_position" and you give it the VALUE "sway_position":

image

Instead, just change its type from HashedString to Vector3 and use that variable directly.

image

Then of course adjust all places where this is used.

Generally: The blackboard is needed when data needs to be shared between different systems. So for example the animation system and player script detect things like "touching_ground" etc, and your script reacts to that. Same, if you wanted to report something to the player script, then you'd write that to a blackboard.

But as long as state is exclusive to the script, ie. nobody else needs to know about it, you can just use the script variables directly. They live for as long as the script component lives.

veksha:
I removed use of global blackboard. For some reason i thought all variables are local, but now it makes sense. Very clever and convenient system. 👍

jankrassnigg:
Script looks really good now. Let's wait for @C-Core to take another look, but I think this is fine now. Really great work!

C-Core (approved):

PR #2016, opened 2026-08-08
Fixed that pins didn't get transferred to a pasted Expression node

veksha:
1. create Visual Script
2. add Expression node
3. create some input/output pins
4. copy that node (ctrl+c)
5. paste (ctrl+v)
6. there are no pins on pasted node.

image

### error
also if you copy/paste Expression node _together with nodes connected to its pins_ - you will get an error:

image

### The fix:
move `InternalCreatePins` from BeforeObjectAdded event to AfterObjectAdded event.

### Comments
I don't really know why my fix works, because i'm not an expert. Just tried some idea and it happened to do the job. Please check and correct my fix if it is breaking something.


C-Core:
Thanks for looking into this issue. I will try to verify it once I'm back from vacation. This is quite delicate code so it needs to be tested thoroughly.

C-Core (approved):
I have tested it and haven't found any issues so far. So let's merge it if CI is happy as well.

PR #2020, opened 2026-08-17
Fixed #1970: Show manual transform errors in Asset Curator

### LINKED ISSUE #1970: [BUG] Broken scene does not show up in Curator when required plugin is missing
* Make sure the SampleGamePlugin is NOT compiled in Visual Studio
* Open the Sample Game Project.
* Transform the main scene
* The log will show an error (missing component type), but nothing shows up in the Asset Curator panel

atirna:
## Why

A manual transform that cannot open its asset document, or fails while transforming it, only writes an error to the log. Its transform state stays unchanged, so the Asset Curator filter does not list the broken asset.

## Changes

- record `TransformError` for a failed manual transform
- use the sub-asset's asset info when reporting the error
- extend `Asset Document Tests / Save on Transform` with the invalid OBJ case and assert the two expected path errors

## Verification

- `git diff --check`: pass
- the Windows editor test remains the runtime check for this patch

Fixes #1970

### DIFF
--- a/Code/UnitTests/EditorTest/AssetDocument/AssetDocumentTest.cpp
+++ b/Code/UnitTests/EditorTest/AssetDocument/AssetDocumentTest.cpp
@@ -3,7 +3,6 @@
 #include <EditorFramework/Assets/AssetCurator.h>
 #include <EditorFramework/Assets/AssetDocument.h>
 #include <EditorTest/AssetDocument/AssetDocumentTest.h>
-#include <Foundation/IO/FileSystem/FileWriter.h>
 #include <Foundation/IO/OSFile.h>
 #include <TestFramework/Utilities/TestLogInterface.h>
 #include <ToolsFoundation/FileSystem/FileSystemModel.h>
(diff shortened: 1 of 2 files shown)


Code/UnitTests/EditorTest/AssetDocument/AssetDocumentTest.cpp
```diff
@@ -164,6 +165,33 @@ void ezEditorAssetDocumentTest::SaveOnTransform()
...
+
+    pAcc->StartTransaction("Set Invalid Mesh");
+    EZ_TEST_BOOL(pAcc->SetValueByName(pMeshAsset, "MeshFile", "Meshes/Invalid.obj").Succeeded());
+    pAcc->FinishTransaction();
+    EZ_TEST_BOOL(pDoc->SaveDocument().Succeeded());
+
+    const ezTransformStatus res = ezAssetCurator::GetSingleton()->TransformAsset(pDoc->GetGuid(), ezTransformFlags::ForceTransform | ezTransformFlags::TriggeredManually);
+    EZ_TEST_BOOL(res.Failed());
+
+    ezAssetInfo* pAssetInfo = ezAssetCurator::GetSingleton()->GetAssetInfo(pDoc->GetGuid());
```
Sanakan8472: `GetAssetInfo` is private, you will have to use `GetSubAsset()->m_pAssetInfo` for thread safety.

Sanakan8472 (changes requested):

Code/UnitTests/EditorTest/AssetDocument/AssetDocumentTest.cpp
atirna: switched the test to GetSubAsset()->m_pAssetInfo so it takes the curator lock. same seam as the File Operations checks in this file.

PR #2038, opened 2026-08-30
Render pipeline rework

Sanakan8472:
## Render pipeline refactor

* **Pass graph**: The pass graph is no longer part of `ezRenderPipeline`. It has been rewritten and now lives in its own `ezRenderPipelinePassGraph` class, which owns the passes, pins and connections and performs dead pass culling, dependency sorting and render pass creation.
The graph is also now immutable: The passes and connections have to be passed in at construction time into the `ezRenderPipeline` constructor.
* **Sub-graphs**: Render pipeline assets can be reused inside other render pipeline assets. A `ezSubGraphNode` references another pipeline, and boundary nodes (`ezSubGraphTextureInputNode`, `ezSubGraphTextureOutputNode` and their buffer variants) declare its pins. Referenced pipelines are inlined during asset transform, so boundary nodes never reach a runtime pipeline.\
This is a sub-graph render pipeline asset:
  image
  This is the use of it in another render pipeline:
  image
* **Switch passes**: `ezTextureSwitchPass` and `ezBufferSwitchPass` forward one of up to eight inputs to their output, selected by a blackboard value. Branches that only feed unselected inputs are culled from the pipeline. This allows the behaviour of the immutable render pipeline to change at runtime. If no blackboard entry is present or the value does not match any of the switch values, the node falls back to the first switch value.
  image
* **Buffer support**: Pins are now either texture or buffer pins, and connections carry the corresponding resource type. Texture and buffer pins can not be connected to each other.

## Texture format refactor

`ezSourcePass` and `ezHistorySourcePass` no longer name a concrete texture format. They describe what they need through type, minimum precision and minimum channel count, and the cheapest supported format is picked per device.
  image

## View blackboard changes

* **Switch Support**: The view now tracks for each switch the blackboard it needs to source the value from. If a switch value changes, the `ezRenderPipelinePassGraph` has to be rebuilt (culling and sorting).
* **Property Mapping Fixes**: Removing an entry from a view blackboard now correctly falls back the property to the world blackboard, or restores the property's default value.

## Vulkan

Swap chain creation only requests stretch scaling if the surface supports it for the present mode in use, and clamps the image extent to the matching range.

## RendererTest

`GpuPipelineTest` covers dead pass culling, dependency and pass-through sorting, cycle detection, switches, sub-graph inlining, buffer and mixed texture/buffer connections, and view blackboard property and switch mapping.


jankrassnigg:
These nodes can be created from the menu, but seem to be empty. Are they base classes?

grafik

PR #2039, opened 2026-08-31
Fix unused variable warning

AlexandrKedrov:
Build fails for VS2019 Debug build with `$CMAKE_ARGS += "-DEZ_3RDPARTY_TRACY_SUPPORT=OFF"` flag


jankrassnigg:
Thanks! Though be aware that vs 2019 is not on our list of supported compilers anymore, and I'm not even sure how long we'll keep vs 2022 around.

AlexandrKedrov:
> Thanks! Though be aware that vs 2019 is not on our list of supported compilers anymore, and I'm not even sure how long we'll keep vs 2022 around.

Oh. I meant VS19 aka VS 2022. I always mess up with this enumeration...

PR #2040, opened 2026-08-31
Fix character sliding on tilted surfaces

AlexandrKedrov:
I found the simple solution to the problem that was brought by NicoRTX in the discord channel. 

Previously, the update order was:
1. Stick character to the ground: `m_fVelocityUp = groundVerticalVelocity;`
2. Apply gravity: `m_fVelocityUp += gravity;`
3. Move character with `m_fVelocityUp`

My change make the order slightly different:
1. Stick character to the ground: `m_fVelocityUp = groundVerticalVelocity;`
2. Move character with `m_fVelocityUp`
3. Apply gravity: `m_fVelocityUp += gravity;`

However, even with my change (or the change proposed by NicoRTX) the character still slips when landing on tilted surfaces (for example, after jumping). I can try fixing that but I don't see any robust solution without adding extra sweep tests, that I don't want to add.

### DIFF
--- a/Code/EnginePlugins/JoltPlugin/Character/Implementation/JoltDefaultCharacterComponent.cpp
+++ b/Code/EnginePlugins/JoltPlugin/Character/Implementation/JoltDefaultCharacterComponent.cpp
@@ -728,6 +728,14 @@ void ezJoltDefaultCharacterComponent::UpdateCharacter()
     vRootVelocity -= vLatDir * fProj;
   }
 
+  // retrieve the actual up velocity
+  float groundVerticalVelocity = GetJoltCharacter()->GetGroundVelocity().GetZ();
+  if (GetJoltCharacter()->GetGroundState() == JPH::CharacterVirtual::EGroundState::OnGround // If on ground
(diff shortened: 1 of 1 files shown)


veksha:
This is what I do locally. Don't know if it's a best way.

```diff
@@ -760,9 +760,9 @@ void ezJoltDefaultCharacterComponent::UpdateCharacter()
   {
     ClampLateralVelocity();
     ClampUpVelocity();
-  }

-  m_fVelocityUp += GetUpdateTimeDelta() * pModule->GetCharacterGravity().z;
+    m_fVelocityUp += GetUpdateTimeDelta() * pModule->GetCharacterGravity().z;
+  }

   ApplyRotationZ();

```

AlexandrKedrov:
It looks very nice! However, I will note that it uses `bWasOnGround` that is saved from the "previous" frame, and its values might be a bit outdated.  It shouldn't be a problem though, and I don't mind opting for your solution as far as it looks even more simple.

jankrassnigg:
Seems to be working well. Thanks for the contribution!

PR #2041, opened 2026-09-01
Fixed material permutation variable handling + extended terrain material

jankrassnigg:
If a material sets a permutation variable to a fixed value, e.g. "BLEND_MODE = OPAQUE", but the shader reads that somewhere other than the shader code (e.g. the material config section), it would not work so far. The work around was to still expose the variable to the UI. This change fixes that.

Also extended the terrain material list.

### DIFF
--- a/Code/EditorPlugins/Assets/EditorPluginAssets/MaterialAsset/MaterialAsset.cpp
+++ b/Code/EditorPlugins/Assets/EditorPluginAssets/MaterialAsset/MaterialAsset.cpp
@@ -55,26 +55,49 @@ namespace
     return EZ_FAILURE;
   }
 
+  /// Adds preprocessor defines for a permutation variable that the shader pinned to a fixed value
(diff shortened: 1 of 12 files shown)


Code/EditorPlugins/Assets/EditorPluginAssets/MaterialAsset/MaterialAsset.cpp
```diff
@@ -55,34 +55,107 @@ namespace
...
+
+          permutationProperties.PushBack(pProp);
+        }
+      }
+
+      // Permutation variables that the shader pinned to a fixed value in [PERMUTATIONS] are not
+      // exposed as material properties, but [MATERIALCONFIG] may still evaluate them. Define those
+      // as well, skipping any that the material exposes, whose own value takes precedence.
+      {
+        ezHybridArray<ezHashedString, 16> permVars;
```
C-Core: TempHybridArray

PR #2043, opened 2026-09-01
Physics simulation for static actors for easier placement

jankrassnigg:
In the editor, run the simulation, select an object that uses a static Jolt actor, then press "K" (keep current transform). The first time, this will turn the static actor into a dynamic actor and physically simulate it, so it will fall down etc. When it settled, and you like the position, press "K" again, this time to record the transform. When the simulation is stopped, the editor will automatically move the object to the recorded location.

This was already working for dynamic physics objects before, now it also works for static and kinematic objects, as long as they have convex shapes.

Video below shows this: 
* Start simulation
* Press "K" to simulate the otherwise static rock
* Press "K" again to record location
* Stop simulation -> rock stays where it was

https://github.com/user-attachments/assets/6d32f90d-4c73-43e3-b6c5-27247f871b7e

### DIFF
--- a/Code/EditorPlugins/Scene/EnginePluginScene/SceneContext/SceneContext.cpp
+++ b/Code/EditorPlugins/Scene/EnginePluginScene/SceneContext/SceneContext.cpp
@@ -3,6 +3,7 @@
 #include <EnginePluginScene/SceneContext/SceneContext.h>
 #include <EnginePluginScene/SceneView/SceneView.h>
 
+#include <Core/Interfaces/PhysicsWorldModule.h>
 #include <Core/Interfaces/SoundInterface.h>
 #include <Core/Prefabs/PrefabResource.h>
 #include <Core/World/EventMessageHandlerComponent.h>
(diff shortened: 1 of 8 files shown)


wdstudiosma:
I like this! reminds me of source 2 hammer's simulate object

jankrassnigg:
> I like this! reminds me of source 2 hammer's simulate object

We've had this feature for years, now it's just even more useful: https://ezengine.net/pages/docs/editor/run-scene.html#keep-simulation-changes

veksha:
is Jolt Box Shape Component supported for this placement?
I have a static mesh object with a **child** object which has Jolt Box Shape Component with some position offset (important).
after trying physics simulation on my object it is correctly fallen to the ground, but after i save its position with pressing K again and stopping the simulation I get wrong position offset for my object.

veksha:
sorry, the real problem is in LocalUniformScaling being 0.6. this is why position is wrong.

jankrassnigg:
Generally yes. If your top level object has scaling, then that will probably make problems. However, if you put that into a prefab, then it will work for the prefab. I can take a look whether scaling can be made to work as well.

jankrassnigg:
@veksha I just tried with different variations, using a mesh, using a box, with position offset and without, with local scaling and without, it all worked perfectly. If you can still reproduce this, can you please open a ticket and attach a prefab with an object where it doesn't work?

PR #2047, opened 2026-09-02
Added reflection probe prioritization for faster results

jankrassnigg:
Reflection probes are now prioritized using two queues:

* A "first time" queue, for probes that have no content yet.
* A "refresh" queue, for probes that already have content.

All probes that never got updated yet first end up in the first queue. Dynamic probes or probes that need a refresh for some other reason then go to the second queue. The first queue is always drained first, so a probe with nothing to show is never held up by probes that already look correct.

Both queues are sorted by the already implemented size and distance based priority, so the probes around the camera are updated first. In the first queue the skylight probe gets the highest priority. Thus skylights are now updated FIRST, so far they were updated LAST and then invalidated all other probes.

In the refresh queue a probe's priority also grows the longer it waits. Without this, sorting purely by distance starves distant probes indefinitely: with more probes than can be refreshed between two skylight updates, the ones far away would never get their turn again.

The "first time" queue allows for bursts of updates: a skylight without content renders all 6 sides plus the filter step in a single frame, other probes at least 2 sides per frame, to make the scene show up properly faster. Probes that only refresh stay within the configured budget, so the cost in a running scene is unchanged.


jankrassnigg:
The reflection test has a difference. It is very hard to see what the original image actually showed. The new image looks like it is more correct. But please look closely.

jankrassnigg:
After checking more closely, it looks like in the original image the reflection of the ground was not yet lit by the sky light. Therefore the reflection is darker. The image is captured after 30 frames. A couple of frames later it would converge to the same result as it does now. Actually nice proof that the old code had a significant delay. After another fix, it now actually converges in 8 frames (skylight is done fully in frame 0, each static probe updates 2 sides per frame + 1 convolution frame -> 4 frames for each probe).

Code/Engine/RendererCore/Lights/Implementation/ReflectionProbeUpdater.cpp
```diff
@@ -142,6 +174,17 @@ ezResult ezReflectionProbeUpdater::StartFilterUpdate(const ezReflectionProbeRef&
   return EZ_FAILURE;
 }
 
+bool ezReflectionProbeUpdater::IsFirstBakeInProgress() const
```
Sanakan8472, line 179: If `cvar_RenderingReflectionPoolMaxUpdatesPerFrame` is set to 2+ and sharing budget to false, this function will falsely return true even though no first bake is in progress.

Sanakan8472 (approved):

PR #2052, opened 2026-09-07
Updated Meshopt to v1.2

jankrassnigg:
The new meshopt version also implements tangent space generation, similar to MikkTSpace, but much faster, so this now replaced that code by the meshopt code.


C-Core:
Is it only similar to MikkT or is the result exactly the same? If it is not the same we cannot replace it. The tangent space has to match the one in the baking program and MikkT is the standard today for most baking programs.

jankrassnigg:
> Is it only similar to MikkT or is the result exactly the same? If it is not the same we cannot replace it. The tangent space has to match the one in the baking program and MikkT is the standard today for most baking programs.

There is an option to make it identical. I chose to use the new/better functionality, but we can use the identical one.

Code/ThirdParty/meshoptimizer/meshoptimizer.h
```diff
@@ -898,6 +923,33 @@ MESHOPTIMIZER_EXPERIMENTAL size_t meshopt_opacityMapEntrySize(int level, int sta
  */
 MESHOPTIMIZER_EXPERIMENTAL size_t meshopt_opacityMapCompact(unsigned char* data, size_t data_size, unsigned char* levels, unsigned int* offsets, size_t omm_count, int* omm_indices, size_t triangle_count, int states);
 
+/**
+ * Tangent generation options
+ */
+enum
+{
+	/* Produce tangents compatible with MikkTSpace (same weighting and fallbacks) at the cost of reduced quality. Not recommended unless normal maps are baked. */
+	meshopt_TangentCompatible = 1 << 0,
```
jankrassnigg, line 932: This is the flag for compatibility.

PR #2053, opened 2026-09-08
ezBlackboardCondition is now shown in one line

C-Core:
When you have a couple of conditions the UI became very cluttered.
Before:
Screenshot 2026-09-08 114043
After:
Screenshot 2026-09-08 114240

* Made ezBlackboardCondition a custom variant type including patch for state machine transitions.
* New widget to fit everything in one line.
* Extracted dynamic enum button/menu logic into its own class to make it re-usable

### DIFF
--- a/Code/Engine/Core/Utils/Blackboard.h
+++ b/Code/Engine/Core/Utils/Blackboard.h
@@ -250,9 +250,6 @@ struct EZ_CORE_DLL ezBlackboardCondition
 
   bool IsConditionMet(const ezBlackboard& blackboard) const;
 
-  ezResult Serialize(ezStreamWriter& inout_stream) const;
-  ezResult Deserialize(ezStreamReader& inout_stream);
-
   bool operator==(const ezBlackboardCondition& rhs) const
   {
     return m_sEntryName == rhs.m_sEntryName && m_fComparisonValue == rhs.m_fComparisonValue && m_Operator == rhs.m_Operator;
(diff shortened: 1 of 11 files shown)


Code/Engine/Core/Utils/Blackboard.h
```diff
@@ -252,6 +252,30 @@ struct EZ_CORE_DLL ezBlackboardCondition
...
+  {
+    return m_sEntryName == rhs.m_sEntryName && m_fComparisonValue == rhs.m_fComparisonValue && m_Operator == rhs.m_Operator;
+  }
 };
 
 EZ_DECLARE_REFLECTABLE_TYPE(EZ_CORE_DLL, ezBlackboardCondition);
+EZ_DECLARE_CUSTOM_VARIANT_TYPE(ezBlackboardCondition);
+
+EZ_CORE_DLL void operator<<(ezStreamWriter& inout_stream, const ezBlackboardCondition& value);
+EZ_CORE_DLL void operator>>(ezStreamReader& inout_stream, ezBlackboardCondition& ref_value);
```
jankrassnigg: These should not be needed, `ezBlackboardCondition` already has `Serialize`, so this should work out of the box.

jankrassnigg (approved):
Very nice. Could use the same for Blackboard Entry values.

Code/Engine/Core/Utils/Blackboard.h
C-Core: I think the custom variant type needs those, I can maybe remove the Serialize, Deserialze methods instead.

PR #2062, opened 2026-09-13
Disable D3D11 by default and use Vulkan instead

jankrassnigg:


C-Core:
I don't think we should disable DX11 yet since vulkan performance is still lacking.

jankrassnigg:
> I don't think we should disable DX11 yet since vulkan performance is still lacking.

Vulkan is already the default option. This doesn't change anything, really, just that we use Vulkan on CI, and that people don't compile D3D unless they really opt into it, which they already need to do.

I don't know what the latest performance is, but we won't REMOVE D3D for a while. Still, I don't think that hanging on to D3D as the default build has any advantage.

PR #2066, opened 2026-09-16
Unhide Selected

Cutecore-OS:
Hi, it's me, NicoRTX!
Currently, pressing `H` only hides selected objects. If a hidden object is selected, there is no direct way to unhide only that object the only option is to reveal everything via "Show Hidden Objects", which is inconvenient.

### Changes
* **Smart Toggle (`H`):** If all selected objects are hidden, executing the action shows them again; otherwise, it hides them.
* **Dynamic UI:** Action name and icon automatically switch between `Hide Selected` and `Unhide Selected`.
image

I’m not entirely sure if it introduced any subtle bugs. So someone more experienced can review it or implement a cleaner, more proper solution into the engine.


jankrassnigg:
Thank you for the PR, will check it out over the weekend, pretty busy at work atm.

jankrassnigg:
Works really well, thanks!

PR #2068, opened 2026-09-21
Added CVars to control texture resolution

jankrassnigg:
* "Rendering.Textures.DropMips" says how many mipmaps to drop from every loaded texture.
* "Rendering.Textures.MinResolution" says at what resolution NOT to drop further mipmaps. So for instance don't go below 256.
* "Rendering.Textures.MaxResolution" says what resolution never to exceed. For instance, clamp at 2K.

Default values don't drop any mipmaps. When changed, all textures are automatically reloaded.


C-Core:
What is the usecase for these CVars?
While it is fine for experimentation it won't work to implement proper texture quality settings since it is applied to all textures. There are many textures that must not be changed e.g. UI textures or textures used for data in the renderpipeline etc. Just having a min/max resolution is not enough to properly filter those. Texture resolution in it self says nothing about quality without knowing in which context the texture is used. Unreal has texture groups for that. You would basically configure the 3 values you have for each group.

jankrassnigg:
> What is the usecase for these CVars? While it is fine for experimentation it won't work to implement proper texture quality settings since it is applied to all textures. There are many textures that must not be changed e.g. UI textures or textures used for data in the renderpipeline etc. Just having a min/max resolution is not enough to properly filter those. Texture resolution in it self says nothing about quality without knowing in which context the texture is used. Unreal has texture groups for that. You would basically configure the 3 values you have for each group.

Yeah, that's the intended use case. Hadn't thought about texture groups yet. We do have the filter quality level on textures, which controls anisotropic stuff. Would probably need to replace that by some texture group thing.

CVars are quite useful for controlling such options. The question is, whether something like texture groups can be configured that way or not.

Either way, you CAN have this as one global option (still very useful for testing), even if you need a more fine grained way later.

PR #2069, opened 2026-09-21
Prevent RML error message "could not locate data model"

jankrassnigg:


Code/EnginePlugins/RmlUiPlugin/Implementation/SystemInterface.h
```diff
@@ -15,4 +15,27 @@ namespace ezRmlUiInternal
...
+  /// message instead of an error.
+  ///
+  /// A document declares the data models it wants to use through 'data-model' attributes, but those models
+  /// only exist once a data binding has been added to the context. The document is loaded before that
+  /// happens and is loaded again by ezRmlUiCanvasComponentBase::AddDataBinding once a binding exists, so
+  /// the first load legitimately refers to models that aren't there yet. In the editor, which does not run
+  /// the game code that adds the bindings, that state is permanent.
+  /// RmlUi logs an error for it every time and has no way of knowing that it is expected, so the message is
+  /// filtered out on this side instead.
+  class ScopedMissingDataModelsAllowed
```
C-Core: Feels overkill to have an exta scope object just for this single error message when a simple static bool would be enough. We can simply use EZ_SCOPE_EXIT to reset to the previous value.

Code/EnginePlugins/RmlUiPlugin/Components/Implementation/RmlUiCanvasComponentBase.cpp
```diff
@@ -140,6 +141,11 @@ void ezRmlUiCanvasComponentBase::SetRmlResource(const ezRmlUiResourceHandle& hRe
 
     if (m_pContext != nullptr)
     {
+      // Without any data binding the document cannot resolve the data models it declares. That is the normal
+      // state before game code adds its bindings (and permanently in the editor, which doesn't run that code),
+      // and AddDataBinding loads the document again once a binding exists, so it is not worth reporting.
+      const ezRmlUiInternal::ScopedMissingDataModelsAllowed allowMissingDataModels(HasDataBindings() == false);
```
C-Core: The right place to suppress the error would be in ezRmlUi::LoadDocumentFromResource and ezRmlUi::LoadDocumentFromString just before ref_context.LoadDocument

C-Core (changes requested):

C-Core (approved):

